PR #4, opened 2024-06-17
Remove useless `indent` information from `Buffer`

GuillaumeGomez:
First real PR for this project. :)

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -70,26 +70,26 @@ impl<'a> Generator<'a> {
         let mut buf = Buffer::new();
 
         self.impl_template(ctx, &mut buf)?;
-        self.impl_display(&mut buf)?;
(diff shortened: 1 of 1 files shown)


Kijewski:
Good call! Maybe we can later add a `prettyprint` feature to indent the generated code using [prettyplease](https://crates.io/crates/prettyplease], but I think that's very low on my TODO list.

With this change `Buffer::writeln` is infallible. Do you want to remove the Result type result?

GuillaumeGomez:
Good call, let's remove it. :)

GuillaumeGomez:
Done!

Kijewski (approved):
Nice! Much easier to read.

Kijewski:
Thanks! I have a "merge" button now, but I don't have the privileges to add hooks.

GuillaumeGomez:
Let me fix that. :)

PR #5, opened 2024-06-17
Fix cfg using non-existing feature

GuillaumeGomez:
Fixes this warning:

```
warning: unexpected `cfg` condition value: `toml`
   --> rinja_derive/src/config.rs:576:11
    |
576 |     #[cfg(feature = "toml")]
    |           ^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `basic-toml`, `config`, `humansize`, `num-traits`, `serde`, `serde-json`, `urlencode`, `with-actix-web`, `with-axum`, `with-rocket`, and `with-warp`
    = help: consider adding `toml` as a feature in `Cargo.toml`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
```

It's been a while that this was the case though. Seems like these tests were basically never run?

### DIFF
--- a/rinja_derive/Cargo.toml
+++ b/rinja_derive/Cargo.toml
@@ -14,7 +14,7 @@ rust-version = "1.65"
 proc-macro = true
 
 [features]
-config = ["serde", "basic-toml"]
+config = ["dep:serde", "dep:basic-toml"]
 humansize = []
 urlencode = []
 serde-json = []
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::{env, fs};
 
-#[cfg(feature = "serde")]
+#[cfg(feature = "config")]
 use serde::Deserialize;
 
 use crate::{CompileError, FileInfo, CRATE};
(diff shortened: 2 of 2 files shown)


Kijewski:
I think we should update the features to suppress implicit features like `some_crate`, and use `dep:some_crate` instead. Then it will be clearer what the actual feature names are.

rinja_derive/Cargo.toml
```diff
@@ -14,7 +14,9 @@ rust-version = "1.65"
 proc-macro = true
 
 [features]
-config = ["serde", "basic-toml"]
+config = ["serde", "toml"]
```
Kijewski, line 17: I think the use of `serde` and `basic-toml` are implementation details, so the line should read `config = ["dep:serde", "dep:basic-toml"]`, without exposing them as features.

GuillaumeGomez: It's what I did originally but `cargo` was unhappy with `feature = "basic-toml"`. Hence why I went full in.

GuillaumeGomez:
And fixed clippy too. :)

rinja_derive/Cargo.toml
Kijewski: Oh, yeah. I guess *all* usages of `#[cfg(feature = "toml")]`, `#[cfg(feature = "basic-toml")]`, and `#[cfg(feature = "serde")]` are wrong in `rinja_derive`, and should be `#[cfg(feature = "config")]` instead.

GuillaumeGomez: Now that makes more sense. Changing to that then!

Kijewski (approved):

Kijewski:
I love clippy. :) Those kind of bugs and shortcomings are impossible to find unless to read the whole code base repeatedly.

GuillaumeGomez:
It's not clippy but rustc in this case. But yes, it's really awesome!

PR #6, opened 2024-06-17
 Add versioned book on readthedocs.org 

Kijewski:
This PR adds the file `.readthedocs.yaml`, which is used by readthedocs.org to generate the book on their servers. RTD can be used easily host docs for every rinja version, so users can explore how to use the current stable version, the current HEAD, and an older version.

For older versions than the current PR, the books won't be built, because we did not provided the needed `.readthedocs.yaml`.

Also an `index.hbs` template skeleton is dynamically generated to made mdbook and RTD work better together. (Only the left sidebar was changed, but you have to override the whole file to make any changes.)


GuillaumeGomez (approved):

GuillaumeGomez:
Awesome, thanks!

GuillaumeGomez:
Would it be possible to make it updated through github actions whenever a PR is merged? If so, what do you need to make it possible?

Oh and final point: should we reset all versions to `0.1.0`?

PR #8, opened 2024-06-18
Actually execute update-theme.py

Kijewski:
Darn it, sorry! Hopefully this is the last PR to generate the docs. The hook works fine and the docs are rendered and deployed: <https://rinja.readthedocs.io/en/latest/>. But the index.hbs template is not generated, so the version selector is in it's default place and partially overlaps the text.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,8 +5,12 @@ version: 2
 build:
   os: ubuntu-lts-latest
   tools:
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
It's fine, don't worry. Please merge when you want.

PR #14, opened 2024-06-18
derive: add benchmark

### LINKED ISSUE #12: Add benchmarks for derive macro
We could use something like we have [here](https://github.com/rinja-rs/rinja/blob/master/rinja_derive/src/tests.rs#L7). We can do an `include_str!` and then call the parser on it in the bench I think?

Kijewski:
This PR adds the crate "rinja_derive_standalone", which is just like "rinja_derive", though not a "proc_macro". This way we can easily expose it's internals for testing and benchmarking.

Right now, the PR is more or less a prove of concept, and it probably needs a handful more useful benchmark use cases to be worth the hassle. 

Resolves #12.

### DIFF
--- a/rinja_derive_standalone/Cargo.toml
+++ b/rinja_derive_standalone/Cargo.toml
@@ -1,0 +1,42 @@
+[package]
+name = "rinja_derive_standalone"
+version = "0.13.0"
+description = "Procedural macro package for Rinja"
+homepage = "https://github.com/rinja-rs/rinja"
+repository = "https://github.com/rinja-rs/rinja"
+license = "MIT/Apache-2.0"
+workspace = ".."
+readme = "README.md"
+edition = "2021"
+rust-version = "1.65"
+
+[features]
+default = ["__standalone"]
+__standalone = []
+config = ["dep:serde", "dep:basic-toml"]
+humansize = []
+urlencode = []
+serde_json = []
+num-traits = []
+with-actix-web = []
+with-axum = []
+with-rocket = []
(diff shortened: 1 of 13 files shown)


rinja_derive_standalone/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+workspace = ".."
+readme = "README.md"
+edition = "2021"
+rust-version = "1.65"
+publish = false
+
+[features]
+default = ["__standalone"]
+__standalone = []
+config = ["serde", "basic-toml"]
```
GuillaumeGomez, line 17: Missing the `:dep` part you mentioned. ;)

GuillaumeGomez:
Looks like a great start! Can you add documentation on how to run these benchmarks as well please?

rinja_derive_standalone/Cargo.toml
Kijewski: Hehe, yeah. I leave this task to tomorrow-me. :D

GuillaumeGomez: I approve this way of seeing things! :laughing:

GuillaumeGomez:
Awesome, when the `Cargo.toml` features are updated, then let's merge!

GuillaumeGomez:
Just need to fix the merge conflict and it can get merged. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #17, opened 2024-06-18
Don't capture `std::fmt::Error` in `rinja::Error`

Kijewski:
`std::fmt::Error` does not know why it failed, only that it failed, as it has a single value. This PR removes the captured value, to make the code a bit more dense.


GuillaumeGomez (approved):

GuillaumeGomez:
I love such simplifications. :)

PR #18, opened 2024-06-19
Use ISO 646 alternative operators for bit ops + proper error messages for spaces around the `|filter` operator

### LINKED ISSUE #16: Simplify syntax by removing all binary operations (like `|`/`&`/`^`)
It is something I expect to be rarely used in templates and it makes some syntax errors very confusing, like:

```jinja
{% a | indent %}
```

In this case, the only difference between making an `or` binary operation or applying a filter on `a` are the whitespace characters around `|`.

Instead, I suggest to just remove them and eventually add them as filters (I'm not convinced it's worth it though but why not).

I don't include other mathematical operations (like `+`, `/`, `*`) in this removal though as there is no syntax ambiguity in this case as far as I know.

What do you think @Kijewski?

Kijewski:
This change allows simplifying the use of filter expressions, because you won't have to care about spaces around the `|` pipe operator.

Todo:

- [x] update documentation
- [x] cherry-pick https://github.com/djc/askama/pull/979#issuecomment-1983654358 #20 

Resolves #16.


testing/tests/whitespace.rs
```diff
@@ -39,7 +39,7 @@ fn test_extra_whitespace() {
...
+    assert_eq!(template.render().unwrap(), "\n0\n0\n0\n0\n\n\n\n0\n0\n0\n0\n0\n\na0\na1\nvalue\n\n\n\n\n\n[\n  &quot;a0&quot;,\n  &quot;a1&quot;,\n  &quot;a2&quot;,\n  &quot;a3&quot;\n]\n[\n  &quot;a0&quot;,\n  &quot;a1&quot;,\n  &quot;a2&quot;,\n  &quot;a3&quot;\n][\n  &quot;a0&quot;,\n  &quot;a1&quot;,\n  &quot;a2&quot;,\n  &quot;a3&quot;\n]\n[\n  &quot;a1&quot;\n][\n  &quot;a1&quot;\n]\n[\n  &quot;a1&quot;,\n  &quot;a2&quot;\n][\n  &quot;a1&quot;,\n  &quot;a2&quot;\n]\n[\n  &quot;a1&quot;\n][\n  &quot;a1&quot;\n]1-1-1\n3333 3\n2222 2\n0000\n3333\n\ntruefalse\nfalsefalsefalse\n\n\n\n\n\n\n\n\n\n\n\n\n\n");
```
GuillaumeGomez, line 42: The diff here is worrying.

rinja_parser/src/expr.rs
```diff
@@ -192,9 +209,9 @@ impl<'a> Expr<'a> {
     expr_prec_layer!(or, and, "||");
     expr_prec_layer!(and, compare, "&&");
     expr_prec_layer!(compare, bor, "==", "!=", ">=", ">", "<=", "<");
-    expr_prec_layer!(bor, bxor, "|");
-    expr_prec_layer!(bxor, band, "^");
-    expr_prec_layer!(band, shifts, "&");
+    expr_prec_layer_keyword!(bor, bxor, "bitor", "|");
```
GuillaumeGomez, line 212: Thinking out loud for later: would be nice to still parse `^` and `&` to mention to use the full text alternative. For `|`, need to think about it (where to make it checked more exactly).

testing/tests/whitespace.rs
Kijewski: Yeah, I changed the amount of tests in https://github.com/rinja-rs/rinja/pull/18/files#diff-32df128cd4727b02c717b5945bc12d3fb10a618ce4f7dc034e507fb91409cc7eL37-L38. I'll re-add another test to both lines in the template, so the asserted text does not change.

I don't quite like this kind of testing anyway, because it can't be debugged. Either the data is the same or not. It's a blackbox.

GuillaumeGomez: Agreed but it's quite important sadly. It was very useful for me when debugging block filters. If you have a better idea to compare outputs, I'm very interested. :)

rinja_parser/src/lib.rs
```diff
@@ -730,7 +730,7 @@ impl Level {
 #[allow(clippy::type_complexity)]
 fn filter(i: &str, level: Level) -> ParseResult<'_, (&str, Option<Vec<WithSpan<'_, Expr<'_>>>>)> {
     let (i, (_, fname, args)) = tuple((
-        char('|'),
+        ws(char('|')),
```
GuillaumeGomez, line 733: The jinja spec doesn't allow whitespace characters around the `|` character, so not sure if we should. However we could definitely add a super nice error in case there are whitespace characters around it.

rinja_parser/src/expr.rs
Kijewski: Yes, I re-added them.

rinja_parser/src/expr.rs
```diff
@@ -189,15 +189,24 @@ impl<'a> Expr<'a> {
...
+        bor,
+        tag("=="),
+        tag("!="),
+        tag(">="),
+        tag(">"),
+        tag("<="),
+        tag("<")
+    );
+    expr_prec_layer!(bor, bxor, value("|", keyword("bitor")));
+    expr_prec_layer!(bxor, band, alt((tag("^"), value("^", keyword("xor")))));
```
GuillaumeGomez, line 205: Sorry I meant to emit an error in case the character is used instead of the keyword.

testing/templates/allow-whitespaces.html
```diff
@@ -35,7 +35,7 @@
 {{1+2}}{{ 1+2 }}{{ 1 +2 }}{{ 1+ 2 }} {{ 1 + 2 }}
 {{1*2}}{{ 1*2 }}{{ 1 *2 }}{{ 1* 2 }} {{ 1 * 2 }}
 {{1&2}}{{ 1&2 }}{{ 1 &2 }}{{ 1& 2 }} {{ 1 & 2 }}
-{{1|2}}{{ 1|2 }}{{ 1 |2 }}{{ 1| 2 }} {{ 1 | 2 }}
+{{1 bitor 2}}{{ 1 bitor 2 }}{{ 1 bitor 2}}{{1 bitor 2 }} {{1			bitor			2}}
```
GuillaumeGomez, line 38: Why this weird tab change? Wouldn't adding whitespace characters before `1` and after `2` provide the expected result?

Kijewski: I had no idea what to do in the last test. :) If I remove it, [testing/tests/whitespace.rs:42](https://github.com/rinja-rs/rinja/pull/18/files#diff-b63937165590fc2719d8f31a1554a4e4b98c56f5dece3bd96b73fe7827a3f629) changes.

rinja_parser/src/expr.rs
Kijewski: You're totally right, changed.

rinja_parser/src/lib.rs
Kijewski: I added a hopefully useful error message for this case.

rinja_parser/src/lib.rs
```diff
@@ -727,14 +727,23 @@ impl Level {
...
+    let (j, _) = pair(char('|'), not(char('|')))(j)?;
+
+    if !had_spaces {
+        cut(pair(
+            ws(identifier),
+            opt(|i| Expr::arguments(i, level, false)),
+        ))(j)
+    } else {
+        Err(nom::Err::Failure(ErrorContext::new(
+            "The filter operator `|` must not be preceeded by any whitespace characters.\n\
```
GuillaumeGomez, line 742: Little nit: other error messages don't start with a capital letter.

rinja_parser/src/expr.rs
```diff
@@ -325,6 +308,33 @@ impl<'a> Expr<'a> {
     }
 }
 
+fn token_xor(i: &str) -> ParseResult<'_> {
+    let (i, good) = alt((value(true, keyword("xor")), value(false, char('^'))))(i)?;
+    if good {
+        Ok((i, "^"))
+    } else {
+        Err(nom::Err::Failure(ErrorContext::new(
+            "the binary XOR operator is called 'xor' in rinja",
```
GuillaumeGomez, line 317: Just noticed but we use backtick to "highlight" elements like `xor` in the error messages, not quotes. Sorry, nitpicking time. ^^'

rinja_parser/src/expr.rs
```diff
@@ -325,6 +308,33 @@ impl<'a> Expr<'a> {
...
+fn token_bitand(i: &str) -> ParseResult<'_> {
+    let (i, good) = alt((
+        value(true, keyword("bitand")),
+        value(false, pair(char('&'), not(char('&')))),
+    ))(i)?;
+    if good {
+        Ok((i, "&"))
+    } else {
+        Err(nom::Err::Failure(ErrorContext::new(
+            "the binary AND operator is called 'bitand' in rinja",
```
GuillaumeGomez, line 332: Same here.

rinja_parser/src/lib.rs
```diff
@@ -727,14 +727,23 @@ impl Level {
...
+            The binary OR operator is called `bitor` in rinja.",
```
GuillaumeGomez, line 743: You did it right, here. :)

testing/templates/allow-whitespaces.html
GuillaumeGomez: Oh well, let's keep it as is for now, I'll try to take a look later on.

rinja_parser/src/lib.rs
Kijewski: I was not sure about that, because it's two sentences. Maybe I should join them with a ";", so an initial lower cases character makes more sense? :D

testing/tests/ui/iso646.rs
```diff
@@ -0,0 +1,46 @@
+use rinja::Template;
```
GuillaumeGomez, line 1: Thanks for adding the ui tests for error output!

GuillaumeGomez:
So remains nitpicking and the docs. Do you me want to handle the cherry-picking?

rinja_parser/src/expr.rs
Kijewski: Nitpicking is half the fun of reviewing. :D Fixed.

rinja_parser/src/expr.rs
Kijewski: Also fixed.

rinja_parser/src/lib.rs
Kijewski: Yea me. :D

testing/tests/ui/iso646.rs
Kijewski: I love UI tests, and I love that it's so simple to do them in rust. In most languages you only test the "good" outcomes, but the bad outcomes are just as important.

Kijewski:
> So remains nitpicking and the docs. Do you me want to handle the cherry-picking?

Yes, I'd be thankful if you do it.

GuillaumeGomez:
Say no more, I'm on it!

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #20, opened 2024-06-19
Limit nested filters

GuillaumeGomez:
Here you go (applied your suggestion from your comment as well).


rinja_parser/src/expr.rs
```diff
@@ -199,27 +198,27 @@ impl<'a> Expr<'a> {
...
+    fn filtered(i: &'a str, mut level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
+        #[allow(clippy::type_complexity)]
+        fn filter<'a>(
+            i: &'a str,
+            level: &mut Level,
+        ) -> ParseResult<'a, (&'a str, Option<Vec<WithSpan<'a, Expr<'a>>>>)> {
+            let (i, _) = char('|')(i)?;
+            *level = level.nest(i)?.1;
+            pair(ws(identifier), opt(|i| Expr::arguments(i, *level, false)))(i)
         }
```
Kijewski, line 210: This nested function is the same as [rinja_parser/src/lib.rs](https://github.com/rinja-rs/rinja/pull/20/files#diff-d7d4a4880df4695f086349de84bd46afd2546635da8c8c7480abe26933c0a82f), isn't it? I think you can delete it, then.

GuillaumeGomez: Indeed! Removed it and used the one from `src/lib.rs` instead.

Kijewski (approved):
LGTM, thank you!

PR #21, opened 2024-06-20
 Only use explicit features 

Kijewski:
* Only use explicit features
* serde-json is actually called serde_json
* Remove deprecated no-op features


GuillaumeGomez (approved):

GuillaumeGomez:
Much clearer this way, thanks!

PR #23, opened 2024-06-20
Deny all rust keywords as macro names

Kijewski:
Per <https://github.com/djc/askama/pull/1044#discussion_r1602289767>.

### DIFF
--- a/rinja_parser/src/node.rs
+++ b/rinja_parser/src/node.rs
@@ -592,9 +592,9 @@ impl<'a> Macro<'a> {
             ))),
         ));
         let (j, (pws1, _, (name, params, nws1, _))) = start(i)?;
-        if name == "super" {
+        if is_rust_keyword(name) {
             return Err(nom::Err::Failure(ErrorContext::new(
-                "'super' is not a valid name for a macro",
+                format!("'{name}' is not a valid name for a macro"),
                 i,
             )));
         }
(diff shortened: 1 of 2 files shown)


rinja_parser/src/node.rs
```diff
@@ -1202,3 +1202,99 @@ impl<'a> Comment<'a> {
...
+        *b"typeof__",
+        *b"unsafe__",
+    ];
+    const KW7: &[[u8; MAX_KW_LEN]] = &[*b"unsized_", *b"virtual_"];
+    const KW8: &[[u8; MAX_KW_LEN]] = &[*b"abstract", *b"continue", *b"override"];
+
+    const KWS: &[&[[u8; MAX_KW_LEN]]] = &[KW0, KW1, KW2, KW3, KW4, KW5, KW6, KW7, KW8];
+
+    // Ensure that all strings are ASCII, because we use `from_utf8_unchecked()` further down.
+    const _: () = {
```
GuillaumeGomez, line 1274: Instead of making it a const check at every compilation, wouldn't it better to have it in a `#[test]`? You can add a test inside this function directly to use the exact same code.

GuillaumeGomez:
Please add a ui test too please.

rinja_parser/src/node.rs
Kijewski: Done.

Kijewski:
> Please add a ui test too please.

There always is a test for super: [macro-super.rs](https://github.com/rinja-rs/rinja/blob/master/testing/tests/ui/macro-super.rs). Should I add tests for the other keywords, too?

rinja_parser/src/node.rs
```diff
@@ -1268,25 +1270,22 @@ fn is_rust_keyword(ident: &str) -> bool {
...
-                    assert!(KWS[i][j][k].is_ascii());
-                    k += 1;
-                }
-                j += 1;
+    &[KW0, KW1, KW2, KW3, KW4, KW5, KW6, KW7, KW8]
+};
+
+/// Ensure that all strings are UTF-8, because we use `from_utf8_unchecked()` further down.
+#[test]
+fn ensure_utf8() {
```
GuillaumeGomez, line 1278: Is there a need to duplicate these consts between the two crates? Can't we make them public behind `#[doc(hidden)]` instead?

GuillaumeGomez:
> > Please add a ui test too please.
> 
> There always is a test for super: [macro-super.rs](https://github.com/rinja-rs/rinja/blob/master/testing/tests/ui/macro-super.rs). Should I add tests for the other keywords, too?

Oh if there is one already then it's fine. :+1:

rinja_parser/src/node.rs
Kijewski: > Is there a need to duplicate these consts between the two crates?

Sadly, I think yes. The list items are not the same with `loop self Self` missing from the generator list.

GuillaumeGomez: Then why not add another list for these 3 keywords in `rinja_parser`? Having some much code duplicated is itching me strongly. ^^'

If you want I don't mind doing it. In this case just merge this PR and I'll take a look.

GuillaumeGomez (approved):

rinja_parser/src/node.rs
Kijewski: > In this case just merge this PR and I'll take a look.

Thanks! Sure, go ahead. :)

PR #24, opened 2024-06-20
Remove duplicated code to detect Rust keywords/identifiers

GuillaumeGomez:
Follow-up of https://github.com/rinja-rs/rinja/pull/23.

cc @Kijewski

### DIFF
--- a/rinja_parser/src/node.rs
+++ b/rinja_parser/src/node.rs
@@ -1203,97 +1203,130 @@ impl<'a> Comment<'a> {
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct Ws(pub Option<Whitespace>, pub Option<Whitespace>);
 
-const MAX_KW_LEN: usize = 8;
-const KWS: &[&[[u8; MAX_KW_LEN]]] = {
+#[doc(hidden)]
+pub const MAX_KW_LEN: usize = 8;
+const MAX_REPL_LEN: usize = MAX_KW_LEN + 2;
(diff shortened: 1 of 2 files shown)


rinja_parser/src/node.rs
```diff
@@ -1203,97 +1203,130 @@ impl<'a> Comment<'a> {
...
 fn is_rust_keyword(ident: &str) -> bool {
+    fn is_rust_keyword_inner(
+        kws: &[&[[u8; MAX_REPL_LEN]]],
+        padded_ident: &[u8; MAX_KW_LEN],
+        ident_len: usize,
+    ) -> bool {
+        // Since the individual buckets are quite short, a linear search is faster than a binary search.
+        kws[ident_len]
+            .iter()
+            .any(|&probe| padded_ident == &probe[2..])
```
Kijewski, line 1289: Interesting! You are right, the

```rust
padded_ident == <[u8; parser::node::MAX_KW_LEN]>::try_from(&probe[2..]).unwrap()
```
"optimization" is not needed anymore, and now the same code generated with and without it. When I first wrote the search, it was still needed because otherwise the assembly would needlessly contain length checks.

Kijewski (approved):
Thanks you!

rinja_parser/src/node.rs
GuillaumeGomez: Oh interesting. I checked and couldn't find out why it was needed. I suppose rustc/LLVM got better.

PR #26, opened 2024-06-22
 derive: add librustdoc benchmarks 

Kijewski:


Kijewski:
I've just realized that we already have a handful of files that we can use for benchmarking, oops.

GuillaumeGomez (approved):

GuillaumeGomez:
Convenient. Thanks!

PR #27, opened 2024-06-23
Update recursion ui error

manunio:
I think previous input for filter-recursion ui test was not right as it was not triggering error for recursion itself, for example:

```rust
fn fuzzed_filter_recursion() {
    const TEMPLATE: &str = include_str!("../tests/filter-recursion.txt");
    if let Err(e) = Ast::from_str(TEMPLATE, None, &Syntax::default()) {
        panic!("{e}");
    }
}
```

```sh
---- tests::fuzzed_filter_recursion stdout ----
thread 'tests::fuzzed_filter_recursion' panicked at rinja_parser/src/tests.rs:1121:9:
failed to parse template source at row 1, column 255 near:
"|A|AA|A|A|A|A|AA|A|A|A|A|AA|A|A|A|A|AA|A"...
```


testing/tests/ui/filter-recursion.stderr
```diff
@@ -1,6 +1,6 @@
-error: unclosed expression, missing "}}"
-       failed to parse template source at row 1, column 0 near:
-       "{{ s|a|a|a|a|a|a|a|A|a|a|a|a|a|a|a|a|a|a"...
+error: failed to parse template source
```
GuillaumeGomez, line 1: Thinking out loud but the error is not great. It should indicate that it hits the recursion limit. Gonna open an issue for that.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #28, opened 2024-06-23
Improve recursion limit error


GuillaumeGomez:
Currently it displays:

```
error: failed to parse template source
```

It should instead be (or something along the line):

```
error: hit recursion limit
```

Follow-up of https://github.com/rinja-rs/rinja/pull/27.

PR #29, opened 2024-06-23
Improve error message in case too much nesting

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/28.

### DIFF
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -711,10 +711,10 @@ pub(crate) struct Level(u8);
 impl Level {
     fn nest(self, i: &str) -> ParseResult<'_, Level> {
         if self.0 >= Self::MAX_DEPTH {
-            return Err(ErrorContext::from_err(nom::Err::Failure(error_position!(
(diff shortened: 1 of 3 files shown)


rinja_parser/src/lib.rs
```diff
@@ -711,10 +711,10 @@ pub(crate) struct Level(u8);
 impl Level {
     fn nest(self, i: &str) -> ParseResult<'_, Level> {
         if self.0 >= Self::MAX_DEPTH {
-            return Err(ErrorContext::from_err(nom::Err::Failure(error_position!(
+            return Err(nom::Err::Failure(ErrorContext::new(
+                "reached nested depth limit",
```
Kijewski, line 715: Maybe something like `"your template code is too deeply nested, or last expression is too complex"`?

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Updated!

Kijewski (approved):
Yep, much easier to understand. :)

Issue #30, opened 2024-06-23
`Level` API is error-prone


GuillaumeGomez:
In many cases, we're supposed to call `.leave()` on it, which is definitely not great. Having it done with a closure which would handle the `leave` part automatically would be much better.

PR #31, opened 2024-06-23
Improve Level API a bit

GuillaumeGomez:
Part of https://github.com/rinja-rs/rinja/issues/30.

It already improves the situation a bit. Not sure if the rest is worth improving or not though.

### DIFF
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -635,14 +635,17 @@ impl<'a> State<'a> {
         }
     }
 
-    fn nest<'b>(&self, i: &'b str) -> ParseResult<'b, ()> {
-        let (_, level) = self.level.get().nest(i)?;
+    fn nest<'b, T, F: FnOnce(&'b str) -> ParseResult<'b, T>>(
+        &self,
+        i: &'b str,
+        callback: F,
+    ) -> ParseResult<'b, T> {
+        let prev_level = self.level.get();
+        let (_, level) = prev_level.nest(i)?;
         self.level.set(level);
-        Ok((i, ()))
-    }
-
-    fn leave(&self) {
-        self.level.set(self.level.get().leave());
+        let ret = callback(i);
+        self.level.set(prev_level);
+        ret
     }
 
     fn tag_block_start<'i>(&self, i: &'i str) -> ParseResult<'i> {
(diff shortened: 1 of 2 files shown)


rinja_parser/src/lib.rs
```diff
@@ -635,14 +635,17 @@ impl<'a> State<'a> {
         }
     }
 
-    fn nest<'b>(&self, i: &'b str) -> ParseResult<'b, ()> {
+    fn nest<'b, T, F: FnMut(&'b str) -> ParseResult<'b, T>>(
+        &self,
+        i: &'b str,
+        mut callback: F,
```
Kijewski, line 641: `F` can be `FnOnce`. Currently that does not matter, but maybe we refactor the parser code later.

rinja_parser/src/lib.rs
```diff
@@ -635,14 +635,17 @@ impl<'a> State<'a> {
...
+    ) -> ParseResult<'b, T> {
+        let prev_level = self.level.get();
         let (_, level) = self.level.get().nest(i)?;
```
Kijewski, line 644: You could directly call `prev_level.nest(i)`.

rinja_parser/src/node.rs
```diff
@@ -90,10 +90,7 @@ impl<'a> Node<'a> {
             }
         };
 
-        let (i, _) = s.nest(j)?;
-        let result = func(i, s);
-        s.leave();
-        let (i, node) = result?;
```
Kijewski, line 96: It is much clearer what is going on with you change, and the code does not much more `nom`-like. Clear advancement!

Kijewski (changes requested):

rinja_parser/src/lib.rs
GuillaumeGomez: Good point!

rinja_parser/src/lib.rs
GuillaumeGomez: Indeed!

GuillaumeGomez:
There is still the original `Level`. I'm still wondering if `nest` should return a non-`Copy` type which would contain a mutable reference to the calling `Level` or something along the line. Really not clear about it...

GuillaumeGomez:
Fixed CI. :)

Kijewski (approved):

PR #32, opened 2024-06-24
Make JSON prettifying optional

Kijewski:
This PR adds an optional argument to the `|tojson` filter, which controls if the serialized JSON data gets prettified or not. The arguments works the same as flask's [`|tojson`][flask] filter, which passes the argument to python's [`json.dumps()`][python]:

* Omitting the argument, providing a negative integer, or `None`, then compact JSON data is generated.
* Providing a non-negative integer, then this amount of ASCII spaces is used to indent the data. (Capped to 16 characters.)
* Providing a string, then this string is used as prefix. I attempts are made to ensure that the prefix actually consists of whitespaces, because chances are, that if you provide e.g. `&nsbp;`, then you are doing it intentionally.

This is a breaking change, because it changes the default behavior to not prettify the data. This is done intentionally, because this is how it works in flask.

[flask]: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.tojson
[python]: https://docs.python.org/3/library/json.html#json.dump

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -1364,14 +1364,14 @@ impl<'a> Generator<'a> {
             ));
         }
 
-        buf.write(CRATE);
-        buf.write("::filters::json(");
-        self._visit_args(ctx, buf, args)?;
-        match args.len() {
-            1 => buf.write(", ()"),
-            2 => {}
+        let filter = match args.len() {
+            1 => "json",
+            2 => "json_pretty",
             _ => return Err(ctx.generate_error("unexpected argument(s) in `json` filter", node)),
-        }
+        };
+
+        buf.write(format_args!("{CRATE}::filters::{filter}("));
+        self._visit_args(ctx, buf, args)?;
         buf.write(")?");
         Ok(DisplayWrap::Unwrapped)
     }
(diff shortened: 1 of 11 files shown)


rinja_derive/src/generator.rs
```diff
@@ -1364,12 +1364,14 @@ impl<'a> Generator<'a> {
...
         }
 
-        if args.len() != 1 {
-            return Err(ctx.generate_error("unexpected argument(s) in `json` filter", node));
-        }
         buf.write(CRATE);
         buf.write("::filters::json(");
         self._visit_args(ctx, buf, args)?;
+        match args.len() {
+            1 => buf.write(", ()"),
```
GuillaumeGomez, line 1371: Instead of adding a trait to handle integer/void, why not write a specialized json filter called `json_empty`?

Second question: why not forcing the type of `indent` to be a `usize` so we don't have to handle types and other things? I find it a bit strange to allow users to pass a negative number of even `None` as arguments.

GuillaumeGomez (dismissed):

GuillaumeGomez:
Looks much simpler now, thanks! Please feel free to merge once CI pass.

PR #37, opened 2024-07-01
Favor methods over closure fields

GuillaumeGomez:
Very surprisingly, we can't call methods directly, however closures (which are fields) can be called. I don't think this incentive makes sense as methods are much more common than closures as fields. Also, it now follows the rust syntax way.

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -245,13 +245,11 @@ struct MyTemplate;
 
 ### Struct / trait implementations
 
-Finally, we can invoke functions that are implementation methods of our
-template struct, by referencing `Self` (note the uppercase `S`) as the path,
-before calling our function:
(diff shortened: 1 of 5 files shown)


GuillaumeGomez:
Hopefully, not too many fixes like this should follow, still working on switching docs.rs from `tera`. :3

Kijewski:
To call a function in the current scope, you have to call `{{ self::function(args) }}` then, don't you? Maybe you could add a test to ensure this syntax works, and maybe mention it in he book?

GuillaumeGomez:
Should be the same as the current situation but having more tests is always a good idea!

GuillaumeGomez:
Added test and updated docs (which were not mentioning functions surprisingly).

Kijewski (approved):
Thank you!

PR #39, opened 2024-07-01
Allow to use `*` and `&` in expressions

GuillaumeGomez:
Quite useful. I'm currently blocked because it's missing and `as_ref`/`deref` filters cannot be used in this context.


Kijewski (approved):
Thank you!

PR #40, opened 2024-07-01
Rename `as_ref` filter into `ref`

GuillaumeGomez:
It's not calling the `AsRef` trait and I'd prefer to clear any potential confusion and to follow the same nomenclature as `deref`.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -22,7 +22,6 @@ Enable it with Cargo features (see below for more information).
 * **[Built-in filters][#built-in-filters]:**  
 
   * [`abs`][#abs]
-  * [`as_ref`][#as_ref]
   * [`capitalize`][#capitalize]
   * [`center`][#center]
   * [`deref`][#deref]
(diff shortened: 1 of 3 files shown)


Kijewski:
I think it might be best to simply remove `(as_)ref` and `deref`, what do you think? In #39 you introduce the prefix operators `*` and `&`, and I like these operators much better than filter. Esp. since the filters are kinda magic: You couldn't implement the yourself when you write your own filters.

GuillaumeGomez:
Unfortunately I think these two filters remain useful in filters chaining. For example if you have a filter `to_u32` which expects a `&str`, then you need to do:

```jinja
{{ "{{}}"|format(a)|ref|to_str }}
```

This example is easy to fix without these filters, but some more advanced might not. Allowing users to go around this limitation still seems pertinent to me.

Kijewski:
Ah, yes, you're right. In a filter chain it is useful!

Kijewski (approved):

PR #42, opened 2024-07-02
Generate better errors for top level nodes

GuillaumeGomez:


rinja_derive/src/heritage.rs
```diff
@@ -91,9 +96,28 @@ impl Context<'_> {
                         let path = config.find_template(import.path, Some(path))?;
                         imports.insert(import.scope, path);
                     }
-                    Node::Extends(_) | Node::Macro(_) | Node::Import(_) if !top => {
-                        return Err(CompileError::no_file_info(
-                            "extends, macro or import blocks not allowed below top level",
+                    Node::Extends(e) if !top => {
```
GuillaumeGomez, line 99: I'm a bit sad about needing to split this in 3 parts (because the `Node` isn't the same, so rust isn't happy). But we got a small improvement for this: the error message is now specific to the node type.

softdevca: I feel the same.  A `name()` on node, optionally with a `top_level_only()`, might reduce the specialization.

The bigger problem I find using Rinja (or Askama) is the lack of context in the error messages for templates. It's nice to know what specific keyword caused the problem but it would be even better if it showed a template line number.

GuillaumeGomez: It's exactly what this PR adds. Take a look at the error output [here](https://github.com/rinja-rs/rinja/pull/42/files#diff-3de8e2fe64619a6cbe36d7120211f8c3b211198bc3736605eac04d98a03714a0R2-R3). Almost all errors now have context and everything minus a few exceptions. :)

testing/tests/ui/blocks_below_top_level.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[derive(Template)]
+#[template(source = r#"
+{% block bla %}
+{% import "bla.txt" as blue %}
+{% endblock %}
+"#, ext = "txt")]
+struct MyTemplate3;
+
+#[derive(Template)]
```
softdevca, line 28: This might not be the best location for this test as it checks multiple `extend`s rather than blocks not at the top level

rinja_derive/src/heritage.rs
softdevca: Very helpful, thank you!

testing/tests/ui/blocks_below_top_level.rs
GuillaumeGomez: Fair enough, moving it to its own file.

softdevca (approved):

rinja_derive/src/heritage.rs
GuillaumeGomez: Getting a review from @Kijewski. For now we're two maintainers and we follow the rule to wait for at least one other maintainer's review before merging. So just a bit of patience now. :wink:

Kijewski:
Very useful change!

I de-duplicated the top-level check, and getting the FileInfo of a Node a bit.

Kijewski (approved):

GuillaumeGomez:
I can't approve it says github but I like the changes a lot, thanks to both of you!

PR #44, opened 2024-07-02
`strip_common` test was failing

softdevca:
The test for `strip_common` was failing on Linux when the tests were run from somewhere other than `/home/*`.  

The same test can be performed without the check for Linux or for requiring a specific directory by checking a path that can always be canonicalized.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #45, opened 2024-07-04
Add `.rustfmt.toml` configuration

Kijewski:
This PR configures the formatting with a few defaults that aid readability, in my opinion. One drawback of adding this file is that rustfmt uses unstable features, now, and you have to use nightly to run it:

```sh
cargo +nightly fmt --all
```

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,12 @@
+edition = "2021"
+group_imports = "StdExternalCrate"
+imports_granularity = "Module"
+newline_style = "Unix"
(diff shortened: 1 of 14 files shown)


Kijewski:
This PR is quite opinionated. It's totally OK if you don't like it or if you would use other settings. :)

GuillaumeGomez (approved):

GuillaumeGomez:
I like it, just hope it won't make things too difficult for contributors. Well, we'll see. Thanks!

PR #46, opened 2024-07-04
 Remove a `#[allow(clippy::match_wild_err_arm)]`  and  dedup error message creation 

Kijewski:


rinja_derive/src/config.rs
```diff
@@ -298,38 +298,24 @@ where
...
-            Ok(source)
+) -> Result<String, CompileError> {
+    if let Ok(mut source) = fs::read_to_string(tpl_path) {
+        if source.ends_with('\n') {
+            let _ = source.pop();
         }
+        Ok(source)
+    } else {
+        let msg = format!(
+            "unable to open template file '{}'",
```
GuillaumeGomez, line 312: I think we should provide the `Err()` value in the error message, it's not always obvious what's wrong (like wrong rights).

Kijewski: Good call! Added.

testing/tests/ui/include-a-folder.stderr
```diff
@@ -0,0 +1,10 @@
+error: unable to open template file '$WORKSPACE/target/tests/trybuild/rinja_testing/templates/a_file_that_is_actually_a_folder.html':
+       Is a directory (os error 21)
```
GuillaumeGomez, line 2: Hum... I think having the error text on the same line might be better. What do you think?

PS: Having a test for this is great, thanks!

Kijewski: I wasn't sure if the line does not become too long, then. But chances are that you aren't using such a long path in a real project. :)

GuillaumeGomez: Boarf, I think it's fine. I find it less confusing this way.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #49, opened 2024-07-04
Save template AST across proc-macro calls


GuillaumeGomez:
Currently the AST of parsed templates is only saved in the context of one attribute. It'd likely improve compilation to save them across all calls instead.

Kijewski:
Implemented in #59.

PR #50, opened 2024-07-05
derive: Refactor a few types (less cloning, more re-usability)

Kijewski:
Minor optimizations and pre-requisites needed if we want to implement #49:

* Implement `Hash + PartialEq` for more types
* Use `Cow<str>` instead of `String` if possible
* Don't calculate HashMap entry hashes twice

The derive benchmark runs 3 to 5% faster across the board.

### DIFF
--- a/rinja_derive/src/input.rs
+++ b/rinja_derive/src/input.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-use std::collections::hash_map::HashMap;
+use std::collections::hash_map::{Entry, HashMap};
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


rinja_derive/src/input.rs
```diff
@@ -128,16 +129,17 @@ impl TemplateInput<'_> {
             let mut nested = vec![parsed.nodes()];
             while let Some(nodes) = nested.pop() {
                 for n in nodes {
-                    let mut add_to_check = |new_path: Rc<Path>| -> Result<(), CompileError> {
-                        if !map.contains_key(&new_path) {
+                    let mut add_to_check = |new_path| -> Result<(), CompileError> {
```
GuillaumeGomez, line 132: Is it because the type changed or because the annotation is unnecessary? If it's the later, could you please keep the annotation? It makes it easier to answer the question "why so many `clone`?".

GuillaumeGomez:
Just one nit, otherwise looks good to me, great work!

rinja_derive/src/input.rs
Kijewski: No, the change wasn't needed. I first removed both the argument and return type, but I had to re-add the return type.

I re-added the argument type, too.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #51, opened 2024-07-05
Make it possible to return "safe" content from filters


GuillaumeGomez:
We could for example add an enum to ensure that, for example:

```rust
enum NeedsEscape<T> {
    Yes(T),
    No(T),
}
```

Where `NeedsEscape::Yes` would be the default. Then it'd just be a matter of implementing `From<T: Display> for NeedsEscape` to keep it working for most cases. And in case someone want to add a filter that returns content that should not be escaped, they can.

What do you think @Kijewski ?

PR #53, opened 2024-07-07
Remove `rinja_escape` and make `|escape` a "normal" filter

Kijewski:
`rinja_escaped` was a somewhat odd beast: Some functions and types were unused, some implementation details were open for the world to see, and it was checked at runtime if a value was safe when it could be deduced at compile type. Now `|escape` and `|safe` are proper filters with nothing unusual about them, and the additional crate is gone.

This PR is a first step towards #51.


Kijewski:
```text
$ cd testing
$ cargo bench --bench all
Big table               time:   [345.95 µs 348.52 µs 351.07 µs]
                        change: [-6.9603% -4.5108% -2.0227%] (p = 0.00 < 0.05)
                        Performance has improved.

Teams                   time:   [479.11 ns 480.44 ns 481.86 ns]
                        change: [-9.7181% -8.6280% -7.9141%] (p = 0.00 < 0.05)
                        Performance has improved.
```

rinja/src/filters/escape.rs
```diff
@@ -0,0 +1,149 @@
+use std::convert::Infallible;
+use std::fmt::{self, Display, Formatter, Write};
+use std::str;
+
+/// Marks a string (or other `Display` type) as safe
+///
+/// Use this is you want to allow markup in an expression, or if you know
```
GuillaumeGomez, line 7:
```suggestion
/// Use this if you want to allow markup in an expression, or if you know
```

GuillaumeGomez:
Looks all good to me, just one thing: could you add a test an invalid escaper please? I didn't see one. Would be also nice to improve the error message from `invalid escaper for escape filter` to ``invalid escaper `{provided_escaper_name}` for `escape` filter``.

GuillaumeGomez:
Same for `no escaper defined for extension '{escaping}'`.

Kijewski:
> Looks all good to me, just one thing: could you add a test an invalid escaper please? 

Sure! I updated the error message to contain the unknown escaper, and added a UI test.

GuillaumeGomez:
CI needs to be updated too.

testing/tests/ui/no-such-escaper.stderr
```diff
@@ -0,0 +1,17 @@
...
+ --> LocalEscaper.html:1:38
+       "text|escape(\"latex\")}}`."
+ --> tests/ui/no-such-escaper.rs:3:10
+  |
+3 | #[derive(Template)]
+  |          ^^^^^^^^
+  |
+  = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: no escaper defined for extension 'tex'
```
GuillaumeGomez, line 11: Thinking more about this error message, I don't think it makes much sense, especially for users who just did a typo. What do you think about `unknown extension 'tex'. The available extensions are 'x', 'y', 'z'`. What do you think?

Kijewski:
> CI needs to be updated too.

It is, but I failed to `cargo fmt` the source. :)

testing/tests/ui/no-such-escaper.stderr
Kijewski: Good idea! Added.

rinja_derive/src/lib.rs
```diff
@@ -231,6 +232,30 @@ impl fmt::Display for FileInfo<'_> {
...
+        let exts: BTreeSet<&str> = self
+            .0
+            .iter()
+            .flat_map(|(exts, _)| exts)
+            .map(|x| x.as_ref())
+            .collect();
+        f.write_str("The available extensions are:")?;
+        for (i, mut e) in exts.iter().copied().enumerate() {
+            if e.is_empty() {
+                e = r#""" <empty string>"#;
```
GuillaumeGomez, line 248: I thought it was a debug display before checking. XD

I think using parens would be better here:

```suggestion
                e = r#""" (empty string)"#;
```

One extra thought: wouldn't it be better to rely on `join` instead? So your first iterator could become:

```rust
let exts: String = self
            .0
            .iter()
            .flat_map(|(exts, _)| exts)
            .map(|x| if x.is_empty() { r#""" (empty string)"# else { x.as_ref() })
            .collect::<BTreeSet<&str>>()
            .into_iter()
            .join(", ");
```

No need for the loop below like this. And I think perf is not really a big issue here since we're errorring.

Kijewski: You're right. :) I use `slice::join` and the debug display of the extension, now.

rinja_derive/src/lib.rs
```diff
@@ -231,6 +232,21 @@ impl fmt::Display for FileInfo<'_> {
...
+struct MsgValidEscapers<'a>(&'a [(Vec<Cow<'a, str>>, Cow<'a, str>)]);
+
+impl fmt::Display for MsgValidEscapers<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut exts = self
+            .0
+            .iter()
+            .flat_map(|(exts, _)| exts)
+            .map(|x| format!("{x:?}"))
+            .collect::<Vec<_>>();
```
GuillaumeGomez, line 244: Even better. :+1:

GuillaumeGomez (approved):

GuillaumeGomez:
Once merge conflict is fixed, please merge.

PR #54, opened 2024-07-07
Add HTML safe types

### LINKED ISSUE #51: Make it possible to return "safe" content from filters

Kijewski:
This PR adds an `HtmlSafeMarker` trait. Types that implement this marker are know to never generate strings containing the characters `< > & " '`, so they don't have to be escaped.

All glory goes to \@dtolnay's ["Autoref-based stable specialization"][1] case study / blog entry.

[1]: <https://github.com/dtolnay/case-studies/blob/0a9f083f334e53bc854a80022b1984b1bae36ef6/autoref-specialization/README.md>

This PR is built on #53.

Resolves #51.

### DIFF
--- a/rinja/src/filters/escape.rs
+++ b/rinja/src/filters/escape.rs
@@ -386,6 +386,16 @@ const _: () = {
     }
 };
 
+/// Like [`Safe`], but only for HTML output
+pub struct HtmlSafeOutput<T: fmt::Display>(pub T);
+
+impl<T: fmt::Display> fmt::Display for HtmlSafeOutput<T> {
+    #[inline]
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
 macro_rules! mark_html_safe {
     ($($ty:ty),* $(,)?) => {$(
         impl HtmlSafe for $ty {}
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
I'll review this one once #53 is merged if you don't mind. ;)

Kijewski:
Yeah, I guess the usage was not entirely self-explanatory. :D

Added a section to the book. I also renamed `HtmlSafeMarker` into `HtmlSafe`.

Kijewski:
I added more tests, and replaced `MaybeSafe::safe: bool` with `MaybeSafe::needs_escaping: bool`. A feature that might be dangerous if used wrong should be hard to misunderstand.

rinja/src/filters/escape.rs
```diff
@@ -160,6 +161,276 @@ pub trait Escaper: Copy {
...
+/// assert_eq!(
+///     DivWithBackdoor { klass: "<script>", enable_backdoor: false }.to_string(),
+///     "<div class='&#60;script&#62;'></div>",
+/// );
+/// assert_eq!(
+///     DivWithBackdoor { klass: "<script>", enable_backdoor: true }.to_string(),
+///     "<div class='<script>'></div>",
+/// );
+/// ```
+pub struct MaybeSafe<T: fmt::Display> {
```
GuillaumeGomez, line 263: I think this type would be better as an enum (with variables `Safe` and `NeedsEscaping`?).

Kijewski: Yes, you're right. :+1: It's less error prone. And actually its easier to use I just noticed while implementing the change.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks! Please merge once CI is happy.

PR #55, opened 2024-07-07
 Speed-up HTML escaping by using equal sized entities 

Kijewski:
By using codepoint entities like `'&'` → `"&#38;"`, we have a much smaller lookup table (58 bytes instead of 29× pointer size ~= 232 bytes). This makes the cache happy, and the benchmark run about ~20% faster.

```text
$ cargo bench --bench escape
Escaping                time:   [3.4087 µs 3.4126 µs 3.4168 µs]
                        change: [-19.790% -19.580% -19.354%] (p = 0.00 < 0.05)
                        Performance has improved.
```

This PR is based on #53.

TODO:

- [x] fix all the expected outputs

### DIFF
--- a/rinja/src/filters/escape.rs
+++ b/rinja/src/filters/escape.rs
@@ -1,5 +1,6 @@
 use std::convert::Infallible;
 use std::fmt::{self, Display, Formatter, Write};
+use std::num::NonZeroU8;
 use std::str;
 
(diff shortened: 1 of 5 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
That's great, thanks!

PR #56, opened 2024-07-09
Fix big oof in the book :)

Kijewski:


GuillaumeGomez (approved):

GuillaumeGomez:
You said it was a very bad example, so I went along with it. 🤣

PR #59, opened 2024-07-09
Cache parsing templates

Kijewski:
It compiles, tests work just fine, but I did not benchmark yet if the caching is of any use performance-wise.

### DIFF
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::path::{Path, PathBuf};
-use std::rc::Rc;
+use std::sync::Arc;
 use std::{env, fs};
 
 use parser::node::Whitespace;
@@ -120,7 +120,7 @@ impl<'a> Config<'a> {
         &self,
         path: &str,
         start_at: Option<&Path>,
-    ) -> std::result::Result<Rc<Path>, CompileError> {
+    ) -> std::result::Result<Arc<Path>, CompileError> {
         if let Some(root) = start_at {
             let relative = root.with_file_name(path);
             if relative.exists() {
@@ -294,8 +294,8 @@ fn str_set<'a>(vals: &[&'a str]) -> Vec<Cow<'a, str>> {
 
 pub(crate) fn get_template_source(
     tpl_path: &Path,
-    import_from: Option<(&Rc<Path>, &str, &str)>,
-) -> Result<Rc<str>, CompileError> {
+    import_from: Option<(&Arc<Path>, &str, &str)>,
+) -> Result<Arc<str>, CompileError> {
     match fs::read_to_string(tpl_path) {
         Ok(mut source) => {
             if source.ends_with('\n') {
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -361,30 +361,6 @@ fn str_set<'a>(vals: &[&'a str]) -> Vec<Cow<'a, str>> {
     vals.iter().copied().map(Cow::from).collect()
 }
 
-pub(crate) fn get_template_source(
-    tpl_path: &Path,
-    import_from: Option<(&Arc<Path>, &str, &str)>,
-) -> Result<Arc<str>, CompileError> {
-    match fs::read_to_string(tpl_path) {
-        Ok(mut source) => {
-            if source.ends_with('\n') {
-                let _ = source.pop();
-            }
-            Ok(source.into())
-        }
-        Err(err) => {
-            let msg = format!(
-                "unable to open template file '{}': {err}",
-                tpl_path.to_str().unwrap(),
-            );
-            let file_info = import_from.map(|(node_file, file_source, node_source)| {
-                FileInfo::new(node_file, Some(file_source), Some(node_source))
-            });
-            Err(CompileError::new(msg, file_info))
-        }
-    }
-}
-
 static CONFIG_FILE_NAME: &str = "rinja.toml";
 static DEFAULT_SYNTAX_NAME: &str = "default";
 static DEFAULT_ESCAPERS: &[(&[&str], &str)] = &[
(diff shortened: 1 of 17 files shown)


Kijewski:
It might well be that we have to convert our `Rc`s to `Arc`s to get any performance gains, if rust does not call the derive macro in the same thread. `Arc`s are ridiculously slow compared to `Rc`s, but we could use [triomphe](https://crates.io/crates/triomphe) to mitigate the slow downs a bit.

GuillaumeGomez:
Better bench in any case to ensure any change you make going forward is positive (but you already know that, just love to state the obvious :laughing: ).

Kijewski:
Well, there's a tiny performance regression:

```text
$ cd rinja_derive_standalone; cargo bench
hello_world             time:   [706.57 µs 774.11 µs 836.25 µs]
                        change: [+891.79% +965.79% +1053.5%] (p = 0.00 < 0.05)
                        Performance has regressed.

item_info.html          time:   [805.59 µs 852.78 µs 903.00 µs]
                        change: [+982.26% +1020.3% +1066.4%] (p = 0.00 < 0.05)
                        Performance has regressed.

ctrl+C
```

Not using any includes, etc., the cache is useless in this benchmark, of course, but I have not idea why the slow down are this terrible. :/

rinja_derive/src/config.rs
```diff
@@ -293,27 +297,36 @@ fn str_set<'a>(vals: &[&'a str]) -> Vec<Cow<'a, str>> {
...
-            let file_info = import_from.map(|(node_file, file_source, node_source)| {
-                FileInfo::new(node_file, Some(file_source), Some(node_source))
-            });
-            Err(CompileError::new(msg, file_info))
+        };
+        if source.ends_with('\n') {
+            let _ = source.pop();
         }
-    }
+        Ok(entry.insert(source.into()).clone())
```
GuillaumeGomez, line 328: It'd be better to replace `clone` calls (here and above) with `Rc::clone`. Not useful for performance, but better for reading. :)

rinja_derive/src/input.rs
```diff
@@ -218,6 +219,58 @@ impl TemplateInput<'_> {
...
+        syntax: Syntax<'a>,
+    }
+
+    impl PartialEq for TransientKey<'_> {
+        fn eq(&self, other: &Self) -> bool {
+            Rc::ptr_eq(&self.source, &other.source) && self.syntax == other.syntax
+        }
+    }
+
+    thread_local! {
```
GuillaumeGomez, line 248: Just thought that it's not really useful to keep these statics as thread local.

GuillaumeGomez:
> Not using any includes, etc., the cache is useless in this benchmark, of course, but I have no idea why the slow down are this terrible. :/

Well, it's still less than 1ms. So not really that important I think. Still be interesting to try to understand why these two statics are so bad at this overall.

Also would be interesting to have test cases using a lot of `include`.

Kijewski:
> Also would be interesting to have test cases using a lot of include.

Yeah, but I have no idea how to easily implement this. :/ I guess we could use [vfs](https://crates.io/crates/vfs) to have a fake file-system, but implementing that would be kinda difficult I guess, and I'd defer that task to somewhen in the future.

Kijewski:
Rinja_derive_standalone's benchmarks are useless now: We only measure how much faster the cache is than parsing. The parsing itself is not benchmarked anymore because it's part of the warm-up phase.

<details><summary>Benchmark results</summary>

```text
hello_world             time:   [46.368 µs 46.431 µs 46.492 µs]
                        change: [-9.6500% -9.0053% -8.3628%] (p = 0.00 < 0.05)
                        Performance has improved.

item_info.html          time:   [56.031 µs 56.168 µs 56.293 µs]
                        change: [-18.971% -18.687% -18.425%] (p = 0.00 < 0.05)
                        Performance has improved.

item_union.html         time:   [141.11 µs 141.43 µs 141.80 µs]
                        change: [-26.505% -26.226% -25.940%] (p = 0.00 < 0.05)
                        Performance has improved.

page.html               time:   [607.29 µs 608.38 µs 609.53 µs]
                        change: [-24.460% -24.242% -23.995%] (p = 0.00 < 0.05)
                        Performance has improved.

print_item.html         time:   [130.40 µs 130.81 µs 131.36 µs]
                        change: [-20.143% -19.854% -19.521%] (p = 0.00 < 0.05)
                        Performance has improved.

short_item_info.html    time:   [111.89 µs 112.10 µs 112.34 µs]
                        change: [-21.130% -20.563% -19.876%] (p = 0.00 < 0.05)
                        Performance has improved.

sidebar.html            time:   [169.74 µs 170.02 µs 170.32 µs]
                        change: [-24.447% -24.216% -23.991%] (p = 0.00 < 0.05)
                        Performance has improved.
```
</details>

What you would expect: "hello world" with only one expression is only mildly faster, and the more complex a template, the bigger the speed-up.

GuillaumeGomez (approved):

GuillaumeGomez:
Changes are much smaller, nicely done. :)

However, the MSRV seems to be 1.71. So just need to update it and then it's all good!

Kijewski:
Caching reading the configuration is still missing, then it's ready. :)

GuillaumeGomez:
I have to admit I didn't even think about the `config` file. Good catch! Please ping me once it's ready for ready.

Kijewski:
```
cd rinja_derive_standalone

RUST_BACKTRACE=1 \
MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-env-forward=CARGO_MANIFEST_DIR=" \
CARGO_MANIFEST_DIR="$(pwd)" \
cargo +nightly miri test
```

is unhappy.

I hate that `std::borrow::Borrow` has to return a reference. A reference to what? I cannot let `struct ConfigKey(String, String)` implement `Borrow<(&str, &str)>`, because that would have to return `&(&str, &str)`.

Maybe I can make `struct ConfigKey(Cow<'static, str>, Cow<'static, str>)` work, but things could be much easier.

Kijewski:
It's ready to review! :) The last commit is a overhaul of the previous three commits, but it would be terror to refactor them. I hope that's okay! I think I made the diff as small as possible.

GuillaumeGomez (approved):

GuillaumeGomez:
Nah it's fine, I don't care. :laughing: 

Thanks a lot!

PR #60, opened 2024-07-09
Move `#[derive(Template)]` comment to the macro itself

Kijewski:
Less searching for an explanation.

I also restructured the text to use sections instead of bullet points.

And lastly I made `rinja_derive_standalone` export `derive_template`, too, instead of `derive_template2`. This way we have less macro-magic.

### DIFF
--- a/rinja/src/lib.rs
+++ b/rinja/src/lib.rs
@@ -12,9 +12,34 @@
 //!
 //! # Creating Rinja templates
 //!
-//! An Rinja template is a `struct` definition which provides the template
-//! context combined with a UTF-8 encoded text file (or inline source, see
-//! below). Rinja can be used to generate any kind of text-based format.
+//! The main feature of Rinja is the [`Template`] derive macro
+//! which reads your template code, so your `struct` can implement
+//! the [`Template`] trait and [`Display`][std::fmt::Display], type-safe and fast:
+//!
+//! ```rust
+//! # use rinja::Template;
+//! #[derive(Template)]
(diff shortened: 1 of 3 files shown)


rinja/src/lib.rs
```diff
@@ -23,41 +23,6 @@
 //! The template syntax is very similar to [Jinja](http://jinja.pocoo.org/),
 //! as well as Jinja-derivatives like [Twig](http://twig.sensiolabs.org/) or
 //! [Tera](https://github.com/Keats/tera).
```
GuillaumeGomez, line 25: Would be nice to mention that the main element is the `template` attribute and to link to it.

rinja_derive/src/lib.rs
```diff
@@ -18,41 +18,86 @@ use generator::{Generator, MapChain};
...
-        match build_template(&ast) {
-            Ok(source) => source.parse().unwrap(),
-            Err(e) => {
-                let mut e = e.into_compile_error();
-                if let Ok(source) = build_skeleton(&ast) {
-                    let source: TokenStream = source.parse().unwrap();
-                    e.extend(source);
-                }
-                e
+/// The `template()` attribute
```
GuillaumeGomez, line 27: Missing a dot at the end of the sentence.

GuillaumeGomez:
Simplify things greatly, well done!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #63, opened 2024-07-11
Implement operator `as` in template language

### LINKED ISSUE #61: Add support for `as` operator
To allow this code:

```jinja2
{% let a = 51 as i64 %}
{% if 12.0 == y as f32 %}
```

Not if it's really useful and worth it though... It would likely add a lot of complexity for a very small gain. I'd be in favor to not do it.

Kijewski:
Because it is not always super obvious if an identifier refers to `T` or `&T`, any references are automatically followed.

Resolves #61.

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -296,7 +296,7 @@ struct MyTemplate {
 
 ## Calling functions
 
-If you only provide a function name, `rinja` will assume it's a method. If
+If you only provide a function name, rinja will assume it's a method. If
 you want to call a method, you will need to use a path instead:
 
 ```jinja
@@ -658,6 +658,16 @@ E.g. to test if the least significant bit is set in an integer field:
 {% endif %}
 ```
 
+### Type conversion
+
+You can use the [`as`](https://doc.rust-lang.org/std/keyword.as.html) operator in `{{ … }}`
+expressions, and `{% … %}` blocks. It works the same as in Rust, but with some deliberate
+restrictions:
+
+- You can only use [primitive types](https://doc.rust-lang.org/std/primitive/index.html)
+  like `i32` or `f64` both as source variable type and as target type.
+- If the source is a reference to a primitive type, e.g. `&&&bool`, then rinja automatically
+  dereferences the value until it gets the underlying `bool`.
 
 ## Templates in templates
 
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -388,7 +388,8 @@ impl<'a> Generator<'a> {
             | Expr::Try(_)
             | Expr::Tuple(_)
             | Expr::NamedArgument(_, _)
-            | Expr::FilterSource => {
+            | Expr::FilterSource
+            | Expr::As(_, _) => {
                 *only_contains_is_defined = false;
                 EvaluatedResult::Unknown
             }
@@ -523,7 +524,7 @@ impl<'a> Generator<'a> {
                     // finally dereferences it to `bool`.
                     buf.write("*(&(");
                     buf.write(self.visit_expr_root(ctx, expr)?);
-                    buf.write(") as &bool) {");
+                    buf.write(") as &::core::primitive::bool) {");
                 }
             } else if pos != 0 {
                 buf.write("} else {");
@@ -1410,6 +1411,7 @@ impl<'a> Generator<'a> {
             Expr::FilterSource => self.visit_filter_source(buf),
             Expr::IsDefined(var_name) => self.visit_is_defined(buf, true, var_name)?,
             Expr::IsNotDefined(var_name) => self.visit_is_defined(buf, false, var_name)?,
+            Expr::As(ref expr, target) => self.visit_as(ctx, buf, expr, target)?,
         })
     }
 
@@ -1423,6 +1425,19 @@ impl<'a> Generator<'a> {
             (true, true) | (false, false) => buf.write("true"),
             _ => buf.write("false"),
         }
+        Ok(DisplayWrap::Unwrapped)
+    }
+
+    fn visit_as(
+        &mut self,
+        ctx: &Context<'_>,
+        buf: &mut Buffer,
+        expr: &WithSpan<'_, Expr<'_>>,
+        target: &str,
+    ) -> Result<DisplayWrap, CompileError> {
+        buf.write(format_args!("{CRATE}::helpers::get_primitive_value(&("));
+        self.visit_expr(ctx, buf, expr)?;
+        buf.write(format_args!(")) as ::core::primitive::{target}"));
         Ok(DisplayWrap::Unwrapped)
     }
 
@@ -2494,10 +2509,11 @@ pub(crate) fn is_cacheable(expr: &WithSpan<'_, Expr<'_>>) -> bool {
         Expr::Group(arg) => is_cacheable(arg),
         Expr::Tuple(args) => args.iter().all(is_cacheable),
         Expr::NamedArgument(_, expr) => is_cacheable(expr),
+        Expr::As(expr, _) => is_cacheable(expr),
+        Expr::Try(expr) => is_cacheable(expr),
         // We have too little information to tell if the expression is pure:
         Expr::Call(_, _) => false,
         Expr::RustMacro(_, _) => false,
-        Expr::Try(_) => false,
         // Should never be encountered:
         Expr::FilterSource => unreachable!("FilterSource in expression?"),
     }
--- a/rinja_parser/src/expr.rs
+++ b/rinja_parser/src/expr.rs
@@ -45,6 +45,7 @@ pub enum Expr<'a> {
     Attr(Box<WithSpan<'a, Expr<'a>>>, &'a str),
     Index(Box<WithSpan<'a, Expr<'a>>>, Box<WithSpan<'a, Expr<'a>>>),
     Filter(Filter<'a>),
+    As(Box<WithSpan<'a, Expr<'a>>>, &'a str),
     NamedArgument(&'a str, Box<WithSpan<'a, Expr<'a>>>),
     Unary(&'a str, Box<WithSpan<'a, Expr<'a>>>),
     BinOp(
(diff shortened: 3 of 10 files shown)


rinja_derive/src/generator.rs
```diff
@@ -2181,10 +2195,11 @@ pub(crate) fn is_cacheable(expr: &WithSpan<'_, Expr<'_>>) -> bool {
         Expr::Group(arg) => is_cacheable(arg),
         Expr::Tuple(args) => args.iter().all(is_cacheable),
         Expr::NamedArgument(_, expr) => is_cacheable(expr),
+        Expr::As(expr, _) => is_cacheable(expr),
+        Expr::Try(expr) => is_cacheable(expr),
         // We have too little information to tell if the expression is pure:
         Expr::Call(_, _) => false,
         Expr::RustMacro(_, _) => false,
-        Expr::Try(_) => false,
```
Kijewski, line 2187: If `e` is cacheable, then `e?` should be cacheable, too. What do you think?

rinja_parser/src/expr.rs
```diff
@@ -188,7 +189,28 @@ impl<'a> Expr<'a> {
...
+        let (i, kw) = ws(opt(keyword("as")))(i)?;
+        if kw.is_none() {
+            return Ok((i, lhs));
+        }
+
+        let (j, target) = opt(identifier)(i)?;
+        let target = target.unwrap_or_default();
+        match target {
+            "bool" | "f32" | "f64" | "i8" | "i16" | "i32" | "i64" | "i128" | "isize" | "u8"
+            | "u16" | "u32" | "u64" | "u128" | "usize" => {
```
Kijewski, line 205: IMHO any other conversion should be done in rust code. You should never need to write e.g. `{{ x as Box<dyn Error> }}` in a templating language.

GuillaumeGomez: Yes, let's keep it simple. :+1:

rinja_derive/src/generator.rs
GuillaumeGomez: Hum... I don't see any reason not too. If tests are passing...

GuillaumeGomez:
Question: for now it only seems to be supported in "display expressions". Should it allowed in other cases (like `if` conditions) or not?

Kijewski:
It's allowed in all kind of expressions. Right now, the same expression parser is used in all places. You already could write e.g. `{% if ("<"|e) == "&#60; %}`, and with this PR `{% if 0xff00 as u8 == 0 %}` works, too. I would not advise to using filters in tests, but type conversions, why not?

GuillaumeGomez:
Sounds absolutely fine! Then please mention it in the book. :)

book/src/template_syntax.md
```diff
@@ -658,6 +658,16 @@ E.g. to test if the least significant bit is set in an integer field:
 {% endif %}
 ```
 
+### Type conversion
+
+You can use the operator [`as`](https://doc.rust-lang.org/std/keyword.as.html) in `{{ … }}`
+expressions, and `{% … %}` blocks. It works the same as in Rust, but with some deliberate
+restrictions:
+
+- You can only use [primive types](https://doc.rust-lang.org/std/primitive/index.html)
```
GuillaumeGomez, line 667:
```suggestion
- You can only use [primitive types](https://doc.rust-lang.org/std/primitive/index.html)
```

book/src/template_syntax.md
```diff
@@ -658,6 +658,16 @@ E.g. to test if the least significant bit is set in an integer field:
...
+  like `i32` or `f64` both as source variable type and as target type.
+- If the source is a reference to a primitive type, e.g. `&&&bool`, then rinja automatically
+  dereferences the value until it gets get the unterlying `bool`.
```
GuillaumeGomez, line 670:
```suggestion
- If the source is a reference to a primitive type, e.g. `&&&bool`, then rinja automatically
  dereferences the value until it gets the underlying `bool`.
```

rinja_parser/src/expr.rs
```diff
@@ -190,26 +191,42 @@ impl<'a> Expr<'a> {
...
-            None => return Ok((i, lhs)),
-            Some(None) => {
+        let (j, rhs) = opt(ws(identifier))(i)?;
+        let i = match rhs {
+            Some("is") => j,
+            Some("as") => {
+                let (i, target) = opt(identifier)(j)?;
+                let target = target.unwrap_or_default();
+                return match target {
+                    "bool" | "f32" | "f64" | "i8" | "i16" | "i32" | "i64" | "i128" | "isize"
```
GuillaumeGomez, line 206: Missing f16 and f128. Would be nice to have the full list of primitives somewhere centralized because we use it in different places.

rinja_parser/src/expr.rs
```diff
@@ -190,26 +191,42 @@ impl<'a> Expr<'a> {
...
+            None => {
                 return Err(nom::Err::Failure(ErrorContext::new(
                     "expected `defined` or `not defined` after `is`",
                     // We use `start` to show the whole `var is` thing instead of the current token.
                     start,
                 )));
             }
-            Some(Some(None)) => false,
-            Some(Some(Some(_))) => true,
+            Some(None) => Self::IsDefined,
```
GuillaumeGomez, line 228: Nice improvement! I wonder how I could not think about it directly...

book/src/template_syntax.md
```diff
@@ -658,6 +658,16 @@ E.g. to test if the least significant bit is set in an integer field:
...
+### Type conversion
+
+You can use the operator [`as`](https://doc.rust-lang.org/std/keyword.as.html) in `{{ … }}`
+expressions, and `{% … %}` blocks. It works the same as in Rust, but with some deliberate
+restrictions:
+
+- You can only use [primitive types](https://doc.rust-lang.org/std/primitive/index.html)
+  like `i32` or `f64` both as source variable type and as target type.
+- If the source is a reference to a primitive type, e.g. `&&&bool`, then rinja automatically
+  dereferences the value until it gets get the underlying `bool`.
```
GuillaumeGomez, line 670:
```suggestion
- If the source is a reference to a primitive type, e.g. `&&&bool`, then `rinja` automatically
  dereferences the value until it gets the underlying `bool`.
```

Some parts of the suggestion were skipped. ;)

Kijewski: Oops, fixed now. Thank you for reading my texts, twice. :D

book/src/template_syntax.md
```diff
@@ -658,6 +658,16 @@ E.g. to test if the least significant bit is set in an integer field:
...
 
+### Type conversion
+
+You can use the [`as`](https://doc.rust-lang.org/std/keyword.as.html) operator in `{{ … }}`
+expressions, and `{% … %}` blocks. It works the same as in Rust, but with some deliberate
+restrictions:
+
+- You can only use [primitive types](https://doc.rust-lang.org/std/primitive/index.html)
+  like `i32` or `f64` both as source variable type and as target type.
+- If the source is a reference to a primitive type, e.g. `&&&bool`, then rinja automatically
```
GuillaumeGomez, line 669: Still forgot backticks around `rinja`. :p

We do it in multiple places so for the power of coherency, we need to continue! 🤣

rinja_parser/src/expr.rs
Kijewski: I could have thought about that before approving your PR, too. Sometimes one notices the obvious only after spending some time on other stuff. :)

book/src/template_syntax.md
Kijewski: But "rinja" is a name that we don't use as a code snippet in here. I would rather remove the backticks around the other instances in book/src/template_syntax.md:299 and README.md:3. Maybe replace it with `**Rinja**` in the book (start of a sentence; bold like the first use of a lemma in Wikipedia)?

GuillaumeGomez: We need to pick one, I'm fine with whichever.

rinja_parser/src/expr.rs
Kijewski: Good idea! Rust is even smart enough to overlay lists `&["a"]`, `&["a", "b"]` and `&["a", "b", "c"]`, so they all point to the same memory address, but with different lengths.

rinja_parser/src/expr.rs
```diff
@@ -190,26 +191,40 @@ impl<'a> Expr<'a> {
...
+        let i = match rhs {
+            Some("is") => j,
+            Some("as") => {
+                let (i, target) = opt(identifier)(j)?;
+                let target = target.unwrap_or_default();
+                if crate::PRIMITIVE_TYPES.contains(&target) {
+                    return Ok((i, WithSpan::new(Self::As(Box::new(lhs), target), start)));
+                } else {
+                    return Err(nom::Err::Failure(ErrorContext::new(
+                        "operator `as` expects the name of a primitive type on its right-hand side",
```
GuillaumeGomez, line 209: Sorry, completely missed this message error typo:

```suggestion
                        "`as` operator expects the name of a primitive type on its right-hand side",
```

I would also add the invalid rhs used like:

```suggestion
                        "`as` operator expects the name of a primitive type on its right-hand side, found `{target}`",
```

rinja_parser/src/expr.rs
```diff
@@ -190,26 +191,48 @@ impl<'a> Expr<'a> {
...
+            Some("is") => j,
+            Some("as") => {
+                let (i, target) = opt(identifier)(j)?;
+                let target = target.unwrap_or_default();
+                if crate::PRIMITIVE_TYPES.contains(&target) {
+                    return Ok((i, WithSpan::new(Self::As(Box::new(lhs), target), start)));
+                } else if target.is_empty() {
+                    return Err(nom::Err::Failure(ErrorContext::new(
+                        "`as` operator expects the name of a primitive type on its right-hand side",
+                        j,
```
GuillaumeGomez, line 210: Sorry, I think it'll be the last one (I hope XD). Please use `start` instead of `j`. I think the error would look better in this case if the `as` keyword is displayed as well.

Same below.

Kijewski: Hehe, no need to be sorry, thank you for your reviews! :)

GuillaumeGomez (approved):

GuillaumeGomez:
Looks great, thanks a lot!

PR #64, opened 2024-07-11
Add `ifdef` block: enable section depending on the struct's fields

### LINKED ISSUE #62: Add support for `is defined`
This implementation would be **at best** strongly limited to one level: the current struct fields and the declared variables in the template.

It would also need https://github.com/rinja-rs/rinja/issues/47 to be implemented, so that in case the variable/field is not declared, the inner block is not generated, preventing compilation errors.

Kijewski:
Todos:

- [ ] fix whitespace trimming errors
- [ ] add UI tests
- [ ] add documentation in the book

Resolves #62.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -7,7 +7,8 @@ use std::rc::Rc;
 use std::{cmp, hash, mem, str};
 
 use parser::node::{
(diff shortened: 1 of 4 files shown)


Kijewski:
@GuillaumeGomez, do you have an idea how `Generator::handle_ws`, `Generator::prepare_ws`, etc. work and what I am doing wrong?

Kijewski:
Closed in favor of #78.

PR #66, opened 2024-07-11
phf should be a dev dependency

softdevca:
The `phf` crate is used in a benchmark for testing identifier normalization but not in the library itself.  This moves it to be a dev dependency.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #68, opened 2024-07-12
Run `typos` to fix typos

Kijewski:


Kijewski:
There's a ton of false positives, like converting "lits" (literals) to "list". The tools actually wants to fix the mock Latin in lorem ipsums. I don't know how much of a hassle it would be to set it up right. In most cases "lits" is a typo, just in in one case it isn't.

But maybe it's easier than I think? I guess other will have the exact same problem with other abbreviate words, and lorem ipsums are used a lot. If you want to, feel free to look into how to set it up for the CI, and I'd be happy to "borrow" the feature for other projects. ;)

GuillaumeGomez:
Hum, I'll open an issue for that to be done later on. Found a bug in docs.rs migration I'm trying to solve. ;)

Thanks for this PR!

PR #69, opened 2024-07-12
Move integration crates in their own workspaces

Kijewski:
Everytime you hit ctrl+s in VS code, the whole project in checked, which normally is a nice thing, and I would not like to disable that in my configuration.

But in rinja, the diffent integration crates don't work well with one another and every single one will have compile errors in the automatic check, because `rinja_derive` won't know what the relevant `rinja_xyz` crate is. This renders the automatic check useless. Then it takes ages to compile and recompile all transitively used `proc_macros`.

This PR moves all integration crates in there own, individual workspaces. `cargo check` will ignore them unless you traverse manually into their folders. The CI is set up to do just that.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: ['main']
+    branches: ['master']
   pull_request:
   schedule:
     - cron: "32 4 * * 5"
@@ -24,9 +24,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --all-targets
-      - run: cargo test
+      - run: cargo nextest run
 
   Package:
     strategy:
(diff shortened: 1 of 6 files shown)


.github/workflows/rust.yml
```diff
@@ -38,20 +41,35 @@ jobs:
...
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@nightly
         with:
-          components: rustfmt, clippy
-      - uses: Swatinem/rust-cache@v2
-      - run: cargo fmt --all -- --check
-      - run: cargo clippy --all-targets -- -D warnings
+          components: rustfmt
+      - run: cd rinja                  && cargo fmt -- --check
```
GuillaumeGomez, line 64: Instead of doing it like this, why not using matrix so then we list all folders we want to go into so we can just have `run: cd $FOLDER && cargo fmt -- --check`?

Kijewski: > No need to run the checks in parallel. They are fast enough.

Even though the runners are free for open source projects, I don't think we should use more CPU time than we need to. Be eco friendly and such.

Also, we already have 17 checkmarks on each PR. At some point it gets annoying to scroll until you find the red cross on the failed test. :)

GuillaumeGomez: Can't have that not in parallel?

Kijewski: I don't know if there is a "run is one `run:` instruction multiple times with different env vars" syntax?

GuillaumeGomez: `run` is shell command, so we can likely have an array and iterate over it? (But I suppose it'd need a `cd -` command in addition, I suppose it's fine)

Kijewski: Yeah, should have thought about that! :) A formatting error will be reported like that: <https://github.com/rinja-rs/rinja/actions/runs/9905823875/job/27366138342?pr=69>

Kijewski:
> If you don't want the output, why not using `cd` directly?

For one simple reason: I did not think about that obvious solution. :D In my brain it's hardwired "when iterating over subfolders in bash, then use `pushd && … && popd`". Will change.

.github/workflows/rust.yml
GuillaumeGomez: Sounds perfectly fine. :+1:

.github/workflows/rust.yml
```diff
@@ -38,20 +41,37 @@ jobs:
...
+          components: rustfmt
+      - run: |
+          set -eu
+          for PKG in \
+            rinja rinja_actix rinja_axum rinja_derive rinja_derive_standalone \
+            rinja_parser rinja_rocket rinja_warp testing examples/actix-web-app
+          do
+            cd "$PKG"
+            echo "Testing: $PKG"
+            cargo fmt -- --check
```
GuillaumeGomez, line 72: Just one question: there is no more `clippy` check here. However I see there is one above. Is `--all-targets` run on all crates even if they are in different workspaces?

Kijewski: Does `--all-targets` even do anything if you don't specify a target list somewhere? Isn't all=native in by default?

Clippy checking on +nightly is always stressful to me, because often tests are a bit over-eager in their early stage.

The clippy further up ensures that all packages are tested on +stable. If you run clippy in a workspace root, I think normally only default-members are tested, but I'm not sure.

GuillaumeGomez: Hum... I don't mind running non-nightly clippy. Just want to be sure it'll check all crates, even those which are not in the default workspace.

Kijewski: The list is exhaustive. Everything is checked. We just have to make sure, that the list stays exhaustive. I don't see use adding new crates every other day, so this requirement should be fine. :)

GuillaumeGomez: Sorry, that's not what I meant. Let me try again. 😆 

So now, the default workspace doesn't include all crates. So here, I was wondering if `cargo clippy --all-targets` was checking all crates, even if not in the default workspace. If not, could we make the same loop (and store the list of crate in a variable?) so we can run `clippy` (stable version) on all crates?

Kijewski: I confused myself what "targets" are in this case. In other cases "target" is short for "target platform", but in here "target" is the same as in Make targets: 

> Build all targets. This is equivalent to specifying --lib --bins --tests --benches --examples.

No, clippy won't traverse into folders to find additional crates. If clippy is executed in a workspace root, the workspace is checked (I don't know if all members or only the default-members). If the current folder is not a workspace root, then the current package is tested.

In the **current** configuration **not all** packages are linted. In the **new** configuration **all** packages are linted. The new config is strictly better than the old one. :)

GuillaumeGomez: So confusing. XD Thanks for the clarification! Then all good for me. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #71, opened 2024-07-12
Fix filter block

GuillaumeGomez:
It's the last issue I encountered blocking the docs.rs migration.

So in short: we were writing all "sub instructions" (like `include`) in `writer` as usual. Except that the filter block is supposed to be applied to anything that was output inside it. To go around this problem, I simpler shadow the `writer` variable in a sub-block (the filter block) and then call the filters on it.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -39,6 +39,8 @@ pub(crate) struct Generator<'a> {
     buf_writable: WritableBuffer<'a>,
     // Counter for write! hash named arguments
     named: usize,
+    // Used in blocks to check if we are inside a filter block.
+    is_in_filter_block: usize,
 }
 
 impl<'a> Generator<'a> {
@@ -48,6 +50,7 @@ impl<'a> Generator<'a> {
         heritage: Option<&'n Heritage<'_>>,
         locals: MapChain<'n, Cow<'n, str>, LocalMeta>,
         buf_writable_discard: bool,
+        is_in_filter_block: usize,
     ) -> Generator<'n> {
         Generator {
             input,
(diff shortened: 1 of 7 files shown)


rinja_derive/src/generator.rs
```diff
@@ -39,6 +39,8 @@ pub(crate) struct Generator<'a> {
     buf_writable: WritableBuffer<'a>,
     // Counter for write! hash named arguments
     named: usize,
+    // Used in blocks to check if we are inside a filter block.
+    is_in_filter_block: bool,
```
Kijewski, line 43: You can use `is_in_filter_block: u32`, and use `+= 1` and `-= 1` so you don't need the `prev` variable.

rinja_derive/src/generator.rs
```diff
@@ -92,6 +96,7 @@ impl<'a> Generator<'a> {
             -> {CRATE}::Result<()> {{",
         ));
         buf.writeln(format_args!("use {CRATE}::filters::AutoEscape as _;"));
+        buf.writeln(format_args!("use std::fmt::Write;"));
```
Kijewski, line 99: Please use `use ::core::fmt::Write as _;`. We already use `::core` for everything that is in `core`. The `as _` is useful not to shadow any other `Writer` name the user want's to use, e.g. `std::io::Writer`.

Kijewski (changes requested):
First small nit-picking review. :)

rinja_derive/src/generator.rs
GuillaumeGomez: :+1:

GuillaumeGomez:
Fixed the nits.

Kijewski (approved):

PR #73, opened 2024-07-13
Reimplement `{% filter %}` block and make `paragraphbreaks` safe by default

Kijewski:
This PR reimplements the code generation for `{% filter %}` blocks, so that the data is written directly into its destination writer, without using a buffer. This way it behaves like a specialized `{{ expr|filter1|filter2 }}` would, if the `expr` was a `Template` that contained the body of the filter block.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -34,6 +34,7 @@ Enable it with Cargo features (see below for more information).
   * [`linebreaks`][#linebreaks]
   * [`linebreaksbr`][#linebreaksbr]
   * [`lower|lowercase`][#lower]
+  * [`paragraphbreaks`][#paragraphbreaks]
   * [`ref`][#ref]
   * [`safe`][#safe]
   * [`title`][#title]
--- a/rinja/src/filters/escape.rs
+++ b/rinja/src/filters/escape.rs
@@ -12,9 +12,12 @@ use std::str;
 /// so this filter only takes a single argument of any type that implements
 /// `Display`.
 #[inline]
-pub fn safe(text: impl fmt::Display, escaper: impl Escaper) -> Result<impl Display, Infallible> {
+pub fn safe(
+    text: impl fmt::Display,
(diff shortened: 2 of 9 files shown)


Kijewski:
I did not understand the code of `{% filter %}` blocks at all, so I took a stab at re-implementing them, I hope that's okay! I wouldn't have been able to review the other PR.

My implementation seems to be sane, but your tests, @GuillaumeGomez, expect different whitespaces than my code emits. To me it looks like my implementation works correctly. Could you please have a look in the tests or the implementation is wrong?

GuillaumeGomez:
The output seems correct indeed. The code looks code as well. Very good idea to use a closure! Some tests of mine are missing.

If you don't mind, I'll merge #71, like that you will have tests already and will just need to update them.

Kijewski:
> If you don't mind, I'll merge https://github.com/rinja-rs/rinja/pull/71, like that you will have tests already and will just need to update them.

Sure, that's the simplest solution!

GuillaumeGomez:
Perfect thanks! Just waiting for your review before merging it then. ;)

GuillaumeGomez:
Ok, a rebase and fixed tests and then please ping me for review. :)

GuillaumeGomez (approved):

GuillaumeGomez:
This is so much better, thanks a lot! I also updated the title of the PR about the `paragraphbreaks` filter.

Kijewski:
> This is so much better, thanks a lot!

Thank you! Yeah, I assumed that this feature could be implemented with a closure and after a while of testing it actually worked! :)

> I also updated the title of the PR about the paragraphbreaks filter.

Oops, I didn't do that in this PR. :/ I only changed `|e` and `|safe`.

GuillaumeGomez:
Woups. Being woken up too early by the cat is bad for my brain haha. Please fix my fix of the title. ^^'

PR #75, opened 2024-07-13
filters: Don't use `.to_string()`

Kijewski:
Rinja's `render()` is fallible, so a user might not expect that an input to e.g. `|upper` can cause a panic if its formatting failed.

This PR replaces `impl ToString` parameters with `impl fmt::Display`, and the string building uses a fallible `write!()` call.


GuillaumeGomez (approved):

GuillaumeGomez:
Very good point, thanks!

PR #77, opened 2024-07-13
 filters: proper escaping for `|linebreaks` and friends 

### LINKED ISSUE #74: `|linebreaks`, `|linebreaksbr` and `|paragraphbreaks` don't work as intended with regards to escaping
The examples in the book to e.g. `|paragraphbreaks` are wrong. The output of

```jinja2
{{ "hello\n\nworld"|paragraphbreaks }}
```

isn't 

```html
<p>hello</p><p>world</p>
```

but

```html
&#60;p&#62;hello&#60;/p&#62;&#60;p&#62;world&#60;/p&#62;
```

because the filter is not marked as HTML safe, i.e. it returns `Result<impl fmt::Display, Infallible>` instead of `Result<HtmlSafeOutput<impl fmt::Display>, Infallible>`. That makes it tempting to use `{{ "hello\n\nworld"|paragraphbreaks|safe }}`, but the output of `|paragraphbreaks` is not safe, because it does not escape its input.

Should we auto-escape the input of `|paragraphbreaks` and mark its output as safe? Right now you can do that manually with `{{ "hello\n\nworld"|escape|paragraphbreaks|safe }}`, which looks ridiculous.

Kijewski:
The filters `|linebreaks`, `|linebreaksbr` and `|paragraphbreaks` generate HTML code to be embedded in a page. Having to specify that the output of these filters is `|safe` is cumbersome. Also, these filters need to operate on already escaped HTML data. This could be done by writing `{{ s|escape|linebreaks|safe }}`.

This PR does the input and output formatting escaping for the user. The input gets escaped for HTML (invariant of the selected escaper), and the output gets marked as HTML safe.

Resolves #74.


testing/tests/filter_block.rs
```diff
@@ -254,8 +254,7 @@ fn filter_block_include() {
     {% endif -%}
 {% endfilter -%}
 "#,
-    ext = "html",
-    print = "code"
```
GuillaumeGomez, line 258: Woups. :-.

rinja_derive/src/lib.rs
```diff
@@ -311,9 +311,6 @@ const BUILT_IN_FILTERS: &[&str] = &[
     "into_f64",
     "into_isize",
     "join",
-    "linebreaks",
```
GuillaumeGomez, line 314: Isn't it still a function in `generator.rs`?

Kijewski: Oops, yes, of course.

Kijewski: I sorted the list while re-adding the items. Looks nicer this way. :)

rinja_derive/src/lib.rs
```diff
@@ -313,17 +313,17 @@ const BUILT_IN_FILTERS: &[&str] = &[
...
+    "paragraphbreaks",
     "safe",
     "title",
     "trim",
     "truncate",
     "upper",
     "uppercase",
-    "urlencode",
     "urlencode_strict",
+    "urlencode",
```
GuillaumeGomez, line 326: You failed the sorting for this one. ;)

(0 is smaller than `_` hehe)

Kijewski: I would say the same all the time! But somehow the sort function in VS code, geany, etc. always give precedence to the longer word.

GuillaumeGomez: Wut. Oh well.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

Kijewski:
And thank your for preventing me from introducing an bug in the built-in filters list! :D

GuillaumeGomez:
Sometimes my reviews are useful. :)

PR #78, opened 2024-07-15
Implement `is (not) defined` feature

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/62.

I'm not super happy about how I implemented the parser part (the little "cheat") but found it easier to be done this way to keep the same logic as the other operators. But it forced to have some more error checking.

The other downside is that it makes the code to generate the `if`/`else` branches a bit messy. If you have ideas on how to make it look better by any chance?

### DIFF
--- a/rinja_parser/src/expr.rs
+++ b/rinja_parser/src/expr.rs
@@ -64,6 +64,8 @@ pub enum Expr<'a> {
     Try(Box<WithSpan<'a, Expr<'a>>>),
     /// This variant should never be used directly. It is created when generating filter blocks.
     FilterSource,
+    IsDefined(&'a str),
+    IsNotDefined(&'a str),
 }
 
 impl<'a> Expr<'a> {
@@ -188,7 +190,56 @@ impl<'a> Expr<'a> {
     expr_prec_layer!(band, shifts, token_bitand);
     expr_prec_layer!(shifts, addsub, alt((tag(">>"), tag("<<"))));
     expr_prec_layer!(addsub, muldivmod, alt((tag("+"), tag("-"))));
-    expr_prec_layer!(muldivmod, filtered, alt((tag("*"), tag("/"), tag("%"))));
+    expr_prec_layer!(muldivmod, is_defined, alt((tag("*"), tag("/"), tag("%"))));
+
+    fn is_defined(i: &'a str, level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
+        let (_, level) = level.nest(i)?;
+        let start = i;
+        let (i, lhs) = Self::filtered(i, level)?;
+        let (i, rhs) = opt(preceded(
+            ws(keyword("is")),
+            opt(terminated(opt(keyword("not")), ws(keyword("defined")))),
+        ))(i)?;
+        let is_neg = match rhs {
+            None => return Ok((i, lhs)),
+            Some(None) => {
+                return Err(nom::Err::Failure(ErrorContext::new(
+                    "expected `defined` or `not defined` after `is`",
+                    // We use `start` to show the whole `var is` thing instead of the current token.
+                    start,
+                )));
+            }
+            Some(Some(None)) => false,
+            Some(Some(Some(_))) => true,
+        };
+        let var_name = match *lhs {
+            Self::Var(var_name) => var_name,
+            Self::Attr(_, _) => {
+                return Err(nom::Err::Failure(ErrorContext::new(
+                    "`is defined` operator can only be used on variables, not on their fields",
+                    start,
+                )));
+            }
+            _ => {
+                return Err(nom::Err::Failure(ErrorContext::new(
+                    "`is defined` operator can only be used on variables",
+                    start,
+                )));
+            }
+        };
+
+        Ok((
+            i,
+            WithSpan::new(
+                if is_neg {
+                    Self::IsNotDefined(var_name)
+                } else {
+                    Self::IsDefined(var_name)
+                },
+                start,
+            ),
+        ))
+    }
 
     fn filtered(i: &'a str, mut level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
         let start = i;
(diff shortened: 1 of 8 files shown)


Kijewski:
Sorry, but I won't find the time to review before Friday/Saturday.

rinja_parser/src/expr.rs
```diff
@@ -188,7 +190,59 @@ impl<'a> Expr<'a> {
...
-    expr_prec_layer!(muldivmod, filtered, alt((tag("*"), tag("/"), tag("%"))));
+    expr_prec_layer!(muldivmod, is_defined, alt((tag("*"), tag("/"), tag("%"))));
+
+    fn is_defined(i: &'a str, level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
+        let (_, level) = level.nest(i)?;
+        let start = i;
+        let (i, left) = Self::filtered(i, level)?;
+        // This is cheating: to make it easier, we expect `defined` to be provided afterward
+        // in order to keep the same code logic as the other calls.
+        let (i, right) = many0(pair(tuple((ws(tag("is ")), ws(opt(tag("not "))))), |i| {
```
Kijewski, line 201:
```rs
let (i, is_neg) = preceded(
    ws(keyword("is")),
    cut(terminated(
        opt(keyword("not")),
        ws(keyword("defined")),
    )),
)(i)?;
```

GuillaumeGomez:
> Sorry, but I won't find the time to review before Friday/Saturday.

You say that and then provide suggestion right away. 😆 

But in any case, it's not in a hurry. So please review when you have time and motivation to keep this project fun for you. :)

GuillaumeGomez:
Finally applied your suggestion for the parser. Looks much better like this, thanks!

GuillaumeGomez:
> The other downside is that it makes the code to generate the `if`/`else` branches a bit messy. If you have ideas on how to make it look better by any chance?

Thinking some more about this, I think we can simply create a new very small AST for conditions which would simply remove the conditions (and their associated block) that won't be rendered and then the function can simply loop through them instead of the whole state machine. Gonna do that.

GuillaumeGomez:
Moved the logic out of the `write_if` method. Much closer to what it originally was.

Kijewski:
> Moved the logic out of the `write_if` method. Much closer to what it originally was.

Thank you, much easier to read! I give you a proper review soon-ish.

rinja_parser/src/expr.rs
```diff
@@ -188,7 +190,56 @@ impl<'a> Expr<'a> {
     expr_prec_layer!(band, shifts, token_bitand);
     expr_prec_layer!(shifts, addsub, alt((tag(">>"), tag("<<"))));
     expr_prec_layer!(addsub, muldivmod, alt((tag("+"), tag("-"))));
-    expr_prec_layer!(muldivmod, filtered, alt((tag("*"), tag("/"), tag("%"))));
+    expr_prec_layer!(muldivmod, is_defined, alt((tag("*"), tag("/"), tag("%"))));
+
+    fn is_defined(i: &'a str, level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
+        let (_, level) = level.nest(i)?;
```
Kijewski, line 196: I think we can to without the nesting check in here. We only traverse to a lower level, not the same or a higher level.

rinja_parser/src/expr.rs
```diff
@@ -188,7 +190,56 @@ impl<'a> Expr<'a> {
...
+                    "expected `defined` or `not defined` after `is`",
+                    // We use `start` to show the whole `var is` thing instead of the current token.
+                    start,
+                )));
+            }
+            Some(Some(None)) => false,
+            Some(Some(Some(_))) => true,
+        };
+        match *lhs {
+            Self::Var(_) => {}
```
Kijewski, line 216: Please capture the variable name in here.

rinja_parser/src/expr.rs
```diff
@@ -188,7 +190,56 @@ impl<'a> Expr<'a> {
...
+                    start,
+                )));
+            }
+        }
+
+        Ok((
+            i,
+            WithSpan::new(
+                if is_neg {
+                    Self::IsNotDefined(Box::new(lhs))
```
Kijewski, line 235: The use the variable name in here. No need for boxing. This makes the code a lot more simple, I think.

Kijewski:
Sorry for the iterative reviews! It's a bit hard for me to wrap my head around this feature.

rinja_parser/src/expr.rs
GuillaumeGomez: Don't we need that for `Self::filtered(i, level)?` just below?

rinja_parser/src/expr.rs
GuillaumeGomez: That was an incredible idea! Code got simplified by quite a lot! :D

rinja_parser/src/expr.rs
GuillaumeGomez: Done!

GuillaumeGomez:
> Sorry for the iterative reviews! It's a bit hard for me to wrap my head around this feature.

It's fine. Don't hesitate if you need some clarifications or any help to understand the code (which, if not easy to understand, should be more documented! So don't hesitate to tell me what's unclear so I can add code comments :smiley:)

rinja_parser/src/expr.rs
Kijewski: No, you can just pass the `level` as is. In all other cases we also only increment the recursion level on recursions. The code cannot parse an expression like `y is defined is defined`, so no nesting happens in here.

Kijewski:
I pushed a commit to remove the line of unsafe code. I actually wanted to look if it was easily possible to remove it, and the best test was implementing it. :D

The code looks good, and after reading your comments properly, I understand it perfectly (I think). Great work!

rinja_parser/src/expr.rs
GuillaumeGomez: Good point! Removed this nesting.

GuillaumeGomez:
Fixed the merge conflict and applied your suggestion. :)

> I pushed a commit to remove the line of unsafe code. I actually wanted to look if it was easily possible to remove it, and the best test was implementing it. :D

That's my first implementation. I removed it to reduce the number of lifetimes. If you prefer with references, then I'm fine with it. :)

> The code looks good, and after reading your comments properly, I understand it perfectly (I think). Great work!

Awesome! \o/

Kijewski (dismissed):
Thank you! Please revert the changes in the ui test nesting tests, then you can merge the PR.

GuillaumeGomez:
Reverted wrong nested ui tests changes.

Kijewski (approved):
Thank you, well done!

PR #80, opened 2024-07-16
Don't link latest version of the book

Kijewski:
This way we will be able to decide whether we want to link the bleeding-edges version of the book, or (more likely) the last stable release.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #82, opened 2024-07-19
Add newly added f16 and f128 types in parsing

GuillaumeGomez:
https://github.com/rust-lang/rust/issues/116909

### DIFF
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -432,7 +432,7 @@ fn num_lit(i: &str) -> ParseResult<'_> {
             tag("usize"),
         ))(i)
     };
-    let float_suffix = |i| alt((tag("f32"), tag("f64")))(i);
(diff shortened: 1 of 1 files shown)


Kijewski (changes requested):
The change is fine. The compiler will tell the user that they can't use f128 just yet. 

But I don't like the current implementation, neither for int suffices nor float. IMO it would be better to capture an ident and match the result. This reduces the chance of false positive matches like 1_u321.

GuillaumeGomez:
Ok, merging this one and opening an issue. Gonna give it a try. :)

GuillaumeGomez:
Ah wait no, can't since you didn't approve. Gonna try in this PR then I suppose. :stuck_out_tongue:

Kijewski (approved):
Approved, because it's strictly better than before :)

GuillaumeGomez:
Opening an issue then. ;)

PR #85, opened 2024-07-21
Prevent duplication of `include_bytes!` statements

### LINKED ISSUE #72: Only use `include_byte!` on a file once across templates
Currently, everytime we encounter an `include` or an `extends`, we add an `include_byte!` call in the code. It's pretty bad and since we now keep things across the process, I think we should also keep this information to know if we actually need to generate another `include_byte!` call.

GuillaumeGomez:
Fixes #72.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -3,9 +3,10 @@ use std::collections::hash_map::{Entry, HashMap};
 use std::fmt::{Arguments, Display, Write};
 use std::ops::Deref;
 use std::path::Path;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::{cmp, hash, mem, str};
 
+use once_map::OnceMap;
 use parser::node::{
     Call, Comment, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Match, Whitespace, Ws,
 };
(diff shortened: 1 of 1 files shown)


rinja_derive/src/generator.rs
```diff
@@ -88,6 +89,23 @@ impl<'a> Generator<'a> {
         Ok(buf.buf)
     }
 
+    fn generate_include_bytes<S: AsRef<str>>(&mut self, buf: &mut Buffer, path: S) {
+        static CACHE: OnceLock<OnceMap<String, &'static ()>> = OnceLock::new();
```
Kijewski, line 93: Thinking about it, I don't think using a cache this why is sane:

* Given the file `layout.html` is used in compiling units `a.rs` and `b.rs`, then the include might wind up only in one of the generated objects, say, `a.rs`.
* Afterwards, we remove `a.rs` and any mentions to it.
* Then we alter `layout.html`: `b.rs` does not mention the dependency, so the change won't cause a recompilation.

GuillaumeGomez: If we remove `a.rs`, the crate will be recompiled and `layout.html` will be used in `b.rs`. Or did I miss something?

Kijewski: You cannot assume that the whole crate is recompiled when incremental compilation is enabled, can you?

GuillaumeGomez: Hum... It's a good question. But since you remove `a.rs`, it also means you remove `mod a` somewhere else, so in this case, I wonder if the whole crate isn't simply rebuilt...

Kijewski: But what if the add more indirection? (Assume the usual folder structure):

```rust
mod x1 {
    mod x2 {
        mod a;
    }
    mod x3 {
        mob b;
    }
}
```

Then removing `mod a` while cause a recompilation of `mod x2`, but not `x1`. If the need to rebuild would bubble up, then there could be no incremental compilation.

GuillaumeGomez: The fact that we're debating over this is already a pretty good sign that we should not do it. Closing then! :)

Kijewski: Maybe once upon a time <https://github.com/rust-lang/rust/issues/99515> gets implemented. :) If I understand the proposal correctly, then this could be a game changer, and we could even use spans inside the template file to show error messages, etc.

PR #86, opened 2024-07-22
Try not to use `fmt::Formatter`s

Kijewski:
When you write an `fmt::Display`able type into a `fmt::Write` sink, a `fmt::Formatter` is used to facilitate the writing. It stores a `&dyn fmt::Write` reference to the sink, so less duplicated byte code is generated, but any possible optimizations become void.

This PR lets rinja try to avoid using the normal `write!()` machinery. Indeed, no `write!()` call is made directly anymore in the generated code, but only though a `WriteWritable` helper trait. This aids the runtime performance quite a bit:

```text
$ cd testing/benches
$ cargo bench --bench all

Big table               time:   [405.60 µs 406.46 µs 407.33 µs]
                        change: [-12.463% -12.239% -12.022%] (p = 0.00 < 0.05)
                        Performance has improved.

Teams                   time:   [200.00 ns 200.30 ns 200.61 ns]
                        change: [-51.953% -51.462% -51.036%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Todo:

- [x] update tests (e.g. ryu does not omits trailing `.0`s for floats)
- [x] test properly
- [x] add comments


Kijewski:
Hm, filters (including any explicit `|escape`) still invariantly causes writes through `fmt::Formatter`. I guess I can ignore the case for now because the this is a good speed-up for a start?

rinja/src/filters/mod.rs
```diff
@@ -2,6 +2,13 @@
 //!
 //! Contains all the built-in filter functions for use in templates.
 //! You can define your own filters, as well.
+//!
+//! ## Note
+//!
+//! All **result types of any filter function** in this module is **subject to change** at any
+//! point, and is **not indicated by as semver breaking** version bump.
+//! The traits [`AutoEscape`] and [`WriteWritable`] are used by [`rinja_derive`]'s generated code
+//! to work with all compatible types.
```
Kijewski, line 11: Adding this change is semver breaking in itself, but IMO its still worth it to keep our options open in the future or add more optimizations in the future, e.g. because writing filters is currently not optimized.

testing/tests/simple.rs
```diff
@@ -490,7 +490,7 @@ fn test_num_literals() {
     let template = NumLiterals;
     assert_eq!(
         template.render().unwrap(),
-        "[90, -90, 90, 2, 56, 240, 10.5, 10.5, 100000000000, 105000000000]\n1\n12",
+        "[90, -90, 90, 2, 56, 240, 10.5, 10.5, 100000000000.0, 105000000000.0]\n1\n12",
```
Kijewski, line 493: ryu adds a trailing `.0` for integral floats. IMO ryu gets it more right than rust's defaults, but we could also strip a trailing `.0` to remove the difference.

rinja_derive/src/tests.rs
```diff
@@ -73,11 +75,13 @@ impl ::std::fmt::Display for Foo {{
...
-        "{expr0}",
-        expr0 = &(&&::rinja::filters::AutoEscaper::new(&(s), ::rinja::filters::Text)).rinja_auto_escape()?,
-    )?;
+    match (
+        &((&&::rinja::filters::AutoEscaper::new(&(s), ::rinja::filters::Text)).rinja_auto_escape()?),
+    ) {
+        (expr0,) => {
+            (&&::rinja::filters::Writable(expr0)).rinja_write(writer)?;
+        }
+    }
```
Kijewski, line 84: Using a `match` instead of `{…}` block, we do not need to fear accidental shadowing, and the lifetimes are more easy to understand, IMO.

rinja_derive/src/generator.rs
```diff
@@ -1056,128 +1053,89 @@ impl<'a> Generator<'a> {
...
-
-    fn named_expression(
-        &mut self,
-        buf_expr: &mut Buffer,
-        buf_format: &mut Buffer,
-        expr: String,
-        wrapped: DisplayWrap,
-        cacheable: bool,
-        expr_cache: &mut HashMap<String, usize>,
-    ) -> usize {
```
Kijewski, line 1155: I inlined this function. In my opinion the new functions is small enough by itself.

rinja_derive/src/generator.rs
```diff
@@ -1056,128 +1053,89 @@ impl<'a> Generator<'a> {
...
+                        items.push(s);
+                        it.next();
+                    }
+                    lines.writeln(format_args!(
+                        "writer.write_str({:#?})?;",
+                        items.as_slice().join(""),
+                    ));
                 }
                 Writable::Expr(s) => {
+                    size_hint += 3;
```
Kijewski, line 1099: I kept this `size_hint`. I guess `3` was chosen by a fair dice roll? 

[](https://xkcd.com/221/)

GuillaumeGomez: HAHAHA. It's super strange indeed.

rinja_derive/src/generator.rs
```diff
@@ -1962,28 +1919,59 @@ struct Buffer {
...
+
+    fn write_writer(&mut self, s: &str) -> usize {
+        if self.discard {
+            // nothing to do
+        } else if !self.last_was_write_str {
+            writeln!(self.buf, "writer.write_str({s:#?})?;").unwrap();
+            self.last_was_write_str = true;
+        } else {
+            // strip trailing `")?\n`, leaving an unterminated string
+            let len = self.buf.strip_suffix("\")?;\n").unwrap().len();
```
GuillaumeGomez, line 1966: I'm worried about whether or not this `strip_suffix` will work on windows since it's `\r\n` and we use `writeln` a lot. Maybe we should stop printing backlines? It's not like they're really useful...

rinja_derive/src/tests.rs
GuillaumeGomez: Why not `if let` in this case?

testing/tests/simple.rs
GuillaumeGomez: I don't think it matters much. If people want to change it, they can always using the `format` filter.

rinja_derive/src/generator.rs
```diff
@@ -1962,28 +1919,59 @@ struct Buffer {
...
+            self.buf.truncate(len);
+            // append the new string, adding a stray `"` in the mid of the string
+            writeln!(self.buf, "{s:#?})?;").unwrap();
+            // left shift new string by one to overwrite the stray `"`
+            self.buf.replace_range(len..=len, "");
```
GuillaumeGomez, line 1971: The longer the string, the worse it gets. I think for this, it'd be better to keep a `String` in `self` and simply `clear` every time, then render into and then push the new string without the first `"` into `buf`. Like that, no need for the shift and it also greatly reduces the number of allocations.

rinja_derive/src/generator.rs
Kijewski: I removed any newlines generation. Until #87 the generated code cannot be compared anymore. #87 lets `syn` do the comparison, and `syn` does not care about spacing.

GuillaumeGomez: Can be rebased now. ;)

rinja_derive/src/generator.rs
Kijewski: Fixed in <https://github.com/rinja-rs/rinja/pull/86/commits/9ba70cbce8ba0ab8db173f52f724becef21c4411>. It has a healthy speed-up for other cases, too. The normal <str as fmt::Debug> is optimized for human readability, but we don't need it to be. Simply dump everything as is, except `\" \\ \r`.

rinja_derive/src/generator.rs
```diff
@@ -2286,3 +2286,26 @@ fn normalize_identifier(ident: &str) -> &str {
...
+pub(crate) fn string_escape(dest: &mut String, src: &str) {
+    // SAFETY: we will only push valid str slices
+    let dest = unsafe { dest.as_mut_vec() };
+    let src = src.as_bytes();
+    let mut last = 0;
+
+    // According to <https://doc.rust-lang.org/reference/tokens.html#string-literals>, every
+    // character is valid except `" \ IsolatedCR`. We don't test if the `\r` is isolated or not,
+    // but always escape it.
+    for x in memchr::memchr3_iter(b'\\', b'"', b'\r', src) {
```
GuillaumeGomez, line 2301: This is super weird to require escaping `\r`. Oh well, doesn't matter.

rinja_derive/src/generator.rs
```diff
@@ -124,64 +124,63 @@ impl<'a> Generator<'a> {
...
-        buf.writeln("::Result::Ok(())");
-        buf.writeln("}");
 
-        buf.writeln(format_args!(
-            "const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = {:?};",
+        buf.write(format_args!(
+            "\
+                {CRATE}::Result::Ok(()) \
+            }} \
+            const EXTENSION: ::std::option::Option<&'static ::std::primitive::str> = {:?}; \
```
GuillaumeGomez, line 132: I think it's safe to remove whitespace characters at the end of the "line" too. ;)

Kijewski: It is, removed. :)

rinja_derive/src/tests.rs
Kijewski: Rust's `derive(Debug, Clone)` etc. generate code using `match`es, so I did not really think about it. I guess it's OK either way?

GuillaumeGomez: `if let` is shorter but it's a nit/style prefer, I won't block this PR on this (because I don't really care :laughing:).

rinja/src/filters/escape.rs
```diff
@@ -309,6 +318,20 @@ const _: () = {
         NeedsEscaping(&'a T, E),
     }
 
+    impl<T, E> FastWritable for Wrapped<'_, T, E>
+    where
+        T: AsRef<str> + fmt::Display + ?Sized,
```
GuillaumeGomez, line 323: I think the `fmt::Display` bound can be removed since we only ever use `as_ref()`.

rinja/src/filters/mod.rs
GuillaumeGomez: As long as it doesn't break users code, I'm fine with it.

testing/tests/simple.rs
GuillaumeGomez: Hum... Thinking some more about it, people will expect us to use the Rust implementation to get the same output. I think it's problematic that `ryu` doesn't always display the same thing.

Here what I suggest: I talked with Mara and she said that we could use `ryu`'s algorithm directly in `std` to improve performance. I think it's a better way forward rather than having our own display, the performance difference is not big enough to be worth imo.

rinja/src/filters/escape.rs
Kijewski: Removed.

rinja/src/filters/escape.rs
```diff
@@ -435,13 +459,169 @@ impl<T: HtmlSafe + ?Sized> HtmlSafe for std::sync::RwLockWriteGuard<'_, T> {}
...
+
+/// Used internally by rinja to speed up writing some types.
+///
+/// Types implementing this trait can be written without needing to employ an [`fmt::Formatter`].
+pub trait FastWritable {
+    fn write_into<W: fmt::Write + ?Sized>(&self, dest: &mut W) -> fmt::Result;
+}
+
+const _: () = {
+    macro_rules! specialize_ref {
```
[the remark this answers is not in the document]
Kijewski: It is used just once just below. Should I maybe rename it into `impl_ref` for `impl_for_ref` to make the name more speaking?

testing/tests/simple.rs
Kijewski: The performance is much more impacted by the removal of the `&dyn Write`, the speed-up through `ryu` is minuscule in comparison. It's just the first implementation that came to mind.

But I think it's seldom that people will want to print floats, and then most likely with explicit formatting (`price|fmt(".02")`), so it won't matter much if I remove this specialization.

rinja_derive/src/generator.rs
```diff
@@ -2092,27 +2057,50 @@ struct Buffer {
...
     }
 
-    fn write(&mut self, src: impl BufferFmt) {
+    fn write_writer(&mut self, s: &str) -> usize {
+        const OPEN: &str = r#"writer.write_str(""#;
+        const CLOSE: &str = r#"")?;"#;
+
         if !self.discard {
-            src.append_to(&mut self.buf);
+            if !self.last_was_write_str {
```
GuillaumeGomez, line 2093: This change in particular will improve things SO MUCH. It's a really good idea!

rinja/src/filters/escape.rs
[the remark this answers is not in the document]
Kijewski: "Either … or …", well, I did both :D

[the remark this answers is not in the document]
GuillaumeGomez: Haha. Well, it's the best solution I think. :D

rinja/src/filters/escape.rs
```diff
@@ -435,13 +454,159 @@ impl<T: HtmlSafe + ?Sized> HtmlSafe for std::sync::RwLockWriteGuard<'_, T> {}
...
+/// Used internally by rinja to speed up writing some types.
+///
+/// Types implementing this trait can be written without needing to employ an [`fmt::Formatter`].
+pub trait FastWritable {
+    fn write_into<W: fmt::Write + ?Sized>(&self, dest: &mut W) -> fmt::Result;
+}
+
+const _: () = {
+    // implement FastWritable for a list of reference wrapper types to FastWritable+?Sized
+    macro_rules! impl_for_ref {
```
GuillaumeGomez, line 483: :+1:

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

PR #87, opened 2024-07-22
Every used template gets referenced exactly once

### LINKED ISSUE #72: Only use `include_byte!` on a file once across templates

Kijewski:
Once per `#[template]`, that is. Still multiple times per crate.

Do you think that is enough to resolve #72, @GuillaumeGomez?


rinja_derive/Cargo.toml
```diff
@@ -33,4 +33,10 @@ once_map = "0.4.18"
...
 quote = "1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-syn = "2"
+syn = "2.0.3"
+
+[dev-dependencies]
+console = "0.15.8"
+similar = "2.6.0"
+prettyplease = "0.2.20"
+syn = { version = "2.0.3", features = ["extra-traits", "full"] }
```
Kijewski, line 42: Feel free to tell me if you think this is overkill. I thought "lets make it pretty", but maybe I overdid it a bit :)

GuillaumeGomez:
I think it's definitely not worth it for "by template" approach. I also wondered about it myself but it really only makes sense for multiple templates. But as you underlined, the only solution for that would be `tracked_file`. I'll see what's missing for it to be done.

Kijewski:
I think [this diff](https://github.com/rinja-rs/rinja/pull/87/commits/55774ff36e8cd173d871af81ba38196356147c46#diff-e436fa3544e071f9efdd483195c428ba49a1fad123c3e6edfcb66383b51fdff1L151-L165) might still make it worthwhile. Doesn't it?

rinja_derive/src/tests.rs
```diff
@@ -92,5 +138,25 @@ impl ::std::fmt::Display for Foo {{
...
+    compare(
+        r#"{% include "include1.html" %}"#,
+        &format!(
+            r#"const _: &[::core::primitive::u8] = ::core::include_bytes!({path1:#?});
+            const _: &[::core::primitive::u8] = ::core::include_bytes!({path2:#?});
+            const _: &[::core::primitive::u8] = ::core::include_bytes!({path3:#?});
+            writer.write_str("3")?;
+            writer.write_str("3")?;
+            writer.write_str("3")?;
+            writer.write_str("3")?;"#
```
Kijewski, line 158: Will be joined into one call with <https://github.com/rinja-rs/rinja/pull/86/commits/3c08d0b99f55919fed821058bb9578b4f7eba7b7>.

rinja_derive/Cargo.toml
GuillaumeGomez: It's dev dependencies, so I definitely don't care. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Can you just provide a bench comparison in addition to the improved output please?

Kijewski:
<details>
<summary><code>rinja_derive_standalone</code> benchmark</summary>

```text
hello_world             time:   [48.183 µs 48.253 µs 48.314 µs]
                        change: [-6.5054% -5.3650% -4.2704%] (p = 0.00 < 0.05)
                        Performance has improved.

item_info.html          time:   [58.699 µs 59.057 µs 59.467 µs]
                        change: [-1.5808% -0.9997% -0.4222%] (p = 0.00 < 0.05)
                        Change within noise threshold.

item_union.html         time:   [152.77 µs 152.92 µs 153.11 µs]
                        change: [-3.7432% -3.4421% -3.1460%] (p = 0.00 < 0.05)
                        Performance has improved.

page.html               time:   [651.01 µs 651.83 µs 652.81 µs]
                        change: [-3.8608% -3.5279% -3.1711%] (p = 0.00 < 0.05)
                        Performance has improved.

print_item.html         time:   [141.53 µs 142.07 µs 142.73 µs]
                        change: [-1.4684% -0.6647% +0.4653%] (p = 0.24 > 0.05)
                        No change in performance detected.

short_item_info.html    time:   [122.21 µs 122.49 µs 122.86 µs]
                        change: [-2.3561% -1.9451% -1.5023%] (p = 0.00 < 0.05)
                        Performance has improved.

sidebar.html            time:   [193.36 µs 195.57 µs 197.54 µs]
                        change: [+0.4147% +1.2141% +2.0348%] (p = 0.00 < 0.05)
                        Change within noise threshold.

source.html             time:   [105.18 µs 105.75 µs 106.48 µs]
                        change: [+1.3040% +2.5061% +3.6714%] (p = 0.00 < 0.05)
                        Performance has regressed.

type_layout.html        time:   [176.73 µs 176.98 µs 177.27 µs]
                        change: [-1.6070% -1.1707% -0.7469%] (p = 0.00 < 0.05)
                        Change within noise threshold.

type_layout_size.html   time:   [64.969 µs 65.371 µs 65.771 µs]
                        change: [-3.4571% -2.7019% -1.8661%] (p = 0.00 < 0.05)
                        Performance has improved.
```
</details>

IMO the results are spurious. We don't have a benchmark that uses included files, so the PR should not affect the results in one way or another, except that I replaced one call to `quote!` with `format_args!`.

GuillaumeGomez:
What I was interested about was whether or not we had a perf regression. Since it's not the case, all good, thanks!

PR #90, opened 2024-07-25
parser: use `memchr` to speed-up `skip_till()`

### LINKED ISSUE #88: parser: use `memchr` to speed-up `skip_till()`
`skip_till()` is used in the parser to find the next block `{%`, comment `{#` or expression `{{`. At every character position, it is tested if one of these three substrings follows. Using [`memchr3()`](https://docs.rs/memchr/2.7.4/memchr/fn.memrchr3.html), we could at least skip to the next candidate, a `{`. The syntax for blocks, comments and expressions can be modified by the user, but that does not matter much; we can simply supply `memchr3()` with the first byte in each of these strings.

I did try if [`aho-corasick`](https://crates.io/crates/aho-corasick) was a good option, but it was (at least how I used it) slower. Most likely because building the DFA at runtime is not exactly for free, and because the algorithm can only show its full potential when it tries to search for more or longer needles than 3 needles 2 à bytes.

Kijewski:
`skip_till()` is used in the parser to find the next block `{%`, comment `{#` or expression `{{`. At every character position, it is tested if one of these three substrings follows. Using [`memchr3()`], we could at least skip to the next candidate, a `{`. The syntax for blocks, comments and expressions can be modified by the user, but that does not matter much; we can simply supply `memchr3()` with the first byte in each of these strings.

[`memchr3()`]: <https://docs.rs/memchr/2.7.4/memchr/fn.memrchr3.html>

<details>
<summary>Benchmark results (<code>cd rinja_parser; cargo bench</code>)</summary>

```text
librustdoc/all          time:   [366.54 µs 366.78 µs 367.02 µs]
                        thrpt:  [38.475 MiB/s 38.500 MiB/s 38.525 MiB/s]
                 change:
                        time:   [-17.358% -17.065% -16.820%] (p = 0.00 < 0.05)
                        thrpt:  [+20.221% +20.576% +21.004%]
                        Performance has improved.

librustdoc/item_info    time:   [6.3315 µs 6.3400 µs 6.3495 µs]
                        thrpt:  [24.783 MiB/s 24.820 MiB/s 24.853 MiB/s]
                 change:
                        time:   [-6.5547% -6.4090% -6.2633%] (p = 0.00 < 0.05)
                        thrpt:  [+6.6818% +6.8479% +7.0144%]
                        Performance has improved.

librustdoc/item_union   time:   [39.377 µs 39.551 µs 39.720 µs]
                        thrpt:  [24.850 MiB/s 24.957 MiB/s 25.067 MiB/s]
                 change:
                        time:   [-6.9834% -6.2455% -5.5849%] (p = 0.00 < 0.05)
                        thrpt:  [+5.9153% +6.6616% +7.5077%]
                        Performance has improved.

librustdoc/page         time:   [170.83 µs 170.99 µs 171.23 µs]
                        thrpt:  [36.164 MiB/s 36.213 MiB/s 36.248 MiB/s]
                 change:
                        time:   [-12.413% -12.183% -11.968%] (p = 0.00 < 0.05)
                        thrpt:  [+13.595% +13.873% +14.173%]
                        Performance has improved.

librustdoc/print_item   time:   [21.163 µs 21.234 µs 21.322 µs]
                        thrpt:  [44.280 MiB/s 44.463 MiB/s 44.612 MiB/s]
                 change:
                        time:   [-19.848% -18.613% -17.491%] (p = 0.00 < 0.05)
                        thrpt:  [+21.198% +22.870% +24.763%]
                        Performance has improved.

librustdoc/short_item_info
                        time:   [19.781 µs 19.813 µs 19.846 µs]
                        thrpt:  [45.652 MiB/s 45.727 MiB/s 45.801 MiB/s]
                 change:
                        time:   [-18.027% -17.806% -17.574%] (p = 0.00 < 0.05)
                        thrpt:  [+21.321% +21.663% +21.991%]
                        Performance has improved.

librustdoc/sidebar      time:   [40.694 µs 40.806 µs 40.957 µs]
                        thrpt:  [30.131 MiB/s 30.242 MiB/s 30.325 MiB/s]
                 change:
                        time:   [-14.698% -14.069% -13.456%] (p = 0.00 < 0.05)
                        thrpt:  [+15.548% +16.372% +17.231%]
                        Performance has improved.

librustdoc/source       time:   [15.249 µs 15.264 µs 15.278 µs]
                        thrpt:  [48.251 MiB/s 48.295 MiB/s 48.343 MiB/s]
                 change:
                        time:   [-25.832% -25.678% -25.532%] (p = 0.00 < 0.05)
                        thrpt:  [+34.285% +34.550% +34.829%]
                        Performance has improved.

librustdoc/type_layout_size
                        time:   [9.0168 µs 9.0446 µs 9.0789 µs]
                        thrpt:  [29.832 MiB/s 29.945 MiB/s 30.038 MiB/s]
                 change:
                        time:   [-11.100% -10.437% -9.4426%] (p = 0.00 < 0.05)
                        thrpt:  [+10.427% +11.653% +12.486%]
                        Performance has improved.

librustdoc/type_layout  time:   [34.088 µs 34.114 µs 34.139 µs]
                        thrpt:  [78.860 MiB/s 78.919 MiB/s 78.979 MiB/s]
                 change:
                        time:   [-37.865% -37.723% -37.585%] (p = 0.00 < 0.05)
                        thrpt:  [+60.217% +60.573% +60.941%]
                        Performance has improved.
```
</details>

Resolves #88.

### DIFF
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -20,6 +20,7 @@ use nom::{error_position, AsChar, InputTakeAtPosition};
 
 pub mod expr;
 pub use expr::{Expr, Filter};
+mod memchr_splitter;
 pub mod node;
 pub use node::Node;
 
@@ -362,22 +363,22 @@ fn ws<'a, O>(
 
 /// Skips input until `end` was found, but does not consume it.
 /// Returns tuple that would be returned when parsing `end`.
-fn skip_till<'a, O>(
+fn skip_till<'a, 'b, O>(
+    candidate_finder: impl crate::memchr_splitter::Splitter,
     end: impl FnMut(&'a str) -> ParseResult<'a, O>,
 ) -> impl FnMut(&'a str) -> ParseResult<'a, (&'a str, O)> {
-    enum Next<O> {
-        IsEnd(O),
-        NotEnd,
-    }
-    let mut next = alt((map(end, Next::IsEnd), map(anychar, |_| Next::NotEnd)));
+    let mut next = alt((map(end, Some), map(anychar, |_| None)));
     move |start: &'a str| {
         let mut i = start;
         loop {
-            let (j, is_end) = next(i)?;
-            match is_end {
-                Next::IsEnd(lookahead) => return Ok((i, (j, lookahead))),
-                Next::NotEnd => i = j,
-            }
+            i = match candidate_finder.split(i) {
+                Some((_, j)) => j,
+                None => return Err(nom::Err::Error(ErrorContext::new("`end` not found`", i))),
+            };
+            i = match next(i)? {
+                (j, Some(lookahead)) => return Ok((i, (j, lookahead))),
+                (j, None) => j,
+            };
         }
     }
 }
--- a/rinja_parser/src/memchr_splitter.rs
+++ b/rinja_parser/src/memchr_splitter.rs
@@ -1,0 +1,101 @@
+pub(crate) trait Splitter: Copy {
+    /// If any of the needles was found in the haystack, then split the haystack at the first hit.
+    ///
+    /// Since only the first byte of a needle is inspected, be aware that there can be
+    /// false-positives. Always compare the latter string of the output if it fits the expected
+    /// prefix.
+    fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)>;
+}
+
+impl<T: Splitter + ?Sized> Splitter for &T {
+    #[inline]
+    fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)> {
+        T::split(self, haystack)
+    }
+}
+
+// define and implement a string splitter using memchr
+macro_rules! new_memchr_type {
+    ($struct:ident $split_unchecked:ident $memchr:ident $($field:ident)*) => {
+        #[derive(Debug, Clone, Copy)]
+        pub(crate) struct $struct {
+            $($field: u8,)*
+        }
+
+        impl $struct {
+            #[track_caller]
+            pub(crate) fn new($($field: &str),*) -> Self {
+                Self {
+                    $($field: $field.as_bytes()[0],)*
+                }
+            }
+
+            #[inline]
+            pub(crate) fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)> {
+                // SAFETY: During the construction of `self` we used strings as inputs, and a
+                // string always starts with a byte at char boundary.
+                unsafe { $split_unchecked($(self.$field,)* haystack) }
+            }
+        }
+
+        impl Splitter for $struct {
+            #[inline]
+            fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)> {
+                self.split(haystack)
+            }
+        }
+
+        /// SAFETY: caller has to ensure that the needle is at a char boundary
+        pub(crate) unsafe fn $split_unchecked(
+            $($field: u8,)*
+            haystack: &str,
+        ) -> Option<(&str, &str)> {
+            let idx = memchr::$memchr($($field,)* haystack.as_bytes())?;
+            // SAFETY: The caller ensures that the needles are at char boundary.
+            // The found index `< haystack.len()`.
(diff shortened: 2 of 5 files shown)


rinja_parser/src/lib.rs
```diff
@@ -363,25 +363,78 @@ fn ws<'a, O>(
...
+            }
+
+            impl CandidateFinder for CandidateFinderN {
+                #[inline]
+                fn next(self, haystack: &str) -> Option<&str> {
+                    let haystack = haystack.as_bytes();
+                    let i = self.0.find(haystack)?;
+                    // SAFETY: We know that `memchr` returns a valid position:
+                    // The needles can only be ASCII, or the start of a UTF-8 sequence, so
+                    // only an ASCII character or the start of a UTF-8 sequence can be follow.
```
GuillaumeGomez, line 424: "can be follow"?

GuillaumeGomez:
PR looks good. Just a few things I need to confirm: you mentioned that `memchr` only works on ASCII characters. However, we can overload syntax (like comment start/end for example). Nothing is preventing users to use non-ASCII characters for that. Will it still work in this case? If yes, please add tests to ensure it.

rinja_parser/src/lib.rs
Kijewski: "Can be" that I should read my comments again after writing them. :)

GuillaumeGomez: That's what reviews are meant for. :D

Kijewski:
> you mentioned that memchr only works on ASCII characters

`memchr` works on any `&[u8]`. It does not know or care what the input it.

But we have to care, because we must not split a `&str`. The first byte of a UTF-8 encoded code point is either:
- `0b0xxx_xxxx` (ASCII, 1 byte),
- `0b110x_xxxx` (BMP, 2 bytes),
- `0b1110_xxxx` (3 bytes) or
- `0b1111_0xxx` (4 bytes),
- but never `0b10xx_xxxx`.

Knowing that, we know that the result of `memchr` will never point into a UTF-8 sequence, only at the start of one, because there is no way that `memchr` could have searched for a byte inside a UTF-8 sequence.

Kijewski:
Rebas on the current HEAD. Some other PR did make the parsing faster in the meantime, too, so I re-did the benchmark. Updated the initial comment. Still a good speed-up.

rinja_parser/src/memchr_splitter.rs
```diff
@@ -0,0 +1,97 @@
+pub(crate) trait Splitter: Copy {
+    /// The any of the needles was found in the haystack, then split the haystack at the first hit.
```
GuillaumeGomez, line 2: I think something went wrong with this explanation. XD

Missing a word maybe between `The` and `any` at the beginning?

rinja_parser/src/tests.rs
```diff
@@ -371,8 +371,31 @@ fn change_delimiters_parse_filter() {
         expr_end: "=}",
         ..Syntax::default()
     };
-
     Ast::from_str("{= strvar|e =}", None, &syntax).unwrap();
+
+    let syntax = Syntax {
+        expr_start: "🖎", // U+1F58E == b"\xf0\x9f\x96\x8e"
```
GuillaumeGomez, line 377: Nice :+1:

Please just add a comment on this one explaining it ensures that unicode characters can be used as expression markers.

rinja_parser/src/memchr_splitter.rs
Kijewski: "The" should have been "if". Common typo, the "the" key is right next to the "if" key. 🙈

rinja_parser/src/tests.rs
Kijewski: Do you think that a speaking name is enough?

rinja_parser/src/memchr_splitter.rs
GuillaumeGomez: Obviously 😛

rinja_parser/src/tests.rs
GuillaumeGomez: Yep, the function name makes it obvious for me at least.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #91, opened 2024-07-26
Add handling of boolean literals to condition evaluation

GuillaumeGomez:
This PR always checks for booleans in `condition_evaluation` function. One thing I thought about: we could skip this function entirely if we don't have `is (not) defined` inside a condition. However, it'd require to either go through all items of the conditions to see if one of them is `is (not) defined` or to pass a `&mut bool` to check if the expression encounters one of them.

Another thing I thought: instead of handling the evaluation in the generator, we could potentially handle it in the parser directly (and therefore, not having the branches in the AST if the previous `if` excluded it). Not sure if it's a good idea or not though.

Anyway, this PR is much simpler than that since it only adds checks for boolean literals. :)

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -372,8 +372,7 @@ impl<'a> Generator<'a> {
         only_contains_is_defined: &mut bool,
     ) -> EvaluatedResult {
         match **expr {
-            Expr::BoolLit(_)
-            | Expr::NumLit(_)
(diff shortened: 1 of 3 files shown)


Kijewski:
Good idea! Except for the missing newline, approved. :)

GuillaumeGomez:
Woups. Fixed. :laughing:

Kijewski (approved):

PR #92, opened 2024-07-27
derive: escape strings at compile-time when possible

Kijewski:
Todo:

- [x] look if there is a proper string parsing function
- [x] benchmark derive

PR #93, opened 2024-07-28
Speed-up html-escaping using jetscii

Kijewski:
Rebased on #92.

---

```text
$ cargo bench --bench escape

Before the PR:           [3.6464 µs 3.6512 µs 3.6564 µs]
Impl. without `jetscii`: [3.4837 µs 3.4899 µs 3.4968 µs] ~ 95 % of baseline
Impl with `jetscii`:     [2.0264 µs 2.0335 µs 2.0418 µs] ~ 55 % of baseline
```

Until portable SIMD gets stabilized, I don't think we can do much for non-X86 platforms. And even after it is stabilized, I guess any optimizations should be implemented upstream in memchr and/or jetscii.


rinja/src/html.rs
```diff
@@ -0,0 +1,203 @@
...
+    // Instead of testing the platform, we could test the CPU features. But given that the needed
+    // instruction set SSE 4.2 was instroduced in 2008, that it has an 99.61 % avalibility rate
+    // in Steam's June 2024 hardware survey, and is a prerequisite to run Windows 11, I don't
+    // think we need to care.
+
+    let mut escaped_buf = ESCAPED_BUF_INIT;
+    let mut last = 0;
+
+    for (index, byte) in string.bytes().enumerate() {
+        let escaped = match byte {
```
Kijewski, line 19: Using a bit test in here makes the benchmarks run slower. It could be that the benchmarks outline an unrealistic scenario, though: They consist of comparatively long strings.

Bit testing is slower than a table lookup, if the table is already loaded into the CPU cache, but loading the table takes quite some time \_once\_, too. If the strings are mostly short, and mostly contain no characters that need escaping, then a bit test solution would win.

I guess the current solution is good enough for now. Just something we need to keep in mind, that we might need to revise the benchmark text corpus. Even the "short" string is 27 characters long.

GuillaumeGomez:
I'll take a look once rebased. ;)

Kijewski:
Rebased. For now, that's the last one in my "let's make it speedy" series. :)

rinja/src/html.rs
```diff
@@ -1,71 +1,203 @@
...
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
 #[allow(unused)]
 pub(crate) fn write_escaped_str(mut fmt: impl fmt::Write, string: &str) -> fmt::Result {
-    let mut escaped_buf = *b"&#__;";
+    // Even though [`jetscii`] ships a generic implementation for unsupported platforms,
+    // it is not well optimized for this case. This implementation should work well enough in
+    // the meantime, until portable SIMD gets stabilized.
+
+    // Instead of testing the platform, we could test the CPU features. But given that the needed
+    // instruction set SSE 4.2 was instroduced in 2008, that it has an 99.61 % avalibility rate
```
GuillaumeGomez, line 11:
```suggestion
    // instruction set SSE 4.2 was introduced in 2008, that it has an 99.61 % availability rate
```

This word (avalibili thing) is super hardcore to write... Took me a few tries to get it right. XD

GuillaumeGomez, line 11: One more thought: wouldn't that prevent rinja to be used on embedded targets?

Kijewski: I only know of Intel Galileo that uses an x86 platform for a SOC. Indeed, it does not feature any SSE instructions, but Galileo was never really a thing.

For the case that `jetscii` cannot use SSE 4.2, it implements a fallback. Unless explicitly specified as present or absent in the rustc flags, `jetscii` will query the CPU features at runtime if SSE 4.2 can be used or if the fallback has to be used. The x86-version with `jetscii`'s fallback is slower than our non-X86 version of `write_escaped_str()`, though.

We could add a build script to check if SSE 4.2 was explicitly marked as absent during the compilation, and use the generic version on x86 then, too. But I don't think the case is very likely, and I would rather wait if someone mentions that they want to use rinja on an embedded x86 system.

GuillaumeGomez: Not sure if it's really that useful then if it's only for x86... Might be better to wait for portable simd to get stabilized.

Kijewski:
I excluded x86 (32bit). That makes the code a bit more readable, and if you still use a 32bit x86 system in 2024, then yeah, chances are that you run your code on some aching VIA processor without SSE 4.2 support.

rinja/src/html.rs
Kijewski: Yeah, chances are that you don't drop MBs worth of untrusted data into your templates. I guess most of the time it will be less than 16 bytes, so jetscii would not come into action anyway, but I guess different users have different use cases.

Should I copy the changed generic function into a new PR? 5% faster is not much, but I think the new code is a little more readable, too.

GuillaumeGomez: More readable AND faster? Yes please.

PR #97, opened 2024-07-28
Fix 404 page on readthedocs

### LINKED ISSUE #96: book: the 404-page is broken, wrong base URL
<https://rinja.readthedocs.io/en/latest/ronja.html>

The base URL of the 404-page seems broken. Relative links point at `https://rinja.readthedocs.io/*` instead of `https://rinja.readthedocs.io/en/latest/*`, so the assets cannot be loaded, and the menu on the left only brings you to more 404-pages.

Do we (need to) care?

GuillaumeGomez:
Fixes #96.

The first commit adds a 404 page as described in [readthedocs docs](https://docs.readthedocs.io/en/stable/reference/404-not-found.html) and the second one fixes the "Rinja - Rinja" title for the first chapter (which becomes "Rinja - Introduction").


Kijewski (approved):
Thanks!

PR #98, opened 2024-07-28
Copy `404.html` file to readthedocs output dir

GuillaumeGomez:
Seems like  #97 was not enough: I forgot to copy the `404.html` file in the `.readthedocs.yml` file too.


Kijewski (approved):

Kijewski:
The "This is the latest development version" banner is printed across the "Oops", but that's okay. It has a link to the latest version, too. Thanks for your follow up PR!

PR #99, opened 2024-07-28
Change the error message to make it start at the `as` keyword

GuillaumeGomez:
Follow-up of https://github.com/rinja-rs/rinja/pull/63.

I thought about the case where there is a big expression before the `as` keyword. In this case, it's really not clear when we show everything what's actually wrong. It renders better when we make it start at the `as` keyword so we still keep the keyword to give context but remove the "noise" coming from the expression so the user can see better what's going on.

If it makes sense?

I added a test case to illustrate it if you want to compare. ;)


Kijewski (approved):
Yes, makes sense. :+1:

PR #102, opened 2024-07-29
derive:  better use of spans and file information in error messages

Kijewski:


testing/tests/ui/filter_block_ws.stderr
```diff
@@ -1,5 +1,9 @@
-error: failed to parse template source at row 1, column 27 near:
-       " %}\nHELLO\n{{v}}\n{%- endfilter %}"
+error: failed to parse template source
+ --> <source attribute>:1:28
```
GuillaumeGomez, line 2: For this one we could actually get the current file name.

Kijewski: I don't think that we from which file we were called, do we? Not until <https://github.com/rust-lang/rust/issues/54725> lands. And the path is written just below this error message, so I don't think that is matters much.

rinja_derive/src/input.rs
```diff
@@ -149,7 +152,21 @@ impl TemplateInput<'_> {
...
+            let parsed = match self.syntax.parse(Arc::clone(&source), source_path) {
+                Ok(parsed) => parsed,
+                Err(err) => {
+                    let msg = err
+                        .message
+                        .unwrap_or_else(|| "failed to parse template source".into());
+                    let file_path = err
+                        .file_path
+                        .as_deref()
+                        .unwrap_or(Path::new("<source attribute>"));
```
GuillaumeGomez, line 164: If we have the `source` attribute span, can't we use it instead?

Kijewski: To underline `source` or the value of `source` instead of the `template` token? I think this could get confusion if the error happens in a transcluded file.

GuillaumeGomez: Hum... I think we can know if it comes from a file included. So there are two cases:

1. From the source directly, then we can underline the `source` attribute saying "it comes from template from this attribute"
2. From a file included in the source, then we still underline the `source` attribute and we say "it comes from a template included from this attribute"

What do you think?

Kijewski: Yes, makes sense. Changed.

testing/tests/ui/as-primitive-type.stderr
```diff
@@ -1,59 +1,65 @@
 error: `as` operator expects the name of a primitive type on its right-hand side
-       failed to parse template source at row 1, column 8 near:
-       "as 4567 }}"
- --> tests/ui/as-primitive-type.rs:3:10
+ --> <source attribute>:1:9
   |
-3 | #[derive(Template)]
-  |          ^^^^^^^^
+       1 | {{ 1234 as 4567 }}
```
GuillaumeGomez, line 4: Btw, is it normal to have this super wide space starting here?

rinja_derive/src/config.rs
```diff
@@ -319,13 +323,15 @@ impl<'a> TryInto<Syntax<'a>> for RawSyntax<'a> {
...
+                    format!("delimiters must be at least two characters long: {s:?}"),
+                    None,
+                ));
             } else if s.chars().any(|c| c.is_whitespace()) {
-                return Err(CompileError::no_file_info(format!(
-                    "delimiters may not contain white spaces: {s:?}"
-                )));
+                return Err(CompileError::no_file_info(
+                    format!("delimiters may not contain white spaces: {s:?}"),
+                    None,
```
GuillaumeGomez, line 333: We should show the config span here.

rinja_derive/src/config.rs
```diff
@@ -335,9 +341,12 @@ impl<'a> TryInto<Syntax<'a>> for RawSyntax<'a> {
...
         ] {
             if s1.starts_with(s2) || s2.starts_with(s1) {
-                return Err(CompileError::no_file_info(format!(
-                    "a delimiter may not be the prefix of another delimiter: {s1:?} vs {s2:?}",
-                )));
+                return Err(CompileError::no_file_info(
+                    format!(
+                        "a delimiter may not be the prefix of another delimiter: {s1:?} vs {s2:?}",
+                    ),
+                    None,
```
GuillaumeGomez, line 348: We should show the config span here.

testing/tests/ui/as-primitive-type.stderr
Kijewski: It is aligned to start after "error: ", because it belongs to the rust source that is printed below. At first, I did not find out how to remove the indentation, now I kind like it, because it given you a visual "this happened because of that" hint.

GuillaumeGomez: I have to admit I hate it. ^^'

I'd prefer if we kept coherency with rustc (with columns aligned to the one with the biggest line error number):

```
$ rustc foo.rs 
error[E0425]: cannot find value `Ba` in this scope
  --> foo.rs:10:13
   |
7  | pub struct Bar;
   | --------------- similarly named unit struct `Bar` defined here
...
10 |     let x = Ba;
   |             ^^ help: a unit struct with a similar name exists: `Bar`

error: aborting due to 1 previous error
```

rinja_derive/src/config.rs
Kijewski: Done.

Kijewski:
<https://github.com/rinja-rs/rinja/pull/102/commits/821a147b37c14ebf005ac79f9b9f335ee576a5f8>. In my terminal I can see the messages, but they aren't captured by trybuild or vs code. I guess that proc_macros aren't supposed to just print something into stderr. Maybe I have to wrap the output in some magic message, but if so, I have no idea what.

rinja_derive/src/lib.rs
```diff
@@ -92,20 +94,33 @@ pub fn derive_template(input: TokenStream12) -> TokenStream12 {
...
-        Err(e) => {
-            let mut e = e.into_compile_error();
+        Err(CompileError {
+            msg,
+            span,
+            rendered,
+        }) => {
+            let msg = if rendered {
+                eprintln!("{msg}");
+                "the previous template error derives from"
```
GuillaumeGomez, line 104: Funny that only this part of the error message is displayed. Dark magic.

Kijewski:
I removed the formatting/pretty printing of the annotation in the last commit. Maybe we can merge this stuff for now? Even without the pretty output, it has a ton of minor improvements. I would then open a new PR to get the pretty rendering going.

testing/tests/ui/broken-config.stderr
```diff
@@ -0,0 +1,25 @@
+error: `$WORKSPACE/target/tests/trybuild/rinja_testing/no-such-config.toml` does not exist
+ --> tests/ui/broken-config.rs:4:38
+  |
+4 | #[template(source = "", ext = "txt", config = "no-such-config.toml")]
```
GuillaumeGomez, line 4: I've been thinking quite a lot about this and I think we should use the span of the value and not of the key (so not `source` and `config` but the string literal). Because, especially in this case, it seems like we're saying there is something wrong with the attribute and not with its content.

Kijewski: Yeah, I highlighted the key, not the value, because the key is shorter, and the value can be arbitrarily long. But I guess that's not a good reasoning. Changed.

GuillaumeGomez: It's still they key here. :wink:

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good, thanks a lot!

PR #103, opened 2024-07-30
derive: properly display error locations (second attempt)

Kijewski:
For now, this PR does nothing than to revert 08674bde1183b9224073a3abc1cbb3cdde6eb6b6. It still has the same problem, that the error output is not picked up by tools like trybuild or vs code.


Kijewski:
Superseded by <https://github.com/askama-rs/askama/pull/525>.

Issue #101, opened 2024-07-28
Ready for v0.3.0?


Kijewski:
We have added and changed [a lot](https://github.com/rinja-rs/rinja/compare/v0.2.0...d39ee220fd23633f8bf401cbf76d88d3d76f58bf) since the first release. Do you think we are ready for 0.2.1? I think rinja_parser has to get bumped to 0.3.0, because we added removed enum members, but the current HEAD should be compatible to rinja + rinja_derive 0.2.0.

What do you think, @GuillaumeGomez? Do you have something you wish to add or change before we do the next release?

GuillaumeGomez:
Actually yes! I wanted to add [slice pattern matching](https://github.com/djc/askama/issues/342) before a new release. But that's the only thing I can think of on my side.

Kijewski:
Yes, that can be quite a useful feature, too! I guess the parsing is 99% the same as parsing tuples, except for the delimiters.

Kijewski:
Or do you want to implement `if let [b'H', b'i', b' ', who @ .., b'!'] = b"Hi world!"`, too? I guess that could be difficult. <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ea373fd285476e52d131a83724293372>

GuillaumeGomez:
I suppose the whole `@ ..` should be implemented. Gonna be fun. :3

GuillaumeGomez:
Should we make the release now (I'll let you make this one :stuck_out_tongue:)? Gonna work on the `b`/`c` prefixes later on. Gonna open an issue for that.

Kijewski:
Will do a release later this evening (CEST).  :)

GuillaumeGomez:
Yeay! \o/

Kijewski:
Released: https://github.com/rinja-rs/rinja/releases/tag/v0.3.0 :)

GuillaumeGomez:
Wow looks nice! I wonder if we could automate that too. XD

Anyway, thanks a lot, it's awesome! Announcing on socials.

Kijewski:
When you click on [Releases](https://github.com/rinja-rs/rinja/releases), then on the top-right there is a button "draft new release". There you can select the newest tag, and with "Generate release notes" you get the same output as in my release message. :)

Unless I forget it, I check the "Create a discussion for this release" box, so the easier to answer questions or any comments to the release can get consolidated into one page

GuillaumeGomez:
[...]

PR #104, opened 2024-07-30
Add support for rest pattern

GuillaumeGomez:
As discussed in #101, I added support for it.

### DIFF
--- a/rinja_parser/src/target.rs
+++ b/rinja_parser/src/target.rs
@@ -238,7 +238,7 @@ fn only_one_rest_pattern<'a>(
         if let Target::Rest(s) = target {
             if !allow_named_rest && s.inner.is_some() {
                 return Err(nom::Err::Failure(ErrorContext::new(
-                    format!("`@ ..` cannot be used in {type_kind}"),
+                    "`@ ..` is only allowed in slice patterns",
                     s.span,
                 )));
             }
--- a/rinja_parser/src/target.rs
+++ b/rinja_parser/src/target.rs
@@ -3,17 +3,18 @@ use nom::bytes::complete::tag;
 use nom::character::complete::{char, one_of};
 use nom::combinator::{consumed, map, map_res, opt};
 use nom::multi::separated_list1;
-use nom::sequence::{pair, preceded};
+use nom::sequence::{pair, preceded, tuple};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


GuillaumeGomez:
Rebased and fixed clippy. CI is finally happy. :)

rinja_parser/src/target.rs
```diff
@@ -195,19 +227,29 @@ fn collect_targets<'a, T>(
...
+    allow_named_rest: bool,
+    type_kind: &str,
+) -> Result<Vec<Target<'a>>, ParseErr<'a>> {
+    let mut found_rest = false;
+
+    for target in &targets {
+        if let Target::Rest(s) = target {
+            if !allow_named_rest && s.inner.is_some() {
+                return Err(nom::Err::Failure(ErrorContext::new(
+                    format!("`@ ..` cannot be used in {type_kind}"),
```
Kijewski, line 241: Rust uses "`@ ..` is only allowed in slice patterns". I like that better, because it tells you what you can do, instead of [what you can't do](https://lostpedia.fandom.com/wiki/Don%27t_tell_me_what_I_can%27t_do).

The code looks fine, so if you like the current phrasing better, then it's not a showstopper for me!

GuillaumeGomez: No you're right, I prefer your version. Updating!

GuillaumeGomez:
And done. :)

rinja_parser/src/target.rs
```diff
@@ -195,19 +227,29 @@ fn collect_targets<'a, T>(
...
+    allow_named_rest: bool,
+    type_kind: &str,
+) -> Result<Vec<Target<'a>>, ParseErr<'a>> {
+    let mut found_rest = false;
+
+    for target in &targets {
+        if let Target::Rest(s) = target {
+            if !allow_named_rest && s.inner.is_some() {
+                return Err(nom::Err::Failure(ErrorContext::new(
+                    format!("`@ ..` is only allowed in slice patterns"),
```
Kijewski, line 241: The `format!()` is not needed anymore.

GuillaumeGomez: Even clippy was complaining about it. :3

Kijewski (approved):
Nice! I was afraid of a far bigger diff.

PR #105, opened 2024-07-31
Clarify that rinja doesn't depend on askama but is a fork

GuillaumeGomez:
Someone mentioned that the phrasing was a bit confusing. Fixed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
 # rinja
 
-**Rinja** is based on [Askama](https://crates.io/crates/askama).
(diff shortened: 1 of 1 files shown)


Kijewski:
Maybe reshuffle the paragraph? "Rinja implements […]. Rinja is a fork of Askama. See below for an example."?

GuillaumeGomez:
Good idea, done!

Kijewski (approved):
:+1:

PR #107, opened 2024-08-01
 Bump version numbers to v0.3.0 

### LINKED ISSUE #101: Ready for v0.3.0?

Kijewski:
* rinja 0.2.1 is backward compatible to v0.2.0, only adding new functions traits, not changing existing ones
* rinja_derive 0.3.0 emits code that uses e.g. `rinja::helpers::get_primitive_value()` which were not present in rinja 0.2.0, so it is not backward compatible
* rinja_parser introduces new enum members, so it is not backward compatible

Resolves #101.

### DIFF
--- a/rinja/Cargo.toml
+++ b/rinja/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rinja"
-version = "0.2.0"
+version = "0.3.0"
 description = "Type-safe, compiled Jinja-like templates for Rust"
 documentation = "https://docs.rs/rinja"
 keywords = ["markup", "template", "jinja2", "html"]
@@ -32,7 +32,7 @@ with-rocket = ["rinja_derive/with-rocket"]
 with-warp = ["rinja_derive/with-warp"]
 
 [dependencies]
-rinja_derive = { version = "0.2.0", path = "../rinja_derive" }
+rinja_derive = { version = "=0.3.0", path = "../rinja_derive" }
 
 itoa = "1.0.11"
 
--- a/rinja_derive/Cargo.toml
+++ b/rinja_derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rinja_derive"
-version = "0.2.0"
+version = "0.3.0"
 description = "Procedural macro package for Rinja"
 homepage = "https://github.com/rinja-rs/rinja"
 repository = "https://github.com/rinja-rs/rinja"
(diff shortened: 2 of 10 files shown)


GuillaumeGomez:
I'd prefer if we kept the same version number across all crates. Makes it simpler for users to know what's going on instead of trying to remember which version of X maps to Y.

Kijewski:
Then we must either

* bump the version of rinja to 0.3.0, too, even though it is not a breaking change, or
* pin the rinja_derive version in rinja, and the rinja_parser version in rinja_derive, but accept the people who on update on of the dependencies are in for trouble, or
* do both, so for later releases we can update the pinned version, making later releases easier.

Then I'd prefer the 3rd solution:

rinja/Cargo.toml: version = "0.3.0", rinra_derive = "=0.3.0"  
rinja_derive/Cargo.toml: version = "0.3.0", rinra_parser = "=0.3.0"  
rinja_parser/Cargo.toml: version = "0.3.0"

rinja/Cargo.toml
```diff
@@ -32,7 +32,7 @@ with-rocket = ["rinja_derive/with-rocket"]
 with-warp = ["rinja_derive/with-warp"]
 
 [dependencies]
-rinja_derive = { version = "0.2.0", path = "../rinja_derive" }
+rinja_derive = { version = "=0.3.0", path = "../rinja_derive" }
```
GuillaumeGomez, line 35: Why the equal sign?

rinja_derive/Cargo.toml
```diff
@@ -25,7 +25,8 @@ with-rocket = []
 with-warp = []
 
 [dependencies]
-parser = { package = "rinja_parser", version = "0.2.0", path = "../rinja_parser" }
+parser = { package = "rinja_parser", version = "=0.3.0", path = "../rinja_parser" }
```
GuillaumeGomez, line 28: Same here and below.

rinja/Cargo.toml
Kijewski: The problem is, should we make changes to the parser (say, adding `c"…"` string support), which require changes to the exported enums, then we need to do a semver incompatible release. If we want to keep the version numbers of all three crates the same, then we would need to bumb rinja's version number, too, making things difficult for downstream users.

With the equal signs we can control that all three versions stay related, and that a user cannot do a partial update.

GuillaumeGomez: Fine by me!

GuillaumeGomez (approved):

PR #108, opened 2024-08-01
derive: make sure that static CACHE is never dropped 

Kijewski:
We assume that `static CACHE` of `Config`s is never dropped, so we only pass `&'static Config` around. It is an implementation detail in rust, that `static` variables (currently) aren't dropped at the program's termination. This PR makes sure that this behavior stays the same even if this fact is changed later on.

Also, needless usages of `Arc` and some some `unsafe` code was removed.


GuillaumeGomez (approved):

GuillaumeGomez:
Good idea, thanks!

PR #111, opened 2024-08-05
derive: use fewer unwraps

Kijewski:
`rinja_derive(_standalone)` expects that the environment variable `CARGO_MANIFEST_DIR` is set and panics otherwise. It also expects that the input can be parsed as a struct/enum/union and panics otherwise.

This PR uses `"."` as fallback path if `CARGO_MANIFEST_DIR` is unset, and generates a compiler error if the input TokenStream could not be parsed.

Both error conditions should only occur in `rinja_derive_standalone`, esp. if you have an unexpected runtime like WASM that does not have a filesystem or environment variables.

### DIFF
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -105,7 +105,7 @@ impl Config {
         let config_path = key.0.config_path.as_deref();
         let template_whitespace = key.0.template_whitespace.as_deref();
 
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Less `unwrap` calls makes me happy. :)

Kijewski:
Hehe, me too. :)

PR #112, opened 2024-08-05
FastWritable data used in filters should be FastWritable

Kijewski:
E.g. if you have a `size: usize` that you print with `{{size|safe}}`, then the fast writer without `fmt::write!()` should be used.


Kijewski:
The diff is much bigger than I had anticipated, but the biggest culprit is a moved file. But the performance gain is also much bigger than anticipated!

I guess `small gain` × `a big loop` = `big gain`. :)

GuillaumeGomez (approved):

GuillaumeGomez:
It's impressive for sure, thanks!

Kijewski:
Yeah, I expected maybe 3%, again. The test looks sane, like something that could actually be used in a real context, so I'm quite happy with the PR. :)

PR #116, opened 2024-08-07
derive: understand extension `.rinja` + don't panic if the extension is non-UTF-8

Kijewski:


GuillaumeGomez (approved):

GuillaumeGomez:
Makes sense, thanks!

PR #117, opened 2024-08-07
derive: allow split up template attributes

Kijewski:
In generated code or macros, it might be useful to emit multiple `#[template]` attributes. E.g.

```rust
 #[template(source = "Hello!")]
 #[template(ext = "txt")]
 struct Hello;
```

### DIFF
--- a/rinja_derive/src/input.rs
+++ b/rinja_derive/src/input.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::collections::hash_map::{Entry, HashMap};
 use std::fs::read_to_string;
+use std::iter::FusedIterator;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, OnceLock};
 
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Nice add! Can you add a test like this please:

```rust
#[derive(Template)]
#[template(source = "bla")]
#[template(source = "bla")]
struct Foo;
```

Just to confirm the parsing of attributes is taking duplicates into account when they are not in the same attribute.

Kijewski:
> Just to confirm the parsing of attributes is taking duplicates into account when they are not in the same attribute.

That's done in <https://github.com/rinja-rs/rinja/pull/117/files#diff-d2f7c348bb75850b1706f0efcd72e6b6c82baff6803d2e0aa5ca0521fe159ad2>. Before it failed because there were two `#[template]`s, now in fails because the `source` is given twice.

GuillaumeGomez (approved):

GuillaumeGomez:
Oh wow, how did I miss that... Sorry for missing it. Thanks!

Kijewski:
> Oh wow, how did I miss that...

Hehe, don't worry. I could have made the it more explicit bei adding a new test instead of changing the meaning of an existing one.

PR #118, opened 2024-08-07
parser: tell user proper keyword to end node

Kijewski:


rinja_parser/src/node.rs
```diff
@@ -1057,6 +1057,25 @@ impl<'a> Comment<'a> {
...
+    node: &'g str,
+    expected: &'g str,
+) -> impl Fn(&'a str) -> ParseResult<'a> + 'g {
+    move |start| {
+        let (i, actual) = ws(identifier)(start)?;
+        if actual == expected {
+            Ok((i, actual))
+        } else if actual == "end" {
+            Err(nom::Err::Failure(ErrorContext::new(
+                format!("`{node}` nodes are terminated with `{expected}`"),
```
GuillaumeGomez, line 1070: I would reword it a bit to make it more obvious that the "ender" is missing:

```suggestion
                format!("missing `{expected}` to terminate `{node}` node"),
```

Kijewski: I thought my wording expresses better that `end` is not a keyword. Maybe the combined message would be best?

```rust
"\
missing `{expected}` to terminate `{node}` node
`{node}` nodes are terminated with `{expected}`, not `end`"
```

GuillaumeGomez: I'm still not entirely sold on your message improvement. Also, thinking about this some more, I think the `else if actual == "end"` block should be merged with the `else`.

Message suggestion v3:

```
expected `{expected}` to terminate `{node}` node, found `{actual}`
```

Kijewski: > I think the `else if actual == "end"` block should be merged with the else.

Not possible. `fail()` returns an recoverable error, but for `actual == "end"` we return an unrecoverable failure. (The name `fail()` might not have been the best choice.)

GuillaumeGomez: Oh I see. We tell that it looks to be looking further because it doesn't match the need here. Hum... Did you encounter this problem yourself? (using `end` instead of the right keyword)

Kijewski: I searched what usage questions users were having about askama, and came across this thread: <https://users.rust-lang.org/t/how-to-solve-error-about-askama/59431>

Kijewski:
I added a commit that shows the error messages before this PR: <https://github.com/rinja-rs/rinja/pull/118/commits/0628afaf5efd861dcb4967eae3e9f1b3e5e7735d>. I think it's much better now. :)

GuillaumeGomez (approved):

GuillaumeGomez:
I like it, thanks for going along! Very good idea to look at issues askama users encountered. :)

Kijewski:
> Very good idea to look at issues askama users encountered. :)

Thanks! :) Yes, I can always try to imagine what problems other people might have, but my knowledge of the project is much different from other users.

PR #119, opened 2024-08-07
parser: allow `match` nodes w/o `when` if there is an `else`

Kijewski:


GuillaumeGomez:
Nice idea! Can you update the book to mention `else` for `match` please?

Kijewski:
> Nice idea!

First I only wanted to give a better message if there are no `when` blocks, but then I couldn't think of a good reason why not to allow it if there is an `else`.

> Can you update the book to mention `else` for `match` please?

There is a sentence "The `else` block is equivalent to matching on `_` (matching anything)", but I will try if I can word it better.

book/src/template_syntax.md
```diff
@@ -559,16 +559,49 @@ expand an `Option`:
...
+{% endmatch %}
+```
+
+Or some more complex type, such as a [`Result<T, E>`](https://doc.rust-lang.org/stable/std/result/enum.Result.html):
+
+```jinja
+{% match result %}
+  {% when Ok(val) %} Good: {{ val }}.
+  {% when Err(err) %} Bad: {{ err }}.
+```
```
GuillaumeGomez, line 582:
``````suggestion
{% endmatch %}
```
``````

File: book/src/template_syntax.md [hunk 2]
```diff
@@ -559,16 +559,49 @@ expand an `Option`:
...
+
+Using the placeholder `_` to match against any value without capturing the datum, works too.
+The wildcard operator `..` is used to match against an arbitrary amount of items,
+and the same restrictions as in Rust, e.g. that it can be used only once in a slice or struct:
+
+```jinja
+{% match list_of_ints %}
+  {% when [first, ..] %} The list starts with a {{ first }}
+  {% else %} The list is empty.
```
GuillaumeGomez, line 591: I would use `{% when _ %}` here and then rewrite this example after the `{% else %}` explanation below.

Kijewski: Changed.

Kijewski:
Huh, the alternatives syntax `{% when 1 | 2 | 3 %}` does not actually work, but I am sure I saw it in the code.

GuillaumeGomez (approved):

book/src/template_syntax.md
```diff
@@ -559,16 +559,55 @@ expand an `Option`:
...
+A `{% match %}` must be exhaustive, i.e. all possible inputs must have a case.
+This is most easily done by using proving an `{% else %}` case,
+if not all possible values need an individual handling.
+
+```jinja
+{% match answer %}
+  {% when Ok(42) %} The answer is "42".
+  {% else %} No answer wrong answer?
+{% endmatch %}
+```
```
GuillaumeGomez, line 607:
``````suggestion
The `{% else %}` node is syntactical sugar for `{% when _ %}`.
If used, it must come last, after all other `{% when %}` blocks:

```jinja
{% match answer %}
  {% when Ok(42) %} The answer is "42".
  {% else %} No answer wrong answer?
{% endmatch %}
```

A `{% match %}` must be exhaustive, i.e. all possible inputs must have a case.
This is most easily done by using proving an `{% else %}` case,
if not all possible values need an individual handling.
``````

GuillaumeGomez:
Once my last comment is done, please merge. :)

GuillaumeGomez:
I opened https://github.com/rinja-rs/rinja/issues/130 for the `|` in `when`.

Kijewski:
I hoped if I use the "apply suggestion" button it might not dismiss your review, but alas, it does. :)

GuillaumeGomez:
Hehehe. I can click on the "approve" button again. :3

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the improvement!

Issue #115, opened 2024-08-05
Improve `is (not) defined` feature


GuillaumeGomez:
In some cases, we could remove even more code, like:

```jinja
{% if x is not defined || !x %}
```

Currently, we keep the `!x` but there's no reason we should if it's not defined since the first condition will be `true`. It would allow to be more flexible.

PR #126, opened 2024-08-09
Improve compile-time if conditions handling

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/115.

It finally allows to write:

```jinja
{% if x is defined && x == 12 %}
```

without having compile-time errors because of the `x == 12` part. :)

### DIFF
--- a/rinja_parser/src/memchr_splitter.rs
+++ b/rinja_parser/src/memchr_splitter.rs
@@ -7,7 +7,7 @@ pub(crate) trait Splitter: Copy {
     fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)>;
 }
 
-impl<T: Splitter + ?Sized> Splitter for &T {
+impl<T: Splitter> Splitter for &T {
     #[inline]
     fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)> {
         T::split(self, haystack)
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -392,10 +392,12 @@ impl<'a> Generator<'a> {
 
     fn evaluate_condition(
         &self,
-        expr: &WithSpan<'_, Expr<'_>>,
+        expr: WithSpan<'a, Expr<'a>>,
         only_contains_is_defined: &mut bool,
-    ) -> EvaluatedResult {
-        match **expr {
+    ) -> (EvaluatedResult, WithSpan<'a, Expr<'a>>) {
+        let (expr, span) = expr.deconstruct();
+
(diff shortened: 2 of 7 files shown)


GuillaumeGomez:
Fixed fmt. :)

Kijewski:
Will review in the next few days, today I'm wiped. :)

rinja_parser/src/memchr_splitter.rs
```diff
@@ -7,7 +7,7 @@ pub(crate) trait Splitter: Copy {
     fn split<'a>(&self, haystack: &'a str) -> Option<(&'a str, &'a str)>;
 }
 
-impl<T: Splitter + ?Sized> Splitter for &T {
+impl<T: Splitter> Splitter for &T {
```
Kijewski, line 10: Interesting, why didn't clippy like that? Because Splitter is not `pub`, so it's known that there won't be any unsized Splitters?

GuillaumeGomez: Here's the output:

```
warning: `?Sized` bound is ignored because of a `Sized` requirement
  --> rinja_parser/src/memchr_splitter.rs:10:20
   |
10 | impl<T: Splitter + ?Sized> Splitter for &T {
   |                    ^^^^^^
   |
note: `T` cannot be unsized because of the bound
  --> rinja_parser/src/memchr_splitter.rs:10:9
   |
10 | impl<T: Splitter + ?Sized> Splitter for &T {
   |         ^^^^^^^^
   = note: ...because `Splitter` has the bound `Copy`
   = note: ...because `Copy` has the bound `Clone`
   = note: ...because `Clone` has the bound `Sized`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_maybe_sized
   = note: `#[warn(clippy::needless_maybe_sized)]` on by default
help: change the bounds that require `Sized`, or remove the `?Sized` bound
   |
10 - impl<T: Splitter + ?Sized> Splitter for &T {
10 + impl<T: Splitter> Splitter for &T {
   |
```

Kijewski (approved):
I'm not super happy that the `Expr` need to get cloned, but I don't see a better implementation. Adding e.g. an `enum { &Expr | Box<Expr> }` type would make the code unreadable, because the change would be viral: any `&Expr` would need to be changed down the line.

And since the conditions should not be excessively big (ensured by our nesting level check), this should be fine.

Thanks for the PR! :)

GuillaumeGomez:
I'm not super happy about it either, that's why I added the `contains_bool_lit_or_is_defined` field to only clone when it's actually needed to limit performance impact.

PR #127, opened 2024-08-12
derive: allow supplying source in comment

Kijewski:
Using `#[template(source = "…")]` is tiresome, because you have to escape the input. That makes it difficult to read, too. Using `#[template(path = "…")` can be tiresome, too, if the code is small and/or if you just want to prototype something.

This PR adds a third option to supply the source of a template: You can supply the source in the comments of a template. This works similar to doctest. Any <code>```rinja```</code> block gets extracted, and the combined blocks are the source code of the template.

### DIFF
--- a/book/src/creating_templates.md
+++ b/book/src/creating_templates.md
@@ -114,7 +114,10 @@ recognized:
 As an alternative to supplying the code template code in an external file (as `path` argument),
 or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
 With this feature, you can specify the template code directly in the documentation
-of the template `struct`. Simply add a ```` ```rinja ```` code block to your documentation:
+of the template `struct`.
+
+Instead of `path = "…"` or `source = "…"`, specify `in_doc = true` in the `#[template]` attribute,
+and in the struct's documentation add a ```` ```rinja ```` code block:
 
 ```rust
 /// Here you can put our usual comments.
@@ -134,7 +137,7 @@ of the template `struct`. Simply add a ```` ```rinja ```` code block to your doc
 ///
 /// All comments are still optional, though.
 #[derive(Template)]
-#[template(ext = "html")]
+#[template(ext = "html", in_doc = true)]
 struct Example<'a> {
     lines: &'a str,
 }
@@ -142,8 +145,6 @@ struct Example<'a> {
 
 If you want to supply the template code in the comments,
 then you have to specify the `ext` argument, too, e.g. `#[template(ext = "html")]`.
-If the `path` or `source` argument is given,
-then this argument takes precedence over the code in the comments.
 
-Instead of ```` ```rinja ````, you can also write ```` ```jinja ```` or ```` ```jinja2 ````,
-e.g. to get it to work better in conjunction with sytax highlighters.
+Instead of `rinja`, you can also write `jinja` or `jinja2`,
+e.g. to get it to work better in conjunction with syntax highlighters.
--- a/book/src/creating_templates.md
+++ b/book/src/creating_templates.md
@@ -124,10 +124,10 @@ recognized:
 As an alternative to supplying the code template code in an external file (as `path` argument),
 or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
 With this feature, you can specify the template code directly in the documentation
-of the template `struct`.
+of the template item.
 
 Instead of `path = "…"` or `source = "…"`, specify `in_doc = true` in the `#[template]` attribute,
-and in the struct's documentation add a ```` ```rinja ```` code block:
+and in the item's documentation, add a code block with the `rinja` attribute:
 
 ```rust
 /// Here you can put our usual comments.
--- a/book/src/creating_templates.md
+++ b/book/src/creating_templates.md
@@ -55,6 +55,10 @@ recognized:
       name: &'a str,
   }
   ```
+
+* `in_doc` (as `in_doc = true`):
+  please see the section ["documentation as template code"](#documentation-as-template-code).
+
 * `ext` (as `ext = "txt"`): lets you specify the content type as a file
   extension. This is used to infer an escape mode (see below), and some
   web framework integrations use it to determine the content type.
@@ -66,6 +70,7 @@ recognized:
       name: &'a str,
   }
   ```
+
 * `print` (as `print = "code"`): enable debugging by printing nothing
   (`none`), the parsed syntax tree (`ast`), the generated code (`code`)
   or `all` for both. The requested data will be printed to stdout at
@@ -75,6 +80,7 @@ recognized:
   #[template(path = "hello.html", print = "all")]
   struct HelloTemplate<'a> { ... }
   ```
+
 * `block` (as `block = "block_name"`): renders the block by itself.
   Expressions outside of the block are not required by the struct, and
   inheritance is also supported. This can be useful when you need to
@@ -85,6 +91,7 @@ recognized:
   #[template(path = "hello.html", block = "hello")]
   struct HelloTemplate<'a> { ... }
   ```
+
 * `escape` (as `escape = "none"`): override the template's extension used for
   the purpose of determining the escaper for this template. See the section
   on configuring custom escapers for more information.
@@ -93,6 +100,7 @@ recognized:
   #[template(path = "hello.html", escape = "none")]
   struct HelloTemplate<'a> { ... }
   ```
+
 * `syntax` (as `syntax = "foo"`): set the syntax name for a parser defined
   in the configuration file. The default syntax , "default", is the one
   provided by Rinja.
@@ -101,6 +109,7 @@ recognized:
   #[template(path = "hello.html", syntax = "foo")]
   struct HelloTemplate<'a> { ... }
   ```
+
 * `config` (as `config = "config_file_path"`): set the path for the config file
   to be used. The path is interpreted as relative to your crate root.
   ```rust
@@ -110,6 +119,7 @@ recognized:
   ```
 
 ## Documentation as template code
+[#documentation-as-template-code]: #documentation-as-template-code
 
 As an alternative to supplying the code template code in an external file (as `path` argument),
 or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
--- a/book/src/creating_templates.md
+++ b/book/src/creating_templates.md
@@ -108,3 +108,42 @@ recognized:
   #[template(path = "hello.html", config = "config.toml")]
   struct HelloTemplate<'a> { ... }
   ```
+
+## Documentation as template code
+
+As an alternative to supplying the code template code in an external file (as `path` argument),
+or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
+With this feature, you can specify the template code directly in the documentation
+of the template `struct`. Simply add a ```` ```rinja ```` code block to your documentation:
+
+```rust
+/// Here you can put our usual comments.
+///
+/// ```rinja
+/// <div>{{ lines|linebreaksbr }}</div>
+/// ```
+///
+/// Any usual docs, including tests can be put in here, too:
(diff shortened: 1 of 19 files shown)


GuillaumeGomez:
I like the idea! Don't forget to test with `/**` and `#[doc = ""]` too. :wink:

Kijewski:
> I like the idea! Don't forget to test with `/**` and `#[doc = ""]` too. 😉

Oof, yeah, my code did not work at all with these two. :) Fixed.

GuillaumeGomez:
Does it work with code examples like:

``````rust
///     let x = 12;
///
/// `````
/// let x = 12;
/// `````
``````

?

If not, I'd recommend using `pulldown-cmark` to parse the markdown and extract code blocks from it instead of trying to do it yourself. ;)

**Comment by GuillaumeGomez** [comment]
It's just one of many. Not sure if your code would handle:

```rust
/// `````
/// ```rinja
/// {{bla}}
/// ```
/// `````
```

correctly

Kijewski:
> It's just one of many. Not sure if your code would handle:

It does. I had to turn one panic into an error so I could add the test.

Kijewski:
I use pulldown-cmark, now. It has a different opinion how to interpret ↓, but I guess it knows it better. :)

    /// ```rinja
    /// `````
    /// {{bla}}
    /// `````
    /// ```

Kijewski:
Should we wait until <https://github.com/pulldown-cmark/pulldown-cmark/pull/916> is released to keep our MSRV?

GuillaumeGomez:
Oh that'd be convenient. Another thing I thought about: should we make it an optional feature?

Kijewski:
> Another thing I thought about: should we make it an optional feature?

Yes, good call! It will probably a very niche feature. What would you call the feature flag? Is "code-in-doc" too long?

GuillaumeGomez:
`code-in-doc` sounds good to me. :+1:

GuillaumeGomez:
Oh btw I asked on the `pulldown-cmark` PR and they said hopefully there should be a release this week. :)

GuillaumeGomez:
New `pulldown-cmark` version with reduced MSRV has been published.

Kijewski:
> error: package `pulldown-cmark v0.11.1` cannot be built because it requires rustc 1.71.1 or newer, while the currently active rustc version is 1.71.0

I guess it's acceptable if I raise our MSRV by 0.1. :D

rinja_derive/src/input.rs
```diff
@@ -422,6 +427,127 @@ impl TemplateArgs {
...
+fn collect_rinja_code_blocks(source: String) -> Option<Source> {
+    use pulldown_cmark::{CodeBlockKind, Event, Parser, Tag, TagEnd};
+
+    let mut tmpl_source = String::new();
+    let mut in_rinja_code = false;
+    let mut had_rinja_code = false;
+    for e in Parser::new(&source) {
+        match (in_rinja_code, e) {
+            (false, Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(s)))) => {
+                if s.split(",").any(|s| JINJA_EXTENSIONS.contains(&s)) {
```
GuillaumeGomez, line 531: In this case I think it should be:

```suggestion
                if s.split(",").any(|s| JINJA_EXTENSIONS.contains(&s)) {
```

The reason is because code block attributes are actually a bit complicated: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#custom-css-classes-for-code-blocks

GuillaumeGomez: ... Nevermind, it's exactly what you're doing. End of week urg.

GuillaumeGomez (approved):

GuillaumeGomez:
You also need to update the book. :)

Kijewski:
> You also need to update the book. :)

Added. :)

book/src/creating_templates.md
```diff
@@ -108,3 +108,42 @@ recognized:
...
+    lines: &'a str,
+}
+```
+
+If you want to supply the template code in the comments,
+then you have to specify the `ext` argument, too, e.g. `#[template(ext = "html")]`.
+If the `path` or `source` argument is given,
+then this argument takes precedence over the code in the comments.
+
+Instead of ```` ```rinja ````, you can also write ```` ```jinja ```` or ```` ```jinja2 ````,
```
GuillaumeGomez, line 148:
```suggestion
Instead of `rinja`, you can also write `jinja` or `jinja2`,
```

book/src/creating_templates.md
```diff
@@ -108,3 +108,42 @@ recognized:
...
+/// ```rust
+/// assert_eq!(
+///     Example { lines: "a\nb\nc" }.to_string(),
+///     "<div>a<br/>b<br/>c</div>"
+/// );
+/// ```
+///
+/// All comments are still optional, though.
+#[derive(Template)]
+#[template(ext = "html")]
```
GuillaumeGomez, line 137: I thought about it and I think we should only parse documentation if we have an attribute like `#[template(in_doc = true)]` or something like that. I see two reasons why we should do:
 1. For users, it's not magical anymore when they read source code to know where the template comes from.
 2. For `rinja`, it means we'll only parse doc comments if and only if the attribute is set to `true` (so impact on performance reduced).
 
 What do you think?

Kijewski: Yes, makes sense! Added.

book/src/creating_templates.md
Kijewski: Done.

book/src/creating_templates.md
```diff
@@ -108,3 +108,42 @@ recognized:
   #[template(path = "hello.html", config = "config.toml")]
   struct HelloTemplate<'a> { ... }
   ```
+
+## Documentation as template code
+
+As an alternative to supplying the code template code in an external file (as `path` argument),
+or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
+With this feature, you can specify the template code directly in the documentation
+of the template `struct`. Simply add a ```` ```rinja ```` code block to your documentation:
```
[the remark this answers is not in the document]
Kijewski: Reworded the paragraph.

rinja_derive/src/input.rs
```diff
@@ -422,6 +439,161 @@ impl TemplateArgs {
...
+/// Try to find the souce in the comment, in a "```rinja```" block
+///
+/// This is only done if no path or source was given in the `#[template]` attribute.
+fn source_from_docs(
+    name: &syn::Ident,
+    value: &syn::ExprLit,
+    dest: &mut Option<(Source, Option<Span>)>,
+    ast: &syn::DeriveInput,
+) -> Result<(), CompileError> {
+    let syn::Lit::Bool(syn::LitBool { value: true, .. }) = &value.lit else {
```
GuillaumeGomez, line 452: In this case, please let's just check it's a boolean. And if `false`, we don't parse docs. It'd be a bit weird otherwise to expect only `true`. I suspect I will give it a try and sometimes just switch the boolean value to enable/disable the feature.

Kijewski:
Hm, should `#[template(in_doc = false)]` succeed if the feature is not implemented? I guess yes?

rinja_derive/src/input.rs
Kijewski: Makes sense, done!

GuillaumeGomez:
I'd say yes.

book/src/creating_templates.md
```diff
@@ -101,10 +109,52 @@ recognized:
...
   struct HelloTemplate<'a> { ... }
   ```
+
+## Documentation as template code
+[#documentation-as-template-code]: #documentation-as-template-code
+
+As an alternative to supplying the code template code in an external file (as `path` argument),
+or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
+With this feature, you can specify the template code directly in the documentation
+of the template `struct`.
```
GuillaumeGomez, line 127:
```suggestion
of the template item.
```

Because I think we support `union` or `enum` (or both). Never remember which.

rinja_actix/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rinja_actix"
-version = "0.3.0"
+version = "0.3.1"
```
GuillaumeGomez, line 3: A reason why you're updating versions? Shouldn't it be done in a separate PR?

rinja_derive/src/input.rs
```diff
@@ -55,7 +55,13 @@ impl TemplateInput<'_> {
         // related. In case `source` was used instead of `path`, the value
         // of `ext` is merged into a synthetic `path` value here.
         let &(ref source, source_span) = source.as_ref().ok_or_else(|| {
-            CompileError::new("template `path` or `source` not found in attributes", None)
+            CompileError::new(
+                #[cfg(not(feature = "code-in-doc"))]
+                "specify one template argument `path` OR `source`",
```
GuillaumeGomez, line 60: Maybe let's not put `OR` in capital? XD

rinja_derive/src/input.rs
```diff
@@ -422,6 +433,177 @@ impl TemplateArgs {
     }
 }
 
+/// Try to find the souce in the comment, in a "```rinja```" block
```
GuillaumeGomez, line 436:
```suggestion
/// Try to find the souce in the comment, in a `rinja` code block.
```

rinja_actix/Cargo.toml
Kijewski: Because I added a new feature, but it could be done in another PR, too.

rinja_derive/src/input.rs
```diff
@@ -422,6 +433,177 @@ impl TemplateArgs {
...
+
+    Ok((span, source))
+}
+
+#[cfg(feature = "code-in-doc")]
+fn no_rinja_code_block(name: &syn::Ident, ast: &syn::DeriveInput) -> CompileError {
+    let kind = match &ast.data {
+        syn::Data::Struct(_) => "struct",
+        syn::Data::Enum(_) => "enum",
+        syn::Data::Union(_) => "union",
```
GuillaumeGomez, line 535: So we do support all item types! :laughing:

rinja_derive/src/lib.rs
```diff
@@ -64,6 +64,29 @@ use syn::parse_quote_spanned;
...
+///
+/// E.g. `in_doc = true`
+///
+/// As an alternative to supplying the code template code in an external file (as `path` argument),
+/// or as a string (as `source` argument), you can also enable the `"code-in-doc"` feature.
+/// With this feature, you can specify the template code directly in the documentation
+/// of the template `struct`.
+///
+/// Instead of `path = "…"` or `source = "…"`, specify `in_doc = true` in the `#[template]`
+/// attribute, and in the struct's documentation add a ```` ```rinja ```` code block:
```
GuillaumeGomez, line 77:
```suggestion
/// attribute, and in the struct's documentation add a `rinja` code block:
```

rinja_actix/Cargo.toml
GuillaumeGomez: It's not the first one. Please do it in another PR then. Like that we make publish it too and also ensure we don't have any other features we want to be merged before making this release.

rinja_derive/src/input.rs
Kijewski: Hehe, I simply added cases for all variant of the enum. I actually don't think that we generate valid code for unions, because I think reading from unions is only allowed in an unsafe context.

GuillaumeGomez: Funny. Might worth handling unions correctly at some point.

Kijewski:
All suggestions applied.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks a lot, it's a very cool feature! :)

Kijewski:
Thanks! And thank you for your reviews and suggestions! :)

Issue #128, opened 2024-08-13
 Template parsing issues when using double angle brackets as delimiters


ronnodas:
With `expr` delimiters `<<` and `>>`, trying to use the template `<<a>> and <<b>>` complains about `and` not being a field. Here is a full example that produces the error "no field `and` on type `&HelloTemplate`":
```toml
[[syntax]]
name = "mwe"
expr_start = "<<"
expr_end = ">>"
```

```rust
#[derive(Template)]
#[template(source = "<<a>> and <<b>>", syntax = "mwe", ext="")]
struct HelloTemplate {
    a: u32,
    b: u32,
}
```
Maybe worse, using `(<<a>> and) <<b>>`  as the template produces "failed to parse template source".
```rust
#[derive(Template)]
#[template(source = "(<<a>> and) <<b>>", syntax = "mwe", ext = "")]
struct HelloTemplate {
    a: u32,
    b: u32,
}
```
Everything works fine if the delimiters are changed to eg `<!` and `!>`. I have not experimented with other pairs.

The same issue also exists in askama, which I have filed as https://github.com/djc/askama/issues/1081.

Kijewski:
Thank you for the bug report! As the language is currently parsed, I don't think this is fixable. The parser sees a right shift `>>`, so it wants to parse a binary operation. The string `<<a>> and) <<b>>` is tokenized like: `StartExpr`, `Var(a)`, `Op(">>")`, `Var("and")`, `Unexpected(")")`, …

We should explicitly reject terminators that start with `>>`, `<<`, `..`, `&&`, or `||`, so you and other users get a proper error message.

ronnodas:
That makes sense. It's weird that it works in the python/jinja project I'm porting from.

Kijewski:
They probably first look for the terminator, then try to make sense of everything captured between the delimiters. There are pros and cons to both approaches.

GuillaumeGomez:
It was fixed in #129 so I think it can be closed now. It'll be part of next release.

PR #129, opened 2024-08-13
parser: reject closing delimiters that look like an operator

Kijewski:
Related issue: #128.

### DIFF
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -321,38 +321,28 @@ impl<'a> RawSyntax<'a> {
             comment_end: self.comment_end.unwrap_or(default.comment_end),
         };
 
-        for s in [
-            syntax.block_start,
-            syntax.block_end,
-            syntax.expr_start,
-            syntax.expr_end,
-            syntax.comment_start,
-            syntax.comment_end,
+        for (s, k) in [
+            (syntax.block_start, "opening block"),
+            (syntax.block_end, "closing block"),
+            (syntax.expr_start, "opening expression"),
+            (syntax.expr_end, "closing expression"),
+            (syntax.comment_start, "opening comment"),
+            (syntax.comment_end, "closing comment"),
         ] {
             if s.len() < 2 {
                 return Err(CompileError::new_with_span(
-                    format!("delimiters must be at least two characters long: {s:?}"),
+                    format!(
+                        "delimiters must be at least two characters long. \
+                         The {k} delimiter ({s:?}) is too short",
+                    ),
                     file_info.copied(),
                     config_span,
                 ));
             } else if s.chars().any(|c| c.is_whitespace()) {
                 return Err(CompileError::new_with_span(
-                    format!("delimiters may not contain white spaces: {s:?}"),
-                    file_info.copied(),
-                    config_span,
-                ));
-            }
-        }
-
-        for (s1, s2) in [
-            (syntax.block_start, syntax.expr_start),
-            (syntax.block_start, syntax.comment_start),
-            (syntax.expr_start, syntax.comment_start),
-        ] {
-            if s1.starts_with(s2) || s2.starts_with(s1) {
-                return Err(CompileError::new_with_span(
                     format!(
-                        "a delimiter may not be the prefix of another delimiter: {s1:?} vs {s2:?}",
+                        "delimiters may not contain white spaces. \
+                         The {k} delimiter ({s:?}) contains white spaces",
                     ),
                     file_info.copied(),
                     config_span,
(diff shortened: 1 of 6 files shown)


Kijewski:
@ronnodas, do you think that the error message in this PR is good enough?

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
             }
         }
 
+        for end in [syntax.block_start, syntax.expr_start, syntax.comment_start] {
```
GuillaumeGomez, line 363: Unless I'm missing something, you never check for closing delimiters here, right?

ronnodas:
Yes, the error message looks good but it seems the code is comparing the starting delimiters even though it maybe should be comparing the ending ones.

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
...
+            for prefix in ["<<", ">>", "&&", "..", "||"] {
+                if end.starts_with(prefix) {
+                    let msg = if end == prefix {
+                        format!("a closing delimiter must not start with an operator: {end:?}")
```
GuillaumeGomez, line 367: It'd be nice to make the distinction between closing and opening delimiter (in the current code, it's only opening delimiters).

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
...
+                    } else {
+                        format!(
+                            "a closing delimiter must not start an with operator: \
+                             {end:?} vs {prefix:?}",
```
GuillaumeGomez, line 371: The "vs" part is a bit weird. I'd reword it like: `a closing/opening delimiter must not start with an operator: {end:?} starts with {prefix:?}`.

GuillaumeGomez:
> Yes, the error message looks good but it seems the code is comparing the starting delimiters even though it maybe should be comparing the ending ones.

I mentioned the same thing 😄

Kijewski:
Hehe, yeah, I should look what I copy and paste. Now the terminators are checked and the message is reworded.

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
             }
         }
 
+        for end in [syntax.block_end, syntax.expr_end, syntax.comment_end] {
```
GuillaumeGomez, line 363: I think it should be checked for both starters and enders, not just enders (or starters in the previous version).

Kijewski: It's not needed, because we do not try to understand text blocks. The parsing will work with any starting delimiter. I would not like to introduce restrictions needlessly.

GuillaumeGomez: Fair enough. :+1:

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
...
+            for prefix in ["<<", ">>", "&&", "..", "||"] {
+                if end.starts_with(prefix) {
+                    let msg = if end == prefix {
+                        format!("a closing delimiter must not start with an operator: {end:?}")
```
GuillaumeGomez, line 367: nitpicking: it would be nice to know which block is impacted by this. Ie "expr_end" would be "expression block". So in this case the message would look like:

```
a closing delimiter must not start with an operator. Expression block closing delimiter is {end:?} which is also an operator
```

rinja_derive/src/config.rs
```diff
@@ -360,6 +360,26 @@ impl<'a> RawSyntax<'a> {
...
+                    } else {
+                        format!(
+                            "a closing delimiter must not start an with operator: \
+                             {end:?} starts with {prefix:?}",
```
GuillaumeGomez, line 371: Same here with a little tweak:

```
a closing delimiter must not start with an operator. Expression block closing delimiter ({end:?}) starts with {prefix:?}
```

Kijewski: Makes sense! I changed the text in the other delimiter tests, too.

rinja_derive/src/config.rs
```diff
@@ -321,45 +321,92 @@ impl<'a> RawSyntax<'a> {
...
 
-        for s in [
-            syntax.block_start,
-            syntax.block_end,
-            syntax.expr_start,
-            syntax.expr_end,
-            syntax.comment_start,
-            syntax.comment_end,
+        for (s, k) in [
+            (syntax.block_start, "opening block"),
```
GuillaumeGomez, line 325: Good idea implementing it here too!

rinja_derive/src/config.rs
```diff
@@ -321,45 +321,92 @@ impl<'a> RawSyntax<'a> {
...
+                if end.starts_with(prefix) {
+                    let msg = if end == prefix {
+                        format!(
+                            "a closing delimiter must not start with an operator. \
+                             The {kind} delimiter ({end:?}) is also an operator",
+                        )
+                    } else {
+                        format!(
+                            "a closing delimiter must not start an with operator. \
+                             The {kind} delimiter ({end:?}) starts with {prefix:?}",
```
GuillaumeGomez, line 398:
```suggestion
                             The {kind} delimiter ({end:?}) starts with the {prefix:?} operator",
```

rinja_derive/src/config.rs
```diff
@@ -321,45 +321,92 @@ impl<'a> RawSyntax<'a> {
...
+            (syntax.expr_start, "opening expression"),
+            (syntax.expr_end, "closing expression"),
+            (syntax.comment_start, "opening comment"),
+            (syntax.comment_end, "closing comment"),
         ] {
             if s.len() < 2 {
                 return Err(CompileError::new_with_span(
-                    format!("delimiters must be at least two characters long: {s:?}"),
+                    format!(
+                        "delimiters must be at least two characters long. \
```
GuillaumeGomez, line 335: I realize we don't seem to have ui tests for this. I'll send a PR that.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Kijewski:
Thank you for the reviews and the good idea to call out the offending delimiter.

@ronnodas, thank you for reporting the bug!

rinja_derive/src/config.rs
GuillaumeGomez: Nevermind, tests are at the bottom.

PR #132, opened 2024-08-13
derive: no panics, less work, more compatibility

Kijewski:
Some minor stuff I found while implementing #127; extracted from there.

* allow any amount of invisible delimiters: more compatibility with generated TokenStreams
* don't panic if source+path are missing: print an error message instead
* don't write empty string: because even printing nothing causes work


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #135, opened 2024-08-15
Add support for `|` operator in `when`

### LINKED ISSUE #130: Add support for `|` in `match`/`when`
Currently this code doesn't work because of `|` combinator:

```jinja
{% match item %}
  {% when 1 | 2 %}yep
  {% when _ %}yop
{% endmatch %}
```

GuillaumeGomez:
Fixes #130.

The second commit is simply because I realized that we had a few one character strings around we were sending to `BufferFmt` instead of just pushing a `char`. So I implemented the trait for `char` as well and replaced the few cases.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -673,7 +673,12 @@ impl<'a> Generator<'a> {
             }
 
             self.locals.push();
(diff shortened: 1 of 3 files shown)


Kijewski (approved):
Nice and concise, thank you! :)

PR #141, opened 2024-08-18
Move Syntax configuration and check into parser

### LINKED ISSUE #138: parser: `Syntax` is trusted to have valid data
The fields of `rinja_parser::Syntax` are public and `Ast::from_str()` trusts that the input is valid, e.g. no field contains an empty string.

If a field does contain the an empty string, then in the best case the parser panics. In the worst case it could probably deadlock, in a loop advancing the input by null bytes until you kill the process.

Should the check if the syntax is sane be moved from `rinja_derive::RawSyntax` into `rinja_parser::Syntax`?

Kijewski:
Resolves #138.

### DIFF
--- a/rinja_derive/Cargo.toml
+++ b/rinja_derive/Cargo.toml
@@ -14,7 +14,7 @@ rust-version = "1.71"
 proc-macro = true
 
 [features]
-config = ["dep:serde", "dep:basic-toml"]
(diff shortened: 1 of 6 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #142, opened 2024-08-19
derive: escape `int`, `float`, `bool` at compile time

Kijewski:


rinja_derive/src/generator.rs
```diff
@@ -2347,6 +2210,126 @@ impl<'a> Generator<'a> {
...
+                NumKind::Int(Some(IntKind::U128)) => int(u128::from_str_radix, &value)?,
+                NumKind::Int(None) => match value.starts_with('-') {
+                    true => int(i128::from_str_radix, &value)?,
+                    false => int(u128::from_str_radix, &value)?,
+                },
+                NumKind::Float(Some(FloatKind::F32)) => value.parse::<f32>().ok()?.to_string(),
+                NumKind::Float(Some(FloatKind::F64) | None) => {
+                    value.parse::<f64>().ok()?.to_string()
+                }
+                // implement once `f16` and `f128` are available
```
GuillaumeGomez, line 2306: Please add a `FIXME` here and open an issue so it's not forgotten.

GuillaumeGomez:
I think you also forgot to add tests for floats like `123e456 `.

Kijewski:
> I think you also forgot to add tests for floats like `123e456 `.

Should be encompassed by `four_thousand()`, isn't it? (That was the actual example how I found the error.)

GuillaumeGomez:
Ah I see. Please also add it into `test_num_lit` so it's easy to see what's tested and what's not.

Kijewski:
> Ah I see. Please also add it into `test_num_lit` so it's easy to see what's tested and what's not.

Yeah, makes sense to keep it at one place. Done.

rinja_derive/src/generator.rs
Kijewski: Added.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #144, opened 2024-08-20
doc: use `feature(doc_cfg, doc_auto_cfg)` everywhere

Kijewski:


fuzzing/fuzz/Cargo.toml
```diff
@@ -8,6 +8,10 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--generate-link-to-definition", "--cfg=docsrs"]
```
GuillaumeGomez, line 13: I love seeing my nightly features being used more and more. :)

GuillaumeGomez:
Thanks!

fuzzing/fuzz/Cargo.toml
Kijewski: Thank you for implementing them! :)

PR #152, opened 2024-08-21
Fix locals handling

GuillaumeGomez:
First commit comes from https://github.com/rinja-rs/rinja/pull/151.

In this case, the locals push is done in one place and all the code using this pushed context is inside the callback, preventing to ever forget the `pop` and potentially preventing us from bad code changes again. 

What do you think of this approach?

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -104,6 +104,16 @@ impl<'a> Generator<'a> {
         Ok(buf.buf)
     }
 
+    fn push_locals<T, F: FnOnce(&mut Self) -> Result<T, CompileError>>(
+        &mut self,
+        callback: F,
+    ) -> Result<T, CompileError> {
+        self.locals.scopes.push(HashMap::default());
+        let res = callback(self);
+        self.locals.scopes.pop().unwrap();
+        res
+    }
+
     // Implement `Template` for the given context struct.
     fn impl_template(&mut self, ctx: &Context<'a>, buf: &mut Buffer) -> Result<(), CompileError> {
         self.write_header(buf, format_args!("{CRATE}::Template"), None);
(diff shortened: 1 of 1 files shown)


Kijewski:
Thank you for working on this! I'll try to review tomorrow evening.

rinja_derive/src/generator.rs
```diff
@@ -104,6 +104,16 @@ impl<'a> Generator<'a> {
         Ok(buf.buf)
     }
 
+    fn push_locals<T, F: FnMut(&mut Self) -> Result<T, CompileError>>(
```
Kijewski, line 107: Can be `FnOnce`, so the wrapped code knows that it may "use up" any variables in its closure.

rinja_derive/src/generator.rs
```diff
@@ -104,6 +104,16 @@ impl<'a> Generator<'a> {
...
+        &mut self,
+        mut callback: F,
```
Kijewski, line 109: This variable does not need to be `mut`, then.

rinja_derive/src/generator.rs
```diff
@@ -558,85 +568,92 @@ impl<'a> Generator<'a> {
...
 
-        for (pos, cond_info) in conds.conds.iter().enumerate() {
+        let mut iter = conds.conds.iter().enumerate().peekable();
+        while let Some((pos, cond_info)) = iter.next() {
             let cond = cond_info.cond;
 
-            self.handle_ws(cond.ws);
-            flushed += self.write_buf_writable(ctx, buf)?;
-            if pos > 0 {
-                self.locals.pop();
```
Kijewski, line 567: I like that this line is gone. Looks extremely error-prone.

Kijewski (changes requested):
Except for the `FnMut` → `FnOnce` bound, it looks very good, thank you! I was afraid this would be a much bigger change, but (ignoring indentation changes) this is very palatable.

rinja_derive/src/generator.rs
GuillaumeGomez: Yes, I don't like states that you need to not forget about changing yourself. I'm very fan of callback approach for this reason (even in other language like C and C++).

GuillaumeGomez:
Changed the type to `FnOnce`, good catch!

rinja_derive/src/generator.rs
GuillaumeGomez: Good catch!

Kijewski (approved):
:+1: thank you!

PR #157, opened 2024-08-28
fuzz: add minified corpus after 1,000,000,000 runs

Kijewski:


Kijewski:
Or another branch, maybe? Yeah, gitg is not happy to display 4,000 changes files. :)

In <https://github.com/Kijewski/tzdb> I use a different branch of the same project as submodule and it works well. But I guess it can be confusing?

GuillaumeGomez:
A bit yes. Let's keep things as simple as possible. ;)

Kijewski:
Okay, I'll add a new repo and make this PR use it as submodule.

I'm building individual corpuses (corpusses? corpora?) for the `filters`, `html` and `parser` fuzz targets right now. In the lecture-free period, our compute servers are collecting dust anyway, so I can give them a purpose. :)

GuillaumeGomez:
Hahaha. That's convenient! :D

GuillaumeGomez (approved):

Kijewski:
Oof, somehow `cargo binstall` installed the musl version of cargo-fuzz, which led me to a wild-goose chase how to get the whole thing running with musl. I manually download the pre-compiled cargo-fuzz, now, and everything seems to work fine.

GuillaumeGomez (approved):

GuillaumeGomez:
Nice work! That was quite the wild ride. 😆

PR #160, opened 2024-08-30
ci: fix a few typos and add check

### LINKED ISSUE #70: Add typo check in CI
Things to be careful about (comment from [here](https://github.com/rinja-rs/rinja/pull/68#issuecomment-2225113426)):

> There's a ton of false positives, like converting "lits" (literals) to "list". The tools actually wants to fix the mock Latin in lorem ipsums. I don't know how much of a hassle it would be to set it up right. In most cases "lits" is a typo, just in in one case it isn't.

Kijewski:
This PR adds `_typos.toml`, which lets `typos` ignore generated data. ~~Some filler texts are "fixed", too. It does not matter what the actual text is.~~

Resolves #70.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,26 @@
+[default]
+locale = "en-us"
+
+[files]
+extend-exclude = [
+    # generated files
+    "book/ethicalads-theme.css",
+    "fuzzing/fuzz/artifacts/",
+    "fuzzing/fuzz/corpus/",
+    "target/",
+    "rinja_parser/tests/*.txt",
+    # fillter texts
+    "rinja/benches/strings.inc",
+    # too many false positives
+    "testing/tests/gen_ws_tests.py",
+]
+
+[default.extend-words]
+# French words
+exemple = "exemple"
+existant = "existant"
+# used in tests
+Ba = "Ba"
+fo = "fo"
+Fo = "Fo"
+sur = "sur"
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -54,7 +54,7 @@ Hello
 To be noted, if one of the trimmed characters is a newline, then the only
 character remaining will be a newline.
 
-If you want this to be the default behaviour, you can set `whitespace` to
+If you want this to be the default behavior, you can set `whitespace` to
 `"minimize"`.
 
 To be noted: you can also configure `whitespace` directly into the `template`
--- a/rinja/benches/escape.rs
+++ b/rinja/benches/escape.rs
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use rinja::filters::{escape, Html};
 
 criterion_main!(benches);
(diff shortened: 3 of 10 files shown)


Kijewski:
A bug in <https://github.com/crate-ci/typos> lets the program run forever. Investigating.

_typos.toml
```diff
@@ -0,0 +1,23 @@
...
+]
+
+[default.extend-words]
+# French words
+exemple = "exemple"
+existant = "existant"
+# fillter words
+Ba = "Ba"
+fo = "fo"
+Fo = "Fo"
```
GuillaumeGomez, line 22: I never saw these two as fillers.

book/src/configuration.md
```diff
@@ -54,7 +54,7 @@ Hello
 To be noted, if one of the trimmed characters is a newline, then the only
 character remaining will be a newline.
 
-If you want this to be the default behaviour, you can set `whitespace` to
+If you want this to be the default behavior, you can set `whitespace` to
```
GuillaumeGomez, line 57: This is british english (left) vs US english (right). Both are valid.

rinja/benches/escape.rs
```diff
@@ -13,7 +13,7 @@ fn escaping(b: &mut criterion::Bencher<'_>) {
     Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris consequat tellus sit
     amet ornare fermentum. Etiam nec erat ante. In at metus a orci mollis scelerisque.
     Sed eget ultrices turpis, at sollicitudin erat. Integer hendrerit nec magna quis
-    venenatis. Vivamus non dolor hendrerit, vulputate velit sed, varius nunc. Quisque
+    venenatis. Vivamus non dolor hendrerit, vulputate velit sed, various nunc. Quisque
```
GuillaumeGomez, line 16: Are we fixing latin too? XD

book/src/configuration.md
Kijewski: Yes, but IMHO we should stick to one locale.

I guess in a French text you would probably replace "nonante", too, even if it is valid Belgian French; as I would probably replace instances if "Jänner" ("January", Swiss) or "heuer" ("this current year", Austrian). :)

rinja/benches/escape.rs
Kijewski: Yea, in all other cases "various" would be right. I hope Cicero wouldn't mind that I changed his text. :)

book/src/configuration.md
GuillaumeGomez: Nah I prefer nonante, makes much more sense. Although I'm surprised by how many people know about this compared to how many people living in France know about it. :rofl:

rinja/benches/escape.rs
GuillaumeGomez: Wouldn't it be better in this case to move this text out of source code?

_typos.toml
Kijewski: Hm, yes, "fo", "Fo" & "Ba" are actually not used in filler texts, but "foO", "FoO", "BaR", in casing tests. Will change the comment.

rinja/benches/escape.rs
Kijewski: Probably. It's used in two places anyway.

Kijewski:
> A bug in https://github.com/crate-ci/typos lets the program run forever. Investigating.

I blacklisted the fuzzed files. Reported upstream: <https://github.com/crate-ci/typos/issues/1088>.

rinja/benches/escape.rs
Kijewski: Moved.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good! I'm quite happy that moving the lorem ipsum out of the code allowed to remove a duplicate. Unexpected benefits. :D

PR #165, opened 2024-09-09
parser: add optional `{% endwhen %}`

### LINKED ISSUE #163: syntax for closing a `{% when ... %}` block within a match
Hey! I have enjoyed using this project and am unsure if anything like this would be possible.

The long and the short of it is that most constructs in the syntax have both a `{% something %}` and an `{% endsomething %}`, this is not the case for match (when) arms.

I understand why this is, and agree it should be the default. But, a lot of html templating tools (formatters in particular) work best when each *block* has a start and end.

This is a simplified example of the current syntax. Formatted with [djlint](https://github.com/djlint/djlint), a common and widely used jinja formatter.

```jinja
{% match self.lpn %}
  {% when None %}
  <div class="alert alert-info">Scan a LPN number to begin.</div>
  {% when Some(lpn_value) %}
  {% match customer_returns %}
    {% when None %}
    <div class="alert alert-danger">
      No Customer Returns found for <strong>{{ lpn_value }}</strong>.
    </div>
    {% else %}
  {% endmatch %}
{% endmatch %}
```
**formatter settings**
```toml
[tool.djlint]
custom_blocks = "match"
ignore_blocks = "when"
```
Notice how the content of the `when` arms in on the same indentation level as the `when` itself. Some people may want it that way, and it should be the default.

## Proposed

What I am suggesting, would be to **optionally** allow closing `when` arms. 
This is produced by the same formatter, with slightly different settings.
```jinja
{% match self.lpn %}
  {% when None %}
    <div class="alert alert-info">Scan a LPN number to begin.</div>
  {% endwhen %}
  {% when Some(lpn_value) %}
    {% match customer_returns %}
      {% when None %}
        <div class="alert alert-danger">
          No Customer Returns found for <strong>{{ lpn_value }}</strong>.
        </div>
      {% endwhen %}
      {% when Some(customer_return_value) %}
      {% endwhen %}
    {% endmatch %}
  {% endwhen %}
{% endmatch %}
```

**formatter settings**
```toml
[tool.djlint]
custom_blocks = "match,when"
ignore_blocks = ""
```
Notice how there is now a clear difference in indenation levels. AFAIK, no formatter for html/jinja templates allows this type of template to be configured for, because it can't know where it ends.

I know the difference is minimal, I know it is not really the crates job to worry about stuff like this. But, I think even semantically, it does make sense for the user to have the ability to specifically close a when arm, even if proc macro just throws it away.

I am willing to write the PR for this, if open to the change

Kijewski:
I think we had the same question before in askama, but I cannot find the issue. Even though I don't use djLint or similar linters, I can see how this feature can be useful to people that do.

Resolves #163.

Cc @chrisp60.


rinja_parser/src/node.rs
```diff
@@ -204,6 +204,29 @@ impl<'a> When<'a> {
...
+                        many0(value((), ws(|i| Comment::parse(i, s)))),
+                    ),
+                )),
+            ))),
+            |(span, (pws, nws))| {
+                let comment = Comment {
+                    ws: Ws(pws, nws),
+                    content: "",
+                };
+                Node::Comment(WithSpan::new(comment, span))
```
Kijewski, line 227: The implementation is kinda lazy, that I add a "virtual" comment node. But I guess it's good enough for the foreseeable future. :)

book/src/template_syntax.md
```diff
@@ -619,6 +619,11 @@ You can also match against multiple alternative patterns at once:
 {% endmatch %}
 ```
 
+For better interoperability with linters and auto-formatters like [djLint],
+you can also use a optional `{% endwhen %}` node to close a `{% when %}` case.
```
GuillaumeGomez, line 623: Maybe add an example?

rinja_parser/src/node.rs
GuillaumeGomez: Please add a comment on this code so future us won't wonder why there is a comment generated here. :laughing:

Kijewski: Hehe, yeah, makes sense to write comments in a place where you would actually find them. :D

book/src/template_syntax.md
Kijewski: Done.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks @Kijewski!

@chrisp60: No problem. It's optional so it doesn't make code longer for people who aren't interested into it and for others, they can use it if they want. So everyone should be happy. :)

PR #167, opened 2024-09-10
parser: add context information for unknown tags

Kijewski:


Kijewski:
Please what the diff ignoring whitespaces changes, or it's entirely unreadable.

rinja_parser/src/node.rs
```diff
@@ -167,6 +198,48 @@ impl<'a> Node<'a> {
...
 
+fn cut_node<'a, O>(
+    kind: Option<&'static str>,
+    inner: impl FnMut(&'a str) -> ParseResult<'a, O>,
+) -> impl FnMut(&'a str) -> ParseResult<'a, O> {
+    let mut inner = cut(inner);
+    move |i: &'a str| {
+        let result = inner(i);
+        if let Err(nom::Err::Failure(err) | nom::Err::Error(err)) = &result {
+            if err.message.is_none() {
```
GuillaumeGomez, line 209: I can't wait for the `if let chain` feature to be stable...

GuillaumeGomez (dismissed):

GuillaumeGomez:
This is super cool, thanks! I'll need to re-approve once merge conflicts are fixed I think. ^^'

Kijewski:
> This is super cool, thanks!

Thank you! :) Yes, oftentimes I was baffled myself why something does not parse correctly, so I hope this PR makes it clearer what is actually going on.

> I'll need to re-approve once merge conflicts are fixed I think. ^^'

Indeed :)

rinja_parser/src/node.rs
Kijewski: Yeah, I would like to remove a lot of `let Some(_) = f() else { return; }`*5 instances, too.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks again!

PR #170, opened 2024-09-11
Implement `pluralize` filter

Kijewski:


rinja/src/filters/builtin.rs
```diff
@@ -528,6 +528,215 @@ pub fn title(s: impl fmt::Display) -> Result<String, fmt::Error> {
     Ok(output)
 }
 
+/// For a value of `±1` by default an empty string `""` is returned, otherwise `"s"`.
+///
+/// # Examples
+///
+/// ## With default arguments
+///
+/// ```rust
```
GuillaumeGomez, line 537: You don't need to add `rust` codeblock attribute, it's the default in rust code examples.

rinja/src/filters/builtin.rs
```diff
@@ -528,6 +528,215 @@ pub fn title(s: impl fmt::Display) -> Result<String, fmt::Error> {
...
+    }
+}
+
+/// An integer that can have the value `+1` and maybe `-1`.
+pub trait PluralizeCount {
+    /// A possible error that can occur while checking the value.
+    type Error: Into<Error>;
+
+    /// Returns `true` if and only if the value is `±1`.
+    fn is_singular(&self) -> Result<bool, Self::Error>;
```
GuillaumeGomez, line 631: Is there an implementation which can return something else than `Ok`?

rinja/src/filters/builtin.rs
Kijewski: Oops, you're right. I copied the code from the `linebreaksbr` filter, which had the attribute, too.

rinja/src/filters/builtin.rs
Kijewski: Because I (intentionally) didn't seal the trait, I want to keep it as generic as possible. If you wanted to, you could implement a fallible lookup for your custom type. E.g. acquiring a lock or reading from a database. I would not advise to do that in a template, but it does not hurt make it possible. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks! Not completely convinced by the not-sealed approach for the trait but we can always revisit it later on.

PR #171, opened 2024-09-11
Fix invalid condition optimization

GuillaumeGomez:
The `!` operator was removed, which is pretty bad. ^^'


Kijewski (approved):
`b || !b`?

Kijewski:
Hehe, that meant as a "too be or not to be" yoke, but more tests are welcome! :)

GuillaumeGomez:
I'm too jetlagged for that. :rofl:

PR #174, opened 2024-09-13
filters: escape arguments to `pluralize` at compile time

Kijewski:
We can tell at compile time how to escape `""` or `"s"`. The former does not need to write anything, the latter does not need to be escaped.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -1809,34 +1809,62 @@ impl<'a> Generator<'a> {
         buf: &mut Buffer,
         args: &[WithSpan<'_, Expr<'_>>],
     ) -> Result<(), CompileError> {
-        if args.is_empty() {
-            return Ok(());
-        }
-
         for (i, arg) in args.iter().enumerate() {
             if i > 0 {
                 buf.write(',');
             }
+            self._visit_arg(ctx, buf, arg)?;
+        }
+        Ok(())
+    }
 
-            let borrow = !is_copyable(arg);
-            if borrow {
-                buf.write("&(");
+    fn _visit_arg(
+        &mut self,
+        ctx: &Context<'_>,
+        buf: &mut Buffer,
+        arg: &WithSpan<'_, Expr<'_>>,
+    ) -> Result<(), CompileError> {
+        let borrow = !is_copyable(arg);
+        if borrow {
+            buf.write("&(");
+        }
+        match **arg {
+            Expr::Call(ref left, _) if !matches!(***left, Expr::Path(_)) => {
+                buf.write('{');
+                self.visit_expr(ctx, buf, arg)?;
+                buf.write('}');
             }
-
-            match **arg {
-                Expr::Call(ref left, _) if !matches!(***left, Expr::Path(_)) => {
-                    buf.write('{');
-                    self.visit_expr(ctx, buf, arg)?;
-                    buf.write('}');
-                }
-                _ => {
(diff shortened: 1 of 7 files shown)


rinja_derive/src/generator.rs
```diff
@@ -1564,20 +1564,44 @@ impl<'a> Generator<'a> {
...
             }
+        };
+        if let Some(is_singular) = expr_is_int_lit_plus_minus_one(count) {
+            let value = if is_singular { sg } else { pl };
+            self._visit_auto_escaped_arg(ctx, buf, value)?;
+        } else {
+            buf.write(format_args!("{CRATE}::filters::pluralize("));
+            self._visit_arg(ctx, buf, count)?;
+            for value in [sg, pl] {
+                buf.write(", ");
```
GuillaumeGomez, line 1599:
```suggestion
                buf.write(',');
```

rinja_derive/src/generator.rs
```diff
@@ -2265,6 +2317,67 @@ impl<'a> Generator<'a> {
...
+        plus_one: T,
+    ) -> Option<bool> {
+        Some(from_str_radix(value, 10).ok()? == plus_one)
+    }
+
+    let Expr::NumLit(_, Num::Int(value, kind)) = **expr else {
+        return None;
+    };
+    match kind {
+        Some(IntKind::I8) => is_signed_singular(i8::from_str_radix, value, 1, -1),
```
GuillaumeGomez, line 2342: I'm trying to think of a way to not have this huge match because it's very error prone to have to ensure that everything is working out correctly... Isn't it be possible somehow to have an impl with an associated type? Another (easier) way is to write a macro where we list types and then everything is generated correctly.

rinja_derive/src/generator.rs
```diff
@@ -2265,6 +2317,67 @@ impl<'a> Generator<'a> {
...
+        Some(IntKind::I16) => is_signed_singular(i16::from_str_radix, value, 1, -1),
+        Some(IntKind::I32) => is_signed_singular(i32::from_str_radix, value, 1, -1),
+        Some(IntKind::I64) => is_signed_singular(i64::from_str_radix, value, 1, -1),
+        Some(IntKind::I128) => is_signed_singular(i128::from_str_radix, value, 1, -1),
+        Some(IntKind::Isize) => {
+            if cfg!(target_pointer_width = "16") {
```
GuillaumeGomez, line 2348: Why not doing `isize::from_str_radix`?

Kijewski: The target and host pointer widths may be different.

rinja_derive/src/generator.rs
Kijewski: I added a macro, but it's not much less ugly.

rinja_derive/src/generator.rs
Kijewski: :+1:

rinja_derive/src/generator.rs
GuillaumeGomez: Ah ok, cross-compilation was the issue.

GuillaumeGomez (approved):

rinja_derive/src/generator.rs
GuillaumeGomez: Indeed but at least less error-prone, well done!

Kijewski:
Thanks! :) Should have thought about that already in the first PR.

GuillaumeGomez:
Well I definitely didn't think about that so that makes two of us.

PR #178, opened 2024-09-16
Run `clippy --fix -- -D clippy::pedantic`

Kijewski:


GuillaumeGomez:
CI is unhappy. Also, do you think it's worth having it in CI?

Kijewski:
> CI is unhappy.

Yeah, I added some `#[must_use]`s manually, and clippy is unhappy if the return type is already `#[must_use]`. Fixed.

> Also, do you think it's worth having it in CI?

There is still a ton of pedantic warnings. Only the automatically fixable ones are changed by this PR. IMHO many warnings in `clippy::pedantic` are, well, overly pedantic. :) But I guess we can gradually enable more clippy lints. E.g. from a users' point-of-view [`missing_errors_doc`](https://rust-lang.github.io/rust-clippy/rust-1.81.0/index.html#/missing_errors_doc) is very useful, but has to be done manually.

rinja_derive/src/input.rs
```diff
@@ -93,7 +93,8 @@ impl TemplateInput<'_> {
 
         let escaping = escaping
             .as_deref()
-            .unwrap_or_else(|| path.extension().map(|s| s.to_str().unwrap()).unwrap_or(""));
+            .or_else(|| path.extension().and_then(|s| s.to_str()))
+            .unwrap_or_default();
```
Kijewski, line 97: This change was done manually.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #180, opened 2024-09-19
Improve config delimiters characters check

GuillaumeGomez:
Finally I have something before the next release. :)

### DIFF
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -806,13 +806,13 @@ impl<'a> SyntaxBuilder<'a> {
             comment_end: self.comment_end.unwrap_or(default.comment_end),
         });
 
-        for (s, k) in [
-            (syntax.block_start, "opening block"),
-            (syntax.block_end, "closing block"),
-            (syntax.expr_start, "opening expression"),
-            (syntax.expr_end, "closing expression"),
-            (syntax.comment_start, "opening comment"),
-            (syntax.comment_end, "closing comment"),
+        for (s, k, is_closing) in [
+            (syntax.block_start, "opening block", false),
+            (syntax.block_end, "closing block", true),
+            (syntax.expr_start, "opening expression", false),
+            (syntax.expr_end, "closing expression", true),
+            (syntax.comment_start, "opening comment", false),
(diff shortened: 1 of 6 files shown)


rinja_parser/src/lib.rs
```diff
@@ -826,6 +826,14 @@ impl<'a> SyntaxBuilder<'a> {
                     "delimiters may not contain white spaces. \
                         The {k} delimiter ({s:?}) contains white spaces",
                 ));
+            } else if is_closing
+                && ['(', ')', '.', '>', '<', '&', '|', '!'].contains(&s.chars().next().unwrap())
```
Kijewski, line 830: `-`, `+`, `~`, too, because they are used for whitespace control. I think that's the problem in [testing/tests/ui/broken-config.stderr](https://github.com/rinja-rs/rinja/pull/180/files#diff-12dfffd906f3d77d1c4e493b55f70d28ffd45498e2d36bf824e3540fcfac199b).

Kijewski (changes requested):
Thank you for the PR! It's much better with your change, to look for an offending character than to enumerate all combinations that could contain this character.

GuillaumeGomez:
Finally had some time to update this. I added the missing operators, thanks for the suggestion!

rinja_parser/src/lib.rs
```diff
@@ -824,6 +824,15 @@ impl<'a> SyntaxBuilder<'a> {
                     "delimiters may not contain white spaces. \
                         The {k} delimiter ({s:?}) contains white spaces",
                 ));
+            } else if is_closing
+                && ['(', ')', '-', '+', '~', '.', '>', '<', '&', '|', '!']
```
Kijewski, line 828: Actually, `)` should be fine. We only ever look for `)` in conjunction with a `(`. This way, if someone wants to use e.g. `(( expr ))` it would still work fine. Sorry that I didn't notice that during my first review!

GuillaumeGomez: Nice catch!

GuillaumeGomez:
Updated. :)

Kijewski (approved):
Good work, thanks! :)

Issue #185, opened 2024-09-25
Convert filter tests to use doc tests


Kijewski:
Most built-in filter tests are invoked as normal functions: [builtin.rs#L614-L817](<https://github.com/rinja-rs/rinja/blob/80bb96931b1a325d0ce74af032d62b222c39a233/rinja/src/filters/builtin.rs#L614-L817>). It is good that the functions are tested at all, but it too bad that this kind of test cannot not serve as documentation. Some filters are doc-tested using rinja code blocks [builtin.rs#L53-L72](<https://github.com/rinja-rs/rinja/blob/80bb96931b1a325d0ce74af032d62b222c39a233/rinja/src/filters/builtin.rs#L53-L72>), so the user gets a proper usage example in the documentation. Also, many filters contain some logic in the generator, which has to be tested, too.

It would be nice if some, if not all, filters could be tested with doc tests:

- [x] fmt
- [x] format
- [x] linebreaksbr
- [x] paragraphbreaks
- [x] lower
- [x] lowercase
- [x] upper
- [x] uppercase
- [x] trim
- [x] truncate
- [x] indent
- [x] join
- [x] capitalize
- [x] wordcount
- [x] title
- [x] pluralize
- [x] safe
- [x] escape
- [x] e
- [x] filesizeformat
- [x] json
- [x] urlencode
- [x] urlencode_strict

Wollaston:
I'd be interested in helping out with this one!

Please let me know if you are looking for new contributors and if so, if you would like me to just go through your checklist here top to bottom. My understanding is that you would like doctests added to these filters. Are we adding new tests, or just copying the regular tests to doctests? Thank you!

GuillaumeGomez:
New contributors and contributions are very welcome! From what I can see, in case the filter isn't tested, then you need to create one. If one already exist, I think copying it might be enough.

Kijewski:
Thank you very much, @Wollaston, great work!

PR #187, opened 2024-09-29
Add doctests to filters

Wollaston:
This PR advances work towards resolving #185. The basic structures of the doctests in this PR follow the example given in that issue. A few items I have not resolved prior to raising this PR:

- I was unable to find a way use variables in the `abs` doctest. I was running into an issue where the `Template` macro wanted to use a reference to an `isize` or `i32`, but the `Signed` trait is not implemented for a reference to those types. So, I just did a static example.
- Similarly for the `filters::num_traits`, I was not able to get examples for `into_f64` or `into_isize` working. My tests would just return the same number passed in unchanged. For example, if I passed in a `3` as a variable, I would get a `3` as the return and it would not be formatted as a `f64` (in the `into_f64` example). 

I am happy to continue working on this issue, especially if there is any feedback or guidance to incorporate into a future PR based on the work to date. Many thanks!

### DIFF
--- a/rinja/src/filters/builtin.rs
+++ b/rinja/src/filters/builtin.rs
@@ -20,6 +20,25 @@ const MAX_LEN: usize = 10_000;
 /// {{ value|fmt("{:?}") }}
 /// ```
 ///
+/// ```
+/// # #[cfg(feature = "code-in-doc")] {
+/// # use rinja::Template;
+/// /// ```jinja
+/// /// <div>{{ value|fmt("{:?}") }}</div>
+/// /// ```
+/// #[derive(Template)]
+/// #[template(ext = "html", in_doc = true)]
+/// struct Example {
+///     value: (usize, usize),
+/// }
+///
+/// assert_eq!(
+///     Example { value: (3, 4) }.to_string(),
+///     "<div>(3, 4)</div>"
+/// );
+/// # }
+/// ```
+///
 /// Compare with [format](./fn.format.html).
 pub fn fmt() {}
 
@@ -34,6 +53,25 @@ pub fn fmt() {}
 /// {{ "{:?}{:?}"|format(value, other_value) }}
 /// ```
 ///
+/// ```
+/// # #[cfg(feature = "code-in-doc")] {
+/// # use rinja::Template;
+/// /// ```jinja
+/// /// <div>{{ "{:?}"|format(value) }}</div>
+/// /// ```
+/// #[derive(Template)]
+/// #[template(ext = "html", in_doc = true)]
+/// struct Example {
+///     value: (usize, usize),
+/// }
+///
+/// assert_eq!(
+///     Example { value: (3, 4) }.to_string(),
+///     "<div>(3, 4)</div>"
+/// );
+/// # }
+/// ```
+///
 /// Compare with [fmt](./fn.fmt.html).
 pub fn format() {}
 
@@ -41,6 +79,25 @@ pub fn format() {}
 ///
 /// A single newline becomes an HTML line break `<br>` and a new line
 /// followed by a blank line becomes a paragraph break `<p>`.
+///
+/// ```
+/// # #[cfg(feature = "code-in-doc")] {
+/// # use rinja::Template;
+/// /// ```jinja
+/// /// <div>{{ example|linebreaks }}</div>
+/// /// ```
+/// #[derive(Template)]
+/// #[template(ext = "html", in_doc = true)]
+/// struct Example<'a> {
+///     example: &'a str,
+/// }
+///
+/// assert_eq!(
+///     Example { example: "Foo\nBar\n\nBaz" }.to_string(),
+///     "<div><p>Foo<br/>Bar</p><p>Baz</p></div>"
+/// );
+/// # }
+/// ```
 #[inline]
 pub fn linebreaks(s: impl fmt::Display) -> Result<HtmlSafeOutput<String>, fmt::Error> {
     fn linebreaks(s: String) -> String {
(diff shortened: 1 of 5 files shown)


rinja/src/filters/builtin.rs
```diff
@@ -20,6 +20,25 @@ const MAX_LEN: usize = 10_000;
 /// {{ value|fmt("{:?}") }}
 /// ```
 ///
+/// ```
+/// # #[cfg(feature = "code-in-doc")] {
```
GuillaumeGomez, line 24: I don't think that's what @Kijewski meant when they suggested to add doctests. Why not using `source` instead of a feature inside a doctest? Would make tests smaller and less complex.

Wollaston: I deferred to the example in the issue ([builtin.rs#L53-L72](https://github.com/rinja-rs/rinja/blob/80bb96931b1a325d0ce74af032d62b222c39a233/rinja/src/filters/builtin.rs#L53-L72)) when adding these. I certainly see why `source` would be a simpler solution though. If that is what you and @Kijewski have in mind, I can update these to use `source` instead of the `code-in-doc` feature. Thanks!

GuillaumeGomez: It's fine. Let's wait for confirmation so you don't update if it's what they have in mind.

Kijewski: Sorry for the late answer. I guess I caught Corona, and the coughing is kinda not so great when you need a good night's sleep to have a working brain. :|

Actually, I meant it this way, in a code block, because IMHO it's much more readable in the generated documentation pages.

If you dislike the parsing overhead a lot, @GuillaumeGomez, then I'm OK with the source attribute, too.

GuillaumeGomez: If it's what you had mind, then it's fine for me too.

GuillaumeGomez: Don't forget to rest @Kijewski !

GuillaumeGomez:
You forgot to run `cargo fmt`. ;)

Wollaston:
Apologies - it had formatted that file using stable `cargo fmt` and swapped the imports ordering. I ran it with nightly `cargo fmt` and it reordered it to match what the CI expected. The new commit reflects that change.

It also looks like it also failed the typos check, but outside of the files I modified. 

If there is something else I may have overlooked, please let me know! Thanks!

rinja/src/filters/builtin.rs
```diff
@@ -230,6 +448,26 @@ impl<W: fmt::Write> fmt::Write for TruncateWriter<W> {
...
+/// /// ```
+/// #[derive(Template)]
+/// #[template(ext = "html", in_doc = true)]
+/// struct Example<'a> {
+///     example: &'a str,
+/// }
+///
+/// assert_eq!(
+///     Example { example: "hello\nfoo\nbar" }.to_string(),
+///     "<div>hello\n    foo\n    bar</div>"
```
GuillaumeGomez, line 467: I'm surprised that the first line is not indented.

GuillaumeGomez: Ok, from [jinja docs](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.indent), by default the first line and blank lines are not indented. We should maybe support to add more options.

GuillaumeGomez: Opened https://github.com/rinja-rs/rinja/issues/193 for keeping track of that.

rinja/src/filters/builtin.rs
```diff
@@ -93,6 +169,31 @@ pub fn paragraphbreaks(s: impl fmt::Display) -> Result<HtmlSafeOutput<String>, f
 }
 
 /// Converts to lowercase
+///
+/// ```
+/// /// ```
```
GuillaumeGomez, line 174: I just realized, why is there this line in this code example and below? I think it's a typo because the doctest is supposed to start at <code>```jinja</code>.

rinja/src/filters/builtin.rs
```diff
@@ -102,12 +203,63 @@ pub fn lower(s: impl fmt::Display) -> Result<String, fmt::Error> {
 }
 
 /// Alias for the `lower()` filter
+/// Converts to lowercase
+///
+/// ```
+/// /// ```
```
GuillaumeGomez, line 209: Same here.

rinja/src/filters/builtin.rs
```diff
@@ -102,12 +203,63 @@ pub fn lower(s: impl fmt::Display) -> Result<String, fmt::Error> {
...
+/// ```
 #[inline]
 pub fn lowercase(s: impl fmt::Display) -> Result<String, fmt::Error> {
     lower(s)
 }
 
 /// Converts to uppercase
+///
+/// ```
+/// /// ```
```
GuillaumeGomez, line 240: Same here (gonna stop at this one).

rinja/src/filters/urlencode.rs
```diff
@@ -56,6 +76,26 @@ pub fn urlencode<T>(s: T) -> Result<HtmlSafeOutput<UrlencodeFilter<T>>, Infallib
...
+/// /// ```
+/// #[derive(Template)]
+/// #[template(ext = "html", in_doc = true)]
+/// struct Example<'a> {
+///     example: &'a str,
+/// }
+///
+/// assert_eq!(
+///     Example { example: "/hello/world" }.to_string(),
+///     "<a href=%2Fhello%2Fworld>Example</a>"
```
GuillaumeGomez, line 95: Nitpicking: please put quote or double quote for the `href` attribute.

Wollaston:
I believe I covered all of your notes in the updated commit. Thank you!

GuillaumeGomez:
Sorry, I opened the tab and forgot about it. ^^'

Looks good to me. Can you just squash your changes into one commit please? If you don't know how, I wrote a small blog post explaining how to do it [here](https://blog.guillaume-gomez.fr/Git-tips/1/2).

Wollaston:
Should be good to go! Thanks for the notes.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Kijewski:
Thank you very much!

PR #188, opened 2024-10-02
docs: Update performance.md

zamazan4ik:
- add documentation about PGO usage for optimizing Rinja's performance. As was asked in https://www.reddit.com/r/rust/comments/1ftx5iv/comment/lpy5v01/

### DIFF
--- a/book/src/performance.md
+++ b/book/src/performance.md
@@ -38,3 +38,7 @@ opt-level = 3
 
 This may affect clean compile times in debug mode, but incremental compiles
 will be faster.
+
+## Profile-Guided Optimization (PGO)
+
+To optimize Rinja's performance, you can compile your application with [Profile-Guided Optimization](https://doc.rust-lang.org/rustc/profile-guided-optimization.html). According to the [tests](https://github.com/mitsuhiko/minijinja/pull/588#issuecomment-2387957123), PGO can improve the library performance by 15%.


book/src/performance.md
```diff
@@ -38,3 +38,7 @@ opt-level = 3
 
 This may affect clean compile times in debug mode, but incremental compiles
 will be faster.
+
+## Profile-Guided Optimization (PGO)
+
+To optimize Rinja's performance, you can compile your application with [Profile-Guided Optimization](https://doc.rust-lang.org/rustc/profile-guided-optimization.html). According to the [tests](https://github.com/mitsuhiko/minijinja/pull/588#issuecomment-2387957123), PGO can improve the library performance by 15%.
```
GuillaumeGomez, line 44: We write `rinja` in lowercase so please update. :)

Could you also explain in the book directly instead of linking to a github issue? The explanations there are great so I think you can mostly copy-paste it as is.

zamazan4ik: > We write rinja in lowercase so please update. :)

Sorry - I just copied it from line 30th "... Rinja's derive macros" ;)

> Could you also explain in the book directly instead of linking to a github issue? The explanations there are great so I think you can mostly copy-paste it as is.

Sure. Which details should be copied from the GitHub issue to the book? Just asking because I am sure that you don't need all the details (like my test env). Better clarify upfront then rework multiple times.

GuillaumeGomez: Anything needed to be able to use PGO. What you need to install etc.

zamazan4ik: Instructions about how to use PGO for Rust programs are described in the official Rustc docs that I already linked to the docs: https://doc.rust-lang.org/rustc/profile-guided-optimization.html - it's an official instruction about how to implement PGO for Rust programs (including a note about about `cargo-pgo`). My GitHub issue describes how to reproduce my benchmark results. I guessed that target users are not interested in reproducing my benchmarks but interested in optimizing their `rinja`-based apps.

zamazan4ik: So maybe just remove the GitHub link at all from the docs? :)

Kijewski:
Interesting! @zamazan4ik, do you know if it's possible to see/evaluate the results of the profiling? If it is, then maybe we could add a **few** `likely!()`/`unlikely!()` hints to the code.

zamazan4ik:
> Interesting! @zamazan4ik, do you know if it's possible to see/evaluate the results of the profiling? If it is, then maybe we could add a few likely!()/unlikely!() hints to the code.

Well, it's possible to achieve via analyzing "before" vs "after" assembly LLVM IR and guessing the required changes (or extracting optimization feedback from LLVM). However, I highly recommend don't do it in practice since this way has many drawbacks:

* It's not possible to bring all PGO-based information to the compiler via intrinsic. Like inline weights -  the `inline` attribute doesn't have such a precision
* Different people can have different workloads and your manually written hints can not be the best options for their cases. The conflicts between user hints and PGO-provided information are the dark area in compilers and are completely undocumented - this behavior can be changed from one compiler release to another.

So my personal recommendation is just leaving the things to PGO and spend your efforts to something more important to the projects like high-level/algortihm optimizations that still cannot be done by compilers (at least yet).

GuillaumeGomez:
Thanks!

PR #189, opened 2024-10-02
Improve handling of `*` and `&` operators in functions/methods arguments

### LINKED ISSUE #186: Incoherent behaviour around references
I was writing:

```jinja
{{ foo(bar) }}
```

but rustc complained because `foo` is expecting `T` and received `&T`. So instead I wrote:

```jinja
{{ foo(*bar) }}
```

and now rustc complained because `bar` (of type `T`) couldn't be dereferenced.

I think we should keep the `&()` wrapping even when we add `*` or `&` to make the behaviour coherent instead of what we currently do.

GuillaumeGomez:
Fixes #186.

With this change, it's now actually possible to use correctly `*` and `&` operators for functions/methods arguments based on rustc errors.


rinja_derive/src/generator.rs
```diff
@@ -1957,7 +1957,23 @@ impl<'a> Generator<'a> {
...
+        buf: &mut Buffer,
+        arg: &WithSpan<'_, Expr<'_>>,
+        // This variable is needed because `Expr::Unary` is not copyable but since we might
+        // skip a few levels.
+        need_borrow: bool,
+    ) -> Result<(), CompileError> {
+        if let Expr::Unary(expr @ ("*" | "&"), ref arg) = **arg {
+            buf.write(expr);
+            return self._visit_arg_inner(ctx, buf, arg, true);
+        }
```
Kijewski, line 1975: I think an `else` case would make it more clear what is going on.

Kijewski (approved):
Except for my two nit-picks, I like the change! Makes the use of references much more easy in templates.

rinja_derive/src/generator.rs
GuillaumeGomez: I like early returns so that I don't need `else` blocks. ^^'

Gonna add an empty line to have a visual separation though.

GuillaumeGomez:
(somewhat) applied your comments, so merging. Thanks for the review! :)

PR #190, opened 2024-10-03
derive: remove `once_map` dependency

Kijewski:
This change removes 14 (transitive) dependencies:

* ahash, getrandom, hashbrown, lock_api, once_map, parking_lot, parking_lot_core, redox_syscall, scopeguard, smallvec, stable_deref_trait, wasi, zerocopy, zerocopy_derive

The new implementation is less generic, but on the other hand it's only 30 non-empty lines of code.

The benchmarks are just about unchanged.

### DIFF
--- a/rinja_derive/src/lib.rs
+++ b/rinja_derive/src/lib.rs
@@ -10,10 +10,12 @@ mod input;
 #[cfg(test)]
 mod tests;
 
-use std::borrow::Cow;
-use std::collections::HashMap;
+use std::borrow::{Borrow, Cow};
+use std::collections::hash_map::{Entry, HashMap};
 use std::fmt;
+use std::hash::{BuildHasher, Hash};
 use std::path::Path;
+use std::sync::Mutex;
 
 use config::{Config, read_config_file};
 use generator::{Generator, MapChain};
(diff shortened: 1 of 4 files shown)


rinja_derive/src/lib.rs
```diff
@@ -379,6 +382,40 @@ impl fmt::Display for MsgValidEscapers<'_> {
     }
 }
 
+#[derive(Debug)]
+struct OnceMap<K, V>([Mutex<HashMap<K, V, FxBuildHasher>>; 8]);
```
GuillaumeGomez, line 386: If it comes from the `once_map` crate, can you add a comment saying so please?

Kijewski: The API idea comes from `once_map`, but I copied no code. I did not look into their source to implement my struct, but I daresay their code is not much different. I would not call my code derivative work. Still, I added a notice of the original authorship of the API.

GuillaumeGomez: Oh then doesn't matter. Impressive work!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the improvement. Please merge once CI is happy. :)

Issue #191, opened 2024-10-03
Declarative `macro_rules!` derive macros (currently an unimplemented, pre-feature-gated RFC)


Kijewski:
With <https://github.com/rust-lang/rfcs/pull/3698> it would be possible to remove all integration features like `with-axum`. Instead it would be possible to implement the code generation as a proper macro that is passed as a call-back argument for our generator (untested pseudo-code):

```rust
#[doc(hidden)]
pub use {axum_core, rinja};

#[macro_export]
macro_rules! Template {
    ($($tt:tt)*) => {
        $crate::impl_template! {
            [$($tt)*]
            [$crate::rinja]
            [$crate::impl_integration] // in `rinja` itself this argument would be absent
        }
    }
}

#[doc(hidden)]
#[macro_export]
macro_rules! impl_integration {
    (
        [$ident:ident]
        [$($impl_generics:tt)*]
        [$($ty_generics:tt)*]
    ) => {
        impl $($impl_generics)* $crate::axum_core::response::IntoResponse
        for $ident $($ty_generics)* {
            #[inline]
            fn into_response(self) -> $crate::axum_core::response::Response {
                $crate::into_response(&self)
            }
        }
    }
}
```

This way `--all-features` would work fine, and everyone could easily implement their own integration crate.

PR #192, opened 2024-10-04
generator: better macro hygiene

Kijewski:
* import an alias `rinja`, so we need less string concatenation at runtime
* work fine in a `#![no_implicit_prelude]` context, even if `std` and/or `core` are shadowed

This PR is done in anticipation of #191, but IMHO it is also a good change on its own.

Once the new feature is added (which probably will take quite some time), we can replace `extern crate #CRATE as rinja` with `use #macro_input as rinja` whereby `macro_input` is the identifier `$crate` of the macro.

### DIFF
--- a/rinja/src/helpers.rs
+++ b/rinja/src/helpers.rs
@@ -1,5 +1,3 @@
-#![doc(hidden)]
-
 use std::cell::Cell;
 use std::fmt;
 use std::iter::{Enumerate, Peekable};
--- a/rinja/src/helpers.rs
+++ b/rinja/src/helpers.rs
@@ -4,6 +4,10 @@ use std::cell::Cell;
 use std::fmt;
 use std::iter::{Enumerate, Peekable};
 
+// The re-exports are used in the generated code for macro hygiene. Even if the paths `::core` or
+// `::std` are shadowed, the generated code will still be able to access the crates.
+pub use {core, std};
+
 use crate::filters::FastWritable;
 
 pub struct TemplateLoop<I>
--- a/rinja/src/lib.rs
+++ b/rinja/src/lib.rs
@@ -67,6 +67,8 @@ mod html;
 use std::{fmt, io};
 
 pub use rinja_derive::Template;
+#[doc(hidden)]
+pub use {core, std};
 
 #[doc(hidden)]
 pub use crate as shared;
--- a/rinja/src/lib.rs
+++ b/rinja/src/lib.rs
@@ -67,8 +67,6 @@ mod html;
 use std::{fmt, io};
 
 pub use rinja_derive::Template;
-#[doc(hidden)]
-pub use {core, std};
 
 #[doc(hidden)]
 pub use crate as shared;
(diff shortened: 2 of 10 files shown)


rinja/src/lib.rs
```diff
@@ -67,6 +67,8 @@ mod html;
 use std::{fmt, io};
 
 pub use rinja_derive::Template;
+#[doc(hidden)]
+pub use {core, std};
```
GuillaumeGomez, line 71: You can refer to `core` and `std` by prepending them with `::` in macros. Like `::core::whatever`. If not, then add a comment to explain why we want it. Also, it's really not great because it's not rare for users to do `use rinja::*;`.

Kijewski: The current code already does that, but even fully-qualified names like `::core` and `::std` can be shadowed. See the example in `testing`.

But yeah, I didn't recognize the problem that `use rinja::*` would import the symbol `core`, too, which might be undesirable. Python is smarter than rust in a way: Symbols with a leading underscore are omitted from glob imports. Would be nice if rust did that, too.

I guess I'll simply move the re-exports into the existing `#[doc(hidden)] mod helpers`. The name already exists, so it can't introduce a new problem.

GuillaumeGomez: A new module containing it would be better indeed. :+1:

rinja/src/helpers.rs
```diff
@@ -4,6 +4,8 @@ use std::cell::Cell;
 use std::fmt;
 use std::iter::{Enumerate, Peekable};
 
+pub use {core, std};
```
GuillaumeGomez, line 7: Please still add a comment explaining why we have this import.

GuillaumeGomez: Also: the `helpers` module is also public and therefore might be imported as `use rinja::helpers::*`. So maybe put it into a hidden `reexport` module or something along the line?

Kijewski: It is `#[doc(hidden)]`. If someone does indeed glob-imports it, then they deserve all the pain they get.

GuillaumeGomez: Hum no it's public: https://github.com/rinja-rs/rinja/blob/master/rinja/src/lib.rs#L64. Or maybe you meant something else? But currently as I understand it, the module is public and the reexports are public too.

Kijewski: I placed the attribute inside the module itself:
<https://github.com/rinja-rs/rinja/blob/4b4418980909122bcbd187f093934a8a699af3e2/rinja/src/helpers.rs#L1>

Maybe that was not the best idea. :)

GuillaumeGomez: Damn, completely missed it, sorry about that. If you don't mind, please put it on the import instead so future me won't be lost. :laughing:

Kijewski: Hehe, yeah, it was a quite unconventional placement. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Great work, thanks!

Kijewski:
Thank you! And, as usual, thank you for reviewing! :)

PR #195, opened 2024-10-08
filters: remove `num-traits`

Kijewski:
They were added in <https://github.com/djc/askama/pull/146>.

The example when they might be useful sounds contrived. You can always manually add the package `num-traits` to your project and use the methods directly. `into_isize` and `into_f64` wasn't even documented in the book.


GuillaumeGomez:
Sounds good to me, thanks!

PR #197, opened 2024-10-08
Acually remove `once_map` dependency

Kijewski:
It was only removed from `rinja_derive_standalone` before.


GuillaumeGomez (approved):

GuillaumeGomez:
Woups. Good catch haha.

PR #199, opened 2024-10-12
"object safe" is now "dyn-compatible"

Kijewski:
The phrase was changed in <https://github.com/rust-lang/rust/issues/130852>.

### DIFF
--- a/rinja/src/lib.rs
+++ b/rinja/src/lib.rs
@@ -174,9 +174,11 @@ impl<T: Template + ?Sized> Template for &T {
     const MIME_TYPE: &'static str = T::MIME_TYPE;
 }
 
-/// Object-safe wrapper trait around [`Template`] implementers
+/// [`dyn`-compatible] wrapper trait around [`Template`] implementers
 ///
-/// This trades reduced performance (mostly due to writing into `dyn Write`) for object safety.
+/// This trades reduced performance (mostly due to writing into `dyn Write`) for dyn-compatibility.
+///
+/// [`dyn`-compatible]: <https://doc.rust-lang.org/stable/reference/items/traits.html#object-safety>
 pub trait DynTemplate {
     /// Helper method which allocates a new `String` and renders into it
     fn dyn_render(&self) -> Result<String>;


rinja/src/lib.rs
```diff
@@ -174,9 +174,11 @@ impl<T: Template + ?Sized> Template for &T {
     const MIME_TYPE: &'static str = T::MIME_TYPE;
 }
 
-/// Object-safe wrapper trait around [`Template`] implementers
+/// [`dyn`-compatible] wrapper trait around [`Template`] implementers
 ///
-/// This trades reduced performance (mostly due to writing into `dyn Write`) for object safety.
+/// This trades reduced performance (mostly due to writing into `dyn Write`) for dyn-compatibility.
+///
+/// [`dyn`-compatible]: <https://doc.rust-lang.org/stable/reference/items/traits.html#object-safety>
```
Kijewski, line 181: Waiting for <https://github.com/rust-lang/reference/pull/1512> to land, so the link fragment can be updated.

Kijewski: Implemented in <https://github.com/rust-lang/reference/pull/1666>. The new version is not live, yet, but it will be soon enough.

Kijewski: I don't know when it was releases, but the new link live now.

rinja/src/lib.rs
```diff
@@ -170,9 +170,11 @@ impl<T: Template + ?Sized> Template for &T {
     const MIME_TYPE: &'static str = T::MIME_TYPE;
 }
 
-/// Object-safe wrapper trait around [`Template`] implementers
+/// [`dyn`-compatible] wrapper trait around [`Template`] implementers
```
GuillaumeGomez, line 173:
```suggestion
/// [`dyn`-compatible] wrapper trait around [`Template`] implementors.
```

In rust docs it's "implementors" so I suppose it's the correct word. ^^'

Kijewski: Oops, thank you for proofreading!

(I'm like 99% sure that in English you are supposed to role a dice when to write -ter and when to write -tor. :D)

GuillaumeGomez: Yup, and let's not even try to speak english, the pronunciation is even more random. 😆

Kijewski:
The new link works now with the rust 1.84 release.

rinja/src/lib.rs
```diff
@@ -180,9 +180,11 @@ impl<T: Template + ?Sized> Template for &T {
     const SIZE_HINT: usize = T::SIZE_HINT;
 }
 
-/// Object-safe wrapper trait around [`Template`] implementers
+/// [`dyn`-compatible] wrapper trait around [`Template`] implementors.
 ///
-/// This trades reduced performance (mostly due to writing into `dyn Write`) for object safety.
+/// This trades reduced performance (mostly due to writing into `dyn Write`) for dyn-compatibility.
+///
+/// [`dyn`-compatible]: <https://doc.rust-lang.org/stable/reference/items/traits.html#dyn-compatibility>
```
GuillaumeGomez, line 187: Not sure it's valid markdown.

```suggestion
/// [`dyn`-compatible]: https://doc.rust-lang.org/stable/reference/items/traits.html#dyn-compatibility
```

Kijewski: It is. You can even try it in this chat. :)

GuillaumeGomez: So weird. Normally you use `<>` to linkify a link directly in text, not in definition. Is there any advantage to use this syntax here?

Kijewski: When used in inline text, you have to surround a link with `<…>` so it becomes clickable. I always add braces in all places, so it always looks the same. (I think github-compatible markdown makes anything clickable that starts with `http[s]://`.)

PR #203, opened 2024-10-17
derive: in complex conditions, wrap terms in `as_bool()`

### LINKED ISSUE #201: In an `if` condition with conjuctions `&&` and disjunctions `||`, should each term be "boolified"?
In a condition `{% if (a || b) && (c || d) %}`, rinja ensures that the *whole expression* `P` is a `bool` by (de)referencing as needed: `*((&P) as &bool)`. I think it would be better if each term was "boolified" `B`:

```diff
-B((a || b) && (c || d))
+B((B(a) || B(b)) && (B(C) || B(D)))
```

This "boolification" should be done for the outmost term of a condition, and any nested conjuctions and disjuctions. Any sub-expression that is not `…||…`, `…&&…` or `!…` should end the recursive boolification.

Compare <https://github.com/redlib-org/redlib/pull/290>.

Kijewski:
Before this PR, the whole condition in `{% if cond %}` got wrapped in `as_bool(&cond)`. With is PR, the individual terms in a binary operation `&&` or `||`, or unary operation `!` get wrapped in `as_bool()`.

Resolves #201.

### DIFF
--- a/rinja_derive/src/tests.rs
+++ b/rinja_derive/src/tests.rs
@@ -293,7 +293,7 @@ writer.write_str("bla")?;"#,
         "{% if x == 12 %}bli
          {%- else if x is defined %}12
          {%- else %}nope{% endif %}",
-        r#"if *(&(self.x == 12) as &rinja::helpers::core::primitive::bool) {
+        r#"if rinja::helpers::as_bool(&(self.x == 12)) {
 writer.write_str("bli")?;
 } else {
 writer.write_str("12")?;
@@ -306,7 +306,7 @@ writer.write_str("12")?;
     // are present.
     compare(
         "{% if y is defined || x == 12 %}{{x}}{% endif %}",
-        r"if *(&(self.x == 12) as &rinja::helpers::core::primitive::bool) {
+        r"if rinja::helpers::as_bool(&(self.x == 12)) {
     match (
         &((&&rinja::filters::AutoEscaper::new(&(self.x), rinja::filters::Text)).rinja_auto_escape()?),
     ) {
@@ -347,7 +347,7 @@ writer.write_str("12")?;
     compare(
         "{% if y is defined && y == 12 %}{{y}}{% else %}bli{% endif %}",
         r#"
-if *(&(self.y == 12) as &rinja::helpers::core::primitive::bool) {
+if rinja::helpers::as_bool(&(self.y == 12)) {
     match (
         &((&&rinja::filters::AutoEscaper::new(
             &(self.y),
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
In travel currently. Will take a look when I'm back next week. :)

Kijewski:
No problem at all! Enjoy your trip, I wish you good weather! :) 🌞

rinja_derive/src/tests.rs
```diff
@@ -140,7 +140,7 @@ fn check_if_let() {
     // In this test, we ensure that `query` never is `self.query`.
     compare(
         "{% if let Some(query) = s && !query.is_empty() %}{{query}}{% endif %}",
-        r"if let Some(query,) = &self.s && !query.is_empty() {
+        r"if let Some(query,) = &self.s && !rinja::helpers::as_bool(&(query.is_empty())) {
```
GuillaumeGomez, line 143: In such cases, I'm not sure it's an improvement since the `!` operator tells us that it's already a boolean.

rinja_derive/src/tests.rs
```diff
@@ -569,7 +569,10 @@ fn check_bool_conditions() {
     );
     compare(
         "{% if y == 3 || (x == 12 || true) %}{{x}}{% endif %}",
-        r"if *(&(self.y == 3 || (self.x == 12 || true)) as &rinja::helpers::core::primitive::bool) {
+        r"
+if rinja::helpers::as_bool(&(self.y == 3))
```
GuillaumeGomez, line 573: Same in this case, it's a comparison so we already know for sure it's a boolean.

GuillaumeGomez:
It's work travel so the weather doesn't really matter. :laughing:

Kijewski:
`!` is implemented for `bool` and `&bool`, but not `&&bool`, `&mut bool`, `Rc<bool>`, …

`rinja::helpers::as_bool()` follows all references to get the right type, so the user does doesn't have to know or care what the actual type is, only you can add some amount of `*` in front of it to get a `bool`. The previous `*((&value) as &bool)` trick did this halfway: You still have to care what kind of reference wrapper it is, and e.g. `Rc<_>` won't work.

Also, as seen in <https://github.com/redlib-org/redlib/pull/290#issuecomment-2416447943>, the current implementation can be confusing. In the condition `post_should_be_blurred && prefs.unblur_on_hover != "on"`, why do I need to add an `*` is front of the variable `post_should_be_blurred`?

GuillaumeGomez (approved):

GuillaumeGomez:
Oh I see, makes sense. Thanks for the extra explanation. :)

PR #207, opened 2024-10-23
Add support for default value in macros

### LINKED ISSUE #198: Add support for macro argument default value
It's possible in [jinja](https://jinja.palletsprojects.com/en/3.1.x/templates/#macros) to have default value for macro arguments:

```jinja
{% macro input(name, value='', type='text', size=20) -%}
    <input type="{{ type }}" name="{{ name }}" value="{{value|e }}" size="{{ size }}">
{%- endmacro %}
```

GuillaumeGomez:
Fixes #198.

So I went the easy way and allowed to have an expression as default value. Not sure if it's a good or bad idea.

I still need to update the book, but before doing that, I want to ensure with you if expression as default value is ok with you or if I should limit it.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -7,7 +7,8 @@ use std::sync::Arc;
 use std::{cmp, hash, mem, str};
 
 use parser::node::{
-    Call, Comment, Cond, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Match, Whitespace, Ws,
+    Call, Comment, Cond, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Macro, Match,
+    Whitespace, Ws,
 };
 use parser::{
     CharLit, CharPrefix, Expr, Filter, FloatKind, IntKind, Node, Num, StrLit, StrPrefix, Target,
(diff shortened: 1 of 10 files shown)


Kijewski:
> So I went the easy way and allowed to have an expression as default value. Not sure if it's a good or bad idea.

I think that's a good idea! The only question is when to evaluate the expressions:

```jinja
{% macro sum(a=0, b=a, c=a+b, d=a+b+c) %}
  {{a}} {{b}} {{c}} {{d}}
{% endmacro %}
```

In a call to `sum`, should `a` refer to the the parameter `a`, or to an identifier `a` at the calling location? IMO, the former would be vastly more useful.

GuillaumeGomez:
Currently it's to the parameter. But gonna add a test for that.

Well then, since you like it as is, I'll continue on this implementation. :)

GuillaumeGomez:
Ready for review. :)

rinja_parser/src/node.rs
```diff
@@ -568,17 +569,41 @@ impl<'a> Loop<'a> {
...
+macro_rules! check_duplicated_name {
+    ($names:ident, $arg_name:ident, $i:ident) => {
+        if !$names.insert($arg_name) {
+            return Err(nom::Err::Failure(ErrorContext::new(
+                format!("duplicated argument `{}`", $arg_name),
+                $i,
+            )));
+        }
+    };
+}
```
Kijewski, line 586: Why didn't you write that as a function?

rinja_derive/src/generator.rs
```diff
@@ -2959,11 +2978,12 @@ fn is_copyable(expr: &Expr<'_>) -> bool {
...
-        Expr::BoolLit(_) | Expr::NumLit(_, _) | Expr::StrLit(_) | Expr::CharLit(_) => true,
+        Expr::BoolLit(_)
+        | Expr::NumLit(_, _)
+        | Expr::StrLit(_)
+        | Expr::CharLit(_)
+        | Expr::BinOp(_, _, _) => true,
         Expr::Unary(.., expr) => is_copyable_within_op(expr, true),
-        Expr::BinOp(_, lhs, rhs) => {
-            is_copyable_within_op(lhs, true) && is_copyable_within_op(rhs, true)
-        }
```
Kijewski, line 2966: Is this change correct?

rinja_parser/src/node.rs
GuillaumeGomez: I did originally, but forgot why it didn't work out. Gonna give it another try.

rinja_derive/src/generator.rs
GuillaumeGomez: Yes. I discovered that if you do `a + b`, it was generated as `&(a + b)`, which is wrong. I encountered this bug when allowing expressions with macro params as default value.

GuillaumeGomez:
Converted `check_duplicated_name` into a function again.

GuillaumeGomez:
@Kijewski Forgotten or scared to look at it? 😆

Kijewski:
Scared to look at it. :D But I'll to it tomorrow!

GuillaumeGomez:
Only if you have time and motivation. 😉 And thanks in advance!

Kijewski:
I refactored the code a bit to make it more easily understandable to me, and improved the error message for broken argument lists a bit.

Sorry for taking so long to look into this PR! Thank you for implementing it!

Kijewski (approved):

GuillaumeGomez:
Thanks for the improvements (really loved the commit message "kill the Cows" 🤣 )!

PR #212, opened 2024-10-27
Fix crate version numbers in book

BD103:
I was browsing over [the book](https://rinja.readthedocs.io/en/stable/getting_started.html), and noticed that the recommended version numbers were still left over from `askama`. This PR fixes them to be the latest values.

Note that this is a temporary fix, as the version will need to be bumped at the next minor release (0.4, then 0.5, etc.). While it is possible to automate updating these values, my best recommendation is to have a release checklist in `CONTRIBUTING.md` that notes that these values need to be changed.

Cheers :)

### DIFF
--- a/book/src/getting_started.md
+++ b/book/src/getting_started.md
@@ -4,7 +4,7 @@ First, add the following to your crate's `Cargo.toml`:
 
 ```toml
 # in section [dependencies]
-rinja = "0.12.1"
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #214, opened 2024-10-28
Upgrade from Winnow 0.3 to 0.4

epage:
The biggest change in this release is switching from `Fn(I) -> Result<(I, O, E>` to `Fn(&mut I) -> Result<O, E>`. This offers better ergonomics and performance.  "Checkpoints" were introduced as a way to save off a specific location and return to it later, replacing the input juggling done before.

To make this major shift more doable, compatibility APis were added.  `parse_peek` is like the new `parse_next` except it has the old signature.  `unpeek` turns the old signature into the new one.  These are terrible names but meant to be short-lived in a code base.


GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks a lot!

Kijewski:
Thank you very much for your work! This PR was huge.

PR #216, opened 2024-10-31
Relax `filesizeformat` filter requirements

### LINKED ISSUE #215: Allow `filesizeformat` to also work on `&i64` (and any integers between any level of reference)

GuillaumeGomez:
Fixes #215.

Funnily enough, it also allowed to remove one dependency, so yeay! \o/

### DIFF
--- a/rinja/Cargo.toml
+++ b/rinja/Cargo.toml
@@ -21,11 +21,10 @@ rustdoc-args = ["--generate-link-to-definition", "--cfg=docsrs"]
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["config", "humansize", "urlencode"]
+default = ["config", "urlencode"]
 full = ["default", "code-in-doc", "serde_json"]
 code-in-doc = ["rinja_derive/code-in-doc"]
 config = ["rinja_derive/config"]
-humansize = ["rinja_derive/humansize", "dep:humansize"]
 serde_json = ["rinja_derive/serde_json", "dep:serde", "dep:serde_json"]
 urlencode = ["rinja_derive/urlencode", "dep:percent-encoding"]
 
@@ -39,7 +38,6 @@ with-warp = ["rinja_derive/with-warp"]
 [dependencies]
 rinja_derive = { version = "=0.3.5", path = "../rinja_derive" }
 
-humansize = { version = "2", optional = true }
 num-traits = { version = "0.2.6", optional = true }
 percent-encoding = { version = "2.1.0", optional = true }
 serde = { version = "1.0", optional = true }
--- a/rinja/src/filters/humansize.rs
+++ b/rinja/src/filters/humansize.rs
@@ -1,5 +1,9 @@
 use std::convert::Infallible;
 use std::fmt;
+use std::mem::MaybeUninit;
+use std::str::from_utf8_unchecked;
+
+use super::FastWritable;
 
 /// Returns adequate string representation (in KB, ..) of number of bytes
 ///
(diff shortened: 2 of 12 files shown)


GuillaumeGomez:
Finally fixed Ci \o/

rinja_derive/src/generator.rs
```diff
@@ -1657,18 +1657,11 @@ impl<'a> Generator<'a> {
...
-            ));
-        }
-
         // All filters return numbers, and any default formatted number is HTML safe.
         buf.write(format_args!(
-            "rinja::filters::HtmlSafeOutput(rinja::filters::{name}(",
+            "rinja::filters::HtmlSafeOutput(rinja::filters::{name}("
         ));
         self._visit_args(ctx, buf, args)?;
         buf.write(")?)");
```
Kijewski, line 1667:
```suggestion
        buf.write(format_args!(
            "rinja::filters::HtmlSafeOutput(\
                rinja::filters::{name}(\
                    rinja::helpers::get_primitive_value(&("
        ));
        self._visit_args(ctx, buf, args)?;
        buf.write(
            "\
                    )) as rinja::helpers::core::primitive::f32\
                )?\
            )"
        );
```

So `filesizeformat()` does only need to accept f32.

rinja/src/filters/humansize.rs
```diff
@@ -21,24 +61,97 @@ use humansize::{DECIMAL, ISizeFormatter, ToF64};
...
-pub struct FilesizeFormatFilter(f64);
+pub struct FilesizeFormatFilter(f32);
+
+struct NbAndDecimal(usize);
+impl NbAndDecimal {
+    fn new(nb: f32) -> Self {
+        // `nb` will never be bigger than 999_999 so we're fine with usize.
+        Self(nb as _)
+    }
+}
```
Kijewski, line 77: I would explicitly use `NbAndDecimal(u32)`: `usize` would be too small on 16bit systems, and on x64 sytems, working with `u64` is a lot slower than `u32`.

rinja/src/filters/humansize.rs
```diff
@@ -21,24 +61,97 @@ use humansize::{DECIMAL, ISizeFormatter, ToF64};
...
+        } else {
+            f.write_fmt(format_args!("{}", self.0 / 1_000))
+        }
+    }
+}
 
 impl fmt::Display for FilesizeFormatFilter {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_fmt(format_args!("{}", ISizeFormatter::new(self.0, &DECIMAL)))
+        if self.0 < 1_000. {
```
Kijewski, line 95: I wonder if this whole `if-else-if` block could be implemented as a loop: `for (prefix, limit) in [("", 1e3), ("k", 1e6), ..] {`?

GuillaumeGomez: It could, but the advantage of this approach is that we don't need to format the string of the units. :)

rinja/src/filters/humansize.rs
GuillaumeGomez: I wondered about that but I guess you're right.

rinja_derive/src/generator.rs
GuillaumeGomez: I'm a bit shared about this. With the current approach, if users have something like a newtype, they can just implement the `ToF32` trait on it to make it work. With your suggestion, they'd need to make a different kind of conversion before being able to call the filter. It sounds a bit less practical.

rinja/src/filters/humansize.rs
Kijewski: I guess we could use `[" B", " kB", ..]` including the space and 'B', and have fewer write calls.

GuillaumeGomez: Now I'm not sure to follow what you mean. I have in mind:

```rust
for (unit, div) in [(" KB", 1e6), ...] {
    if self.0 < div {
        f.write_fmt(format_args!("{}{unit}", NbAndDecimal::new(self.0)))
    }
}
```

That's still one extra argument formatting.

rinja_derive/src/generator.rs
Kijewski: I would rather make `rinja::helpers::PrimitiveType` public, so a user could implement something like that
```rust
impl PrimitiveType for DiskSpaceRequirements {
    type Value = u64;

    #[inline]
    fn get(&self) -> u64 {
        self.unpacked_size
    }
}
```

Otherwise, if you want to add an explicit trait, I would rename it to something like `AsFilesize`.

GuillaumeGomez: Good idea! Let's go with `PrimitiveType` as public trait. :)

rinja/src/filters/humansize.rs
Kijewski:
```rust
write!(f, "{}", NbAndDecimal::new(self.0))?;
f.write_str(unit)?;
```
The extra call is there in the current implementation, too, it is just hidden in the unrolled code.

GuillaumeGomez: Ok I see. Let's go with it then. Hopefully, loop will be unrolled at compile-time but it's not a hot path so should be fine.

GuillaumeGomez:
Applied suggestions and fixed CI. :D

Kijewski:
I made some changes to the implementation, and pushed it to your PR.

To test if my idea how to optimize `NbAndDecimal` could work, I kinda had to implement it.  I hope that's OK! :-/

GuillaumeGomez:
Yup it's fine. I was going through the code and I discovered that I badly handled the case where the decimal ends with '0' and when the number is still too big (where we should not show decimals). Great improvements! :D

rinja/Cargo.toml
```diff
@@ -39,7 +38,6 @@ with-warp = ["rinja_derive/with-warp"]
 [dependencies]
 rinja_derive = { version = "=0.3.5", path = "../rinja_derive" }
 
-humansize = { version = "2", optional = true }
```
Kijewski, line 42: I think removing a feature is a breaking change. Maybe it would be better to keep it as a noop for now?

GuillaumeGomez: We can make the next release a major one. Problem solved like that. :)

Kijewski: Works for me! :D

Kijewski (approved):

GuillaumeGomez:
Thanks a lot!

PR #219, opened 2024-11-02
Implement `HtmlSafe` for all reference wrappers

Kijewski:


rinja/src/filters/escape.rs
```diff
@@ -486,6 +476,12 @@ where
 {
 }
 
+crate::impl_for_ref! {
```
GuillaumeGomez, line 479: Do we have tests already for ensuring it's working for all references?

Kijewski: Good call! No, there weren't any tests. Added.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #228, opened 2024-11-09
Move types into their tests

GuillaumeGomez:
It's frequent to copy-paste an existing test and its associated (template) type to create a new test. But it's also very possible to use the wrong (template) type from there. Moving the (template) type directly into its test function greatly limit this possibility and it's something that we both started to do recently so with this PR, everything else is moved to this way of writing tests. :)


Kijewski (approved):
Thank you! As you said, it's much easier to read this way, and more difficult to get wrong when you copy a test!

PR #230, opened 2024-11-10
 parser: reject excessively long syntax delimiters 

Kijewski:
Our first fuzzer finding! Excessively long syntax delimiters, e.g. comment_end, causes timeouts. These should be rejected by the parser.

Related issue: <https://issues.oss-fuzz.com/issues/377965573>

Please only merge after: <https://github.com/rinja-rs/rinja/pull/229>


Kijewski:
I would almost call 32 bytes excessive, too, but maybe someone uses emojis with modifiers as delimiters? I would not want to break their projects. :) But the 64.9 kB delimiter that was found by the fuzzer is indeed without a doubt too long.

The found case is the worst case for a naive string searching algorithm, as it starts with many `\0`s, and most of the file is `\0`s, too. In a later PR we could explore if it would be a good idea to replace [`memchr3()`](https://docs.rs/memchr/2.7.4/memchr/fn.memchr3.html) with [`twoway()`](https://docs.rs/memchr/2.7.4/memchr/arch/all/twoway/index.html). On the other hand, this is not a likely case to ever occur in a real scenario, so restricting the length of the delimiters might be a good enough solution as is.

rinja_parser/src/tests.rs
```diff
@@ -1064,3 +1064,39 @@ fn fuzzed_filter_recursion() {
     const TEMPLATE: &str = include_str!("../tests/filter-recursion.txt");
     assert!(Ast::from_str(TEMPLATE, None, &Syntax::default()).is_err());
 }
+
+#[test]
+#[should_panic = "called `Result::unwrap()` on an `Err` value: \
```
GuillaumeGomez, line 1069: I forgot there was `should_panic`. Rust is really so cool.

rinja_parser/src/tests.rs
```diff
@@ -1064,3 +1064,39 @@ fn fuzzed_filter_recursion() {
...
+    let syntax = SyntaxBuilder {
+        name: "test",
+        block_start: Some("\0*DD\0\0"),
+        block_end: Some(
+            "\0]***NEWFILE\u{1f}***:7/v/.-3/\u{1b}/~~~~z~0/*:7/v/./t/t/.p//NEWVILE**::7/v",
+        ),
+        expr_start: Some("/."),
+        expr_end: None,
+        comment_start: Some(include_str!(
+            "../tests/excessive_syntax_lengths-comment-start.bin"
```
GuillaumeGomez, line 1082: Don't you want to split it into its own test?

rinja_parser/src/tests.rs
```diff
@@ -1064,3 +1064,39 @@ fn fuzzed_filter_recursion() {
...
+        )),
+        comment_end: None,
+    }
+    .to_syntax()
+    .unwrap();
+
+    let (sender, receiver) = std::sync::mpsc::channel();
```
GuillaumeGomez, line 1089: Same here: don't you want to split it into its own test? I think this code is meant to ensure that if the previous code doesn't panic, it'll not time out the test. If so, please add a comment.

Kijewski:
What do you think, do we even need to keep the original failing scenario around, or is it okay if I reduce the test to the verify the main problem?

GuillaumeGomez (approved):

GuillaumeGomez:
I think it's perfectly fine as it is now. Since it's filtered out right away and all delimiters are checked, I don't think it's necessary. Thanks for the fix and improvement!

PR #231, opened 2024-11-10
derive: reject `union`s as they require `unsafe` code 

Kijewski:


GuillaumeGomez (approved):

GuillaumeGomez:
I'm not sure how difficult it would be to support `union`. Apart from adding `unsafe` when accessing its fields, is there anything else?

In the meantime, merging this. Thanks!

GuillaumeGomez:
Sounds good to me. Not high priority for me either.

PR #235, opened 2024-11-11
filters: don't collect into intermediate `String` if we can help it

Kijewski:


rinja/src/filters/builtin.rs
```diff
@@ -10,6 +10,28 @@ use crate::{Error, Result};
...
+///
+/// If `$str` is `&str` or `String`, this macro simply passes on its content.
+/// If it is neither, then the formatted data is collection into `&buffer`.
+///
+/// `return`s with an error if the formatting failed.
+macro_rules! try_to_str {
+    ($src:expr => $buffer:ident) => {
+        match format_args!("{}", $src) {
+            args => {
+                if let Some(s) = args.as_str() {
```
GuillaumeGomez, line 23: I didn't know about this method on `Arguments`. Nice!

GuillaumeGomez (approved):

rinja/src/filters/builtin.rs
Kijewski: Yeah, it's a case of "wouldn't it be nice if there was a function that ...". And then the function exists and it does just work. :)

PR #236, opened 2024-11-12
Add concat operator `~`

Kijewski:
The string concatenation operator `~` is present in jinja and tera: <https://jinja.palletsprojects.com/en/stable/templates/#other-operators>. While it's not the most important operator, it can come in handy e.g. with filters: `{{ a|upper }}{{ b|upper }}` → `{{ (a ~ b)|upper }}`.

### DIFF
--- a/rinja_parser/src/expr.rs
+++ b/rinja_parser/src/expr.rs
@@ -8,7 +8,7 @@ use winnow::combinator::{
     terminated,
 };
 use winnow::error::{ErrorKind, ParserError as _};
-use winnow::token::take_till0;
+use winnow::token::{take_till0, take_till1};
 
 use crate::{
     CharLit, ErrorContext, Level, Num, ParseResult, PathOrIdentifier, StrLit, WithSpan, char_lit,
@@ -68,6 +68,7 @@ pub enum Expr<'a> {
     FilterSource,
     IsDefined(&'a str),
     IsNotDefined(&'a str),
+    Concat(Vec<WithSpan<'a, Expr<'a>>>),
 }
 
 impl<'a> Expr<'a> {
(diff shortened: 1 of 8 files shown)


rinja_parser/src/expr.rs
```diff
@@ -176,7 +177,40 @@ impl<'a> Expr<'a> {
     expr_prec_layer!(bxor, band, token_xor);
     expr_prec_layer!(band, shifts, token_bitand);
     expr_prec_layer!(shifts, addsub, alt((">>", "<<")));
-    expr_prec_layer!(addsub, muldivmod, alt(("+", "-")));
+    expr_prec_layer!(addsub, concat, alt(("+", "-")));
+
+    fn concat(i: &'a str, level: Level) -> ParseResult<'a, WithSpan<'a, Self>> {
+        fn concatted(i: &str, level: Level) -> ParseResult<'_, Option<WithSpan<'_, Expr<'_>>>> {
```
GuillaumeGomez, line 183: "concatted"? XD

testing/tests/ui/concat.rs
```diff
@@ -0,0 +1,46 @@
+use rinja::Template;
```
GuillaumeGomez, line 1: Can you also add these two cases:

```
~ a
a ~
```

please?

GuillaumeGomez:
I really love this new feature. Next release will be super huge. :D

rinja_parser/src/expr.rs
Kijewski: Naming things is hard :D

testing/tests/ui/concat.rs
Kijewski: Added.

Kijewski:
Yeah, and I find it interesting that even in a mature project like ours there is always something one can add or improve. :)

testing/tests/ui/concat.stderr
```diff
@@ -0,0 +1,79 @@
...
+
+error: the concat operator `~` must be surrounded by spaces
+ --> <source attribute>:1:4
+       "~ b~}}"
+  --> tests/ui/concat.rs:39:21
+   |
+39 | #[template(source = r#"{{~a~ b~}}"#, ext = "txt")]
+   |                     ^^^^^^^^^^^^^^^
+
+error: failed to parse template source
```
GuillaumeGomez, line 49: That's what I was afraid of. Do you think it's possible to improve error messages for both cases? Like saying that the `~` operator expects two items or something along the line.

GuillaumeGomez:
Useable for sure, mature I don't know. Still a lot of potential improvements to do like enum support. ;)

testing/tests/ui/concat.stderr
Kijewski: I guess in 9 out of 10 cases `{{ a ~ }}` is a misspelling of `{{ a ~}}` instead of a missing right hand operator. Same for `{{ ~ a }}` → `{{~ a}}`.

I don't know how to programmatically tell both cases apart or how to add a better error message here. :-/

GuillaumeGomez: Right, we don't have enough context for that. Then let's keep it as is (sorry, just saw you added a commit ><).

Kijewski:
Something left to do for this PR?

GuillaumeGomez:
Yes. Can you remove the last commit please? ^^' I find the error message less good.

Kijewski:
Sure! Removed.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #241, opened 2024-11-14
parser: allow whitespace control on `extends`

Kijewski:
Using whitespace control on `extends`, e.g. `{%- extends "some.html" ~%}`, is an explicit syntax error. The test was added in <https://github.com/rinja-rs/rinja/commit/29687db996307463ac2f4de01f67704e909dd1fe>.

IMHO, the syntax should be allowed, even though it does not change anything if you add whitespace control, because anything outside of `{% block %}`s gets omitted. Whitespace control on the far sides of `{% block %}`s is not a syntax error, either. For consistency, I like to always write `{%- -%}` on the outmost scope, regardless if I have to or not.


GuillaumeGomez:
Can you add a test to check the generated result please? (Should not have any impact but better to ensure it with a test)

Kijewski:
I added an exhaustive test with all combinations of control characters, but I'm not 100% sure that is what you meant. :-/

GuillaumeGomez:
It's not but it's a good add already. :wink:  What I meant is checking how whitespace suppressors are handled when you `render()` the template.

Kijewski:
Anything outside of `{% block %}` is discarded in extended templates, anyway. For the generator nothing has hanged.

GuillaumeGomez:
And that's exactly what we need to check (if we didn't already).

Kijewski:
I added a test that the same source code is generated, too.

rinja_derive/src/tests.rs
```diff
@@ -135,6 +112,34 @@ struct Foo {{ {} }}"##,
     }
 }
 
+fn jinja_to_rust(jinja: &str, fields: &[(&str, &str)]) -> syn::Result<syn::File> {
```
GuillaumeGomez, line 115: No no, I meant like https://github.com/rinja-rs/rinja/blob/master/testing/tests/simple.rs#L13-L35. Sorry, should have been more clear. No need to check the generated rust code here.

Kijewski: If the same code is generated, then the same code is rendered. It's simpler this way than writing tests with 4 + 4 = 8 cases (all variants at the front and back), or 4 * 4 = 16 tests (cross of all variants).

GuillaumeGomez: I'm not convinced but there are tests, so all good for me.

GuillaumeGomez (approved):

PR #242, opened 2024-11-14
refactor(parser): Remove redundant closure

epage:
This looks like its switching which `i` its working off of but actually isn't


Kijewski (approved):
Thanks!

PR #243, opened 2024-11-14
derive: refactor `#[template]` parsing

Kijewski:
This allows parsing a `#[template]` attribute without expecting all attributes to be set, which makes the parsing re-usable, which is a prerequisit to implement `enum` variant handling.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks! Can't wait for enum support to be there. :D

PR #244, opened 2024-11-14
Allow idents to start with `_`

GuillaumeGomez:
Until now, we didn't allow `_x` as ident. Fixed it. :)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,6 +1,6 @@
 [licenses]
 version = 2
-allow = ["Apache-2.0", "MIT", "MIT-0", "Unicode-DFS-2016"]
+allow = ["Apache-2.0", "MIT", "MIT-0", "Unicode-3.0"]
 private = { ignore = true }
 
 [[licenses.clarify]]
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -997,7 +997,16 @@ impl<'a> Generator<'a> {
                     _ => Ok(false),
                 }
             }
-            Target::Placeholder("_") => Ok(false),
+            Target::Placeholder(_) => Ok(false),
+            Target::Rest(var_name) => {
+                if let Some(var_name) = **var_name {
+                    match self.is_shadowing_variable(ctx, &Target::Name(var_name), l) {
+                        Ok(false) => {}
+                        outcome => return outcome,
+                    }
+                }
+                Ok(false)
+            }
             Target::Tuple(_, targets) => {
                 for target in targets {
                     match self.is_shadowing_variable(ctx, target, l) {
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -997,6 +997,7 @@ impl<'a> Generator<'a> {
                     _ => Ok(false),
                 }
             }
+            Target::Placeholder("_") => Ok(false),
             Target::Tuple(_, targets) => {
                 for target in targets {
                     match self.is_shadowing_variable(ctx, target, l) {
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -2216,7 +2216,7 @@ impl<'a> Generator<'a> {
         target: &Target<'a>,
     ) {
         match target {
-            Target::Placeholder(s) => buf.write(s),
+            Target::Placeholder(_) => buf.write('_'),
             Target::Rest(s) => {
                 if let Some(var_name) = &**s {
                     self.locals
--- a/rinja_parser/src/lib.rs
+++ b/rinja_parser/src/lib.rs
@@ -633,7 +633,12 @@ fn path_or_identifier(i: &str) -> ParseResult<'_, PathOrIdentifier<'_>> {
             path.extend(rest);
             Ok((i, PathOrIdentifier::Path(path)))
         }
-        (None, name, []) if name.chars().next().map_or(true, char::is_lowercase) => {
+        (None, name, [])
+            if name
+                .chars()
+                .next()
+                .map_or(true, |c| c == '_' || c.is_lowercase()) =>
+        {
             Ok((i, PathOrIdentifier::Identifier(name)))
         }
         (None, start, tail) => {
--- a/rinja_parser/src/target.rs
+++ b/rinja_parser/src/target.rs
@@ -7,60 +7,6 @@ use crate::{
     bool_lit, char_lit, identifier, keyword, num_lit, path_or_identifier, str_lit, ws,
 };
 
-fn check_underscore<'a>(i: &'a str, target: &Target<'a>) -> Result<(), ParseErr<'a>> {
-    match target {
-        Target::Name("_") | Target::Placeholder(_) => {
-            return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
-                "reserved keyword `_` cannot be used here",
-                i,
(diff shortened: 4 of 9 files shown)


rinja_parser/src/lib.rs
```diff
@@ -626,7 +626,12 @@ fn path_or_identifier(i: &str) -> ParseResult<'_, PathOrIdentifier<'_>> {
             path.extend(rest);
             Ok((i, PathOrIdentifier::Path(path)))
         }
-        (None, name, []) if name.chars().next().map_or(true, char::is_lowercase) => {
+        (None, name, [])
+            if name
+                .chars()
+                .next()
+                .map_or(true, |c| c == '_' || c.is_lowercase()) =>
```
Kijewski, line 633: I'm kinda confused if it is okay that this function returns `PathOrIdentifier::Identifier("_")` for the placeholder with this change. Before, `PathOrIdentifier::Path(vec!["_"])` would be returned. Actually, both look wrong to me.

Can you have a look what happens of you add a case ↓ before the changed one?

```rust
(None, "_", []) => fail.parse_next(initial_i),
```

Kijewski:
Good catch! Interesting that this case did not occur earlier.

rinja_parser/src/lib.rs
GuillaumeGomez: it didn't change anything as far as I could see.

GuillaumeGomez:
The original patch was not enough so I greatly extended it. It is now much more useful. :)

Kijewski:
Sorry, I'm not quite happy with the second commit. IMHO on the left-hand side of an assignment, the target, placeholders should always be allowed without restrictions. Maybe you are are only interested in the side-effect of an assignment `{% let _ = it.next() %}`, or it must have a specific form `{% let [_] = slice.try_into().unwrap() %}`.

Maybe you could remove the 2nd commit from this PR and move it into a new one, and I'll have a look at it, then?

GuillaumeGomez:
Good point. Wouldn't you prefer that I update the second commit to allow it instead?

Kijewski:
I have no strong opinion. :) We can keep it in one PR.

GuillaumeGomez:
Then leave me some time so I can check what actual cases are allowed on the lhs.

Kijewski:
Take as much time as you need. :)

cipriancraciun:
I think I've opened this can of worms with my initial issue on Askama: <https://github.com/djc/askama/issues/1105>

(I've tested all of the following examples on the current pull request commit <https://github.com/GuillaumeGomez/rinja/commit/05ae1ba808077d7847c64298717a945042ea0073>)

In essence, for variable names I think the following cases should be covered as they are allowed by Rust (and most other programming languages):
* `{% let _ = 42 %}` -- fails with `error: reserved keyword `_` cannot be used here`;  why would someone use this?
  * perhaps for the side-effect;
  * perhaps when refactoring some code where we one doesn't want to change too much;
* `{% let _x = 42 %}` -- my initial issue, which fails with latest Rinja release, works with this pull request;  why would someone use this?
  * for example I always prefix variable names with `_` to highlight that they are locals, and not some crate or something imported;  (I know that in Rust an identifier starting with `_` is treated specially by the compiler, but it works;)
  * any other reason, like for example coming from Python, and creating some templates for Rinja;
  * for example one could use `__`, i.e. two (or more) underscores, as to mean "reserved for base and partials to communicate between themselves", etc., just as the Python convention;
* `{% let X = 42 %}` or `{% let _X = 42 %}` (i.e. uppercase variables) -- fails with `error: literals are not allowed on the left-hand side of an assignment`;  why would someone use this?
  * for example in a base template, or perhaps in an include, someone would like to define a variable that is all upper-case to denote a constant;

Thus, it would be nice before merging this pull-request, to cover a few more corner-cases.

It would be nice (unless it conflicts with the Jinja syntax) that all Rust valid identifiers (as specified in <https://doc.rust-lang.org/reference/identifiers.html>) to be considered valid identifiers for `let`, `for`, `match` and other places where identifiers are to be used.

GuillaumeGomez:
> `{% let X = 42 %}`

For this one, please open another issue. For the rest, I'll add a regression test for them.

GuillaumeGomez:
I re-allowed the few cases that should not be forbidden. :)

rinja_derive/src/generator.rs
```diff
@@ -1118,6 +1118,7 @@ impl<'a> Generator<'a> {
                     _ => Ok(false),
                 }
             }
+            Target::Placeholder("_") => Ok(false),
```
Kijewski, line 1121: Isn't wildcard `..` handling needed, too?
```rust
Target::Rest(var) => match &**var {
  Some(var) => "what's done in Target::Name(_)",
  None => Ok(false),
}
```

GuillaumeGomez: Very good point!

GuillaumeGomez:
Added handling for `..`.

rinja_derive/src/generator.rs
```diff
@@ -997,6 +997,7 @@ impl<'a> Generator<'a> {
                     _ => Ok(false),
                 }
             }
+            Target::Placeholder("_" | "..") => Ok(false),
```
Kijewski, line 1000: The second case will never be true, though. `Target::Placeholder(_)` only ever captures `"_"`, and `Target::Rest(_)` captures `..` or `.. @ var`.

Both variant should probably contain a `WithSpan<'a, ()>` instead of the captured token, to make the meaning more clear.

GuillaumeGomez: Ok, gonna clean up this code because there is no point in keeping content if it's not used.

rinja_parser/src/target.rs
```diff
@@ -7,6 +7,60 @@ use crate::{
...
 };
 
+fn check_underscore<'a>(i: &'a str, target: &Target<'a>) -> Result<(), ParseErr<'a>> {
+    match target {
+        Target::Name("_") | Target::Placeholder(_) => {
+            return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+                "reserved keyword `_` cannot be used here",
+                i,
+            )));
+        }
```
Kijewski, line 17: IMHO, this use of placeholders is fine. Maybe I want to `{% let _ = iter.next() %}` or `{% let [_] = slice.try_into().unwrap() %}`, because I want to test something in my code, generate an effect, whatever. Sorry, but I think whole function can the removed, because I don't want to restrict what can be skipped on the lhs of an assignment. :-/

GuillaumeGomez: Ok, let's go then. :)

GuillaumeGomez:
Ah, we have a new license in our dependency tree apparently.

deny.toml
```diff
@@ -1,6 +1,6 @@
 [licenses]
 version = 2
-allow = ["Apache-2.0", "MIT", "MIT-0", "Unicode-DFS-2016"]
+allow = ["Apache-2.0", "MIT", "MIT-0", "Unicode-3.0"]
```
Kijewski, line 3: Yeah, the license was changed "[about 5 hours ago](https://crates.io/crates/unicode-ident/versions)" :)

GuillaumeGomez: Lucky me. :')

Kijewski (approved):
Thank you very much for the change! :)

PR #247, opened 2024-11-15
Improve error message when an extension escaper is missing

GuillaumeGomez:
Took me way more time than it should have when I saw the original error to understand what was wrong. Fixed it. :)


Kijewski (approved):
Thank you!

Issue #245, opened 2024-11-15
Allow whitespaces around `|` (the filter operator)


cipriancraciun:
Given that #16 was closed, namely disambiguating between `42|safe` (filter) and `42 | 43` (bitwise or), and thus the whitespace surrounding `|` (i.e. `<space>|<space>`) isn't syntactically meaningful anymore, it would be nice to allow users to include spaces between the filter chain.

This would improve readability, especially with long filters, and ones where arguments are used, like for example:
* `fn(42, 43, 44)|filter_1|filter_2` would be more readable as
* `fn(42, 43, 44) | filter_1 | filter_2`

----

As a counterpoint to this proposal, projects that want migrate from Askama to `rinja`:
* if one currently uses `42 | 43` in Askama, when migrating they'll get the "use bitor operator" error;
* meanwhile, if this feature is implemented (i.e. allowing spaces between `|` for filters), then the user would get a strange error regarding `43` not being a valid filter;

Although, given that one is aware of the advantages of switching to `rinja`, and the required changes (i.e. that the bitwise operators are spelled differently), this risk is minimized.

GuillaumeGomez:
We don't allow whitespace in order to follow jinja's spec to make it compatible (as much as possible) with the official jinja implementation.

Normally, `rinja` provides a good error message to help you understand what's wrong.

Personally, I'd be more inclined to keep things as is.

cipriancraciun:
> We don't allow whitespace in order to follow jinja's spec to make it compatible (as much as possible) with the official jinja implementation.

I've tried to locate the Jinja specification (I searched on Google a little, but found only the Jinja Python documentation, thus the "original" implementation), and the following document is full of examples like these:
=> <https://jinja.palletsprojects.com/en/stable/templates/>
~~~~
{% for key, value in my_dict | dictsort %}
{% set reply | wordwrap %}
{{ data | selectattr('name', '==', 'Jinja') | list | last }}
{{ value | markdown }}
~~~~

Thus, the Python Jinja documentation makes me believe that spaces are allowed between the `|` segments.

----

> Normally, `rinja` provides a good error message to help you understand what's wrong.

Indeed, however, the choice of whitespaces is more a matter of taste, some people (myself included) likes to put whitespaces between things.  :)

Moreover, from a programming language point of view, having whitespaces (that aren't tied to indentation) be syntactically meaningful is a "strange" choice.  :) :)

GuillaumeGomez:
[...]

PR #250, opened 2024-11-17
Remove restriction preventing to have | surrounded by whistespace characters

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/245.

Just checked and jinja2 works when filter operators are surrounded with whitespace characters. From this point, there doesn't seem to be any reason for `rinja` to not allow it as well.


GuillaumeGomez:
The book was still wrong about why whitespace characters were needed. ^^'

rinja_parser/src/lib.rs
```diff
@@ -899,20 +899,10 @@ fn filter<'a>(
     i: &'a str,
     level: &mut Level,
 ) -> ParseResult<'a, (&'a str, Option<Vec<WithSpan<'a, Expr<'a>>>>)> {
-    let (j, _) = skip_ws0.parse_next(i)?;
-    let had_spaces = i.len() != j.len();
-    let (j, _) = ('|', not('|')).parse_next(j)?;
+    let (j, _) = (ws('|'), not('|')).parse_next(i)?;
```
Kijewski, line 902:
```suggestion
    let (j, _) = ws(('|', not('|'))).parse_next(i)?;
```

GuillaumeGomez: 👍

Kijewski (approved):
Yeah, it's easier to read with spaces around the `|`.

GuillaumeGomez:
Agreed. :)

PR #252, opened 2024-11-18
Remove `EXTENSION` constant from `Template`

Kijewski:
I doubt that is needed in any circumstance, and it makes implementing enum variants harder.


GuillaumeGomez (approved):

GuillaumeGomez:
Less code is good. :)

Issue #223, opened 2024-11-05
Add enum support


GuillaumeGomez:
It would allow to do:

```rust
#[derive(Template)]
#[derive(Serialize, Deserialize)]
pub enum QuestionData {
	#[template(path = "./templates/question/editor/multiple-text.html")]
	MultipleChoiceText {
		numbering: Numbering,
		text: Vec<String>,
	},

	#[template(path = "./templates/question/editor/multiple-image.html")]
	MultipleChoiceImage {
		numbering: Numbering,
		images: Vec<i32>,
	},
}
```

And so calling `render` on the enum would be on its variant instead. It was suggested in https://github.com/djc/askama/issues/1104 and I kinda like the idea.

darrenmothersele:
I agree, this would be useful. Something like [this](https://docs.rs/askama-enum/latest/askama_enum/), perhaps?

GuillaumeGomez:
With a default template? Hum why not!

Kijewski:
I think it would a good feature to have!

The main problem I have are the constants in [`Template`](https://docs.rs/rinja/0.3.5/rinja/trait.Template.html). For the `SIZE_HINT` we could simply use the median, I guess. But `EXTENSION` and `MIME_TYPE` might be off, e.g. if one enum variant should generate a plaintext error message.

I think, in a first iteration we could require that the `ext` parameter is the same for all variants. Later on, we could explore how to lift this requirement. If you too think that's a good idea, then I'll have a look how best to salvage [askama-enum](https://github.com/Kijewski/askama-enum/blob/main/src/lib.rs). I think most of the code can be copied.

GuillaumeGomez:
Not sure if median is the best idea here. Maybe taking the biggest value would be better?

chrisp60:
As a user, I would expect the `HINT_SIZE` to derived from the largest variant. This would mimic the rust sizing of enums. Just an opinion. 

Maybe an attribute can be accepted when deriving enums that lets you specify an escape hatch. Unsure if it is worth the maintenance cost for how often it would actually get used. I imagine most users would just take the default and not worry about it.

```rust
#[derive(Template)]
#[derive(Serialize, Deserialize)]
#[template(size_hint = "median")]
// #[template(size_hint = "largest")] default
// #[template(size_hint = Self::MultipleChoiceText)] a "hot path" variant
pub enum QuestionData {
	#[template(path = "./templates/question/editor/multiple-text.html")]
	MultipleChoiceText {
[...]

PR #255, opened 2024-11-19
Implement `enum` variants

Kijewski:
Resolves <https://github.com/rinja-rs/rinja/issues/223>.

### DIFF
--- a/book/src/creating_templates.md
+++ b/book/src/creating_templates.md
@@ -118,6 +118,85 @@ recognized:
   struct HelloTemplate<'a> { ... }
   ```
 
+## Templating `enum`s
+
+You can add derive `Template`s for `struct`s and `enum`s.
+If you add `#[template()]` only to the item itself, both item kinds work exactly the same.
+But with `enum`s you also have the option to add a specialized implementation to one, some,
+or all variants:
+
+```rust
+#[derive(Debug, Template)]
+#[template(path = "area.txt")]
+enum Area {
+    Square(f32),
+    Rectangle { a: f32, b: f32 },
+    Circle { radius: f32 },
+}
+```
+
+```jinja2
+{%- match self -%}
+    {%- when Self::Square(side) -%}
+        {{side}}^2
+    {%- when Self::Rectangle { a, b} -%}
+        {{a}} * {{b}}
+    {%- when Self::Circle { radius } -%}
+        pi * {{radius}}^2
+{%- endmatch -%}
+```
+
+will give you the same results as:
+
+```rust
+#[derive(Template, Debug)]
+#[template(ext = "txt")]
+enum AreaPerVariant {
+    #[template(source = "{{self.0}}^2")]
+    Square(f32),
+    #[template(source = "{{a}} * {{b}}")]
+    Rectangle { a: f32, b: f32 },
+    #[template(source = "pi * {{radius}}^2")]
+    Circle { radius: f32 },
+}
+```
+
+As you can see with the `ext` attribute, `enum` variants inherit most settings of the `enum`:
+`config`, `escape`, `ext`, `syntax`, and `whitespace`.
+Not inherited are: `block`, and `print`.
+
+If there is no `#[template]` annotation for an `enum` variant,
+then the `enum` needs a default implementation, which will be used if `self` is this variant.
+A good compromise between annotating only the template, or all its variants,
+might be using the `block` argument on the members:
+
+```rust
+#[derive(Template, Debug)]
+#[template(path = "area.txt")]
+enum AreaWithBlocks {
+    #[template(block = "square")]
+    Square(f32),
+    #[template(block = "rectangle")]
+    Rectangle { a: f32, b: f32 },
+    #[template(block = "circle")]
+    Circle { radius: f32 },
+}
+```
+
+```jinja2
+{%- block square -%}
+    {{self.0}}^2
+{%- endblock -%}
+
+{%- block rectangle -%}
+    {{a}} * {{b}}
+{%- endblock -%}
+
+{%- block circle -%}
+    pi * {{radius}}^2
+{%- endblock -%}
+```
+
 ## Documentation as template code
 [#documentation-as-template-code]: #documentation-as-template-code
 
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -25,9 +25,12 @@ pub(crate) fn template_to_string(
     input: &TemplateInput<'_>,
     contexts: &HashMap<&Arc<Path>, Context<'_>, FxBuildHasher>,
     heritage: Option<&Heritage<'_, '_>>,
-    target: Option<&str>,
+    tmpl_kind: TmplKind,
 ) -> Result<usize, CompileError> {
-    let ctx = &contexts[&input.path];
+    if tmpl_kind == TmplKind::Struct {
+        buf.write("const _: () = { extern crate rinja as rinja;");
+    }
+
     let generator = Generator::new(
         input,
         contexts,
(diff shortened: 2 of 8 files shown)


rinja_derive/src/input.rs
```diff
@@ -273,6 +273,69 @@
...
+
+impl AnyTemplateArgs {
+    pub(crate) fn new(ast: &syn::DeriveInput) -> Result<Self, CompileError> {
+        let syn::Data::Enum(enum_data) = &ast.data else {
+            return Ok(Self::Item(TemplateArgs::new(ast)?));
+        };
+
+        let enum_args = PartialTemplateArgs::new(ast, &ast.attrs)?;
+        let vars_args = enum_data
+            .variants
```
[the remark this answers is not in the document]
Kijewski: There is no comment??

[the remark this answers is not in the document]
GuillaumeGomez: Oh that's a funny one. XD

rinja_derive/src/input.rs
```diff
@@ -271,6 +271,69 @@ impl TemplateInput<'_> {
     }
 }
 
+pub(crate) enum AnyTemplateArgs {
+    Item(TemplateArgs),
```
GuillaumeGomez, line 275: Might be better to name it `Struct` since we don't support unions currently.

rinja_derive/src/integration.rs
```diff
@@ -223,3 +232,265 @@ fn string_escape(dest: &mut String, src: &str) {
...
+    ));
+    Ok(biggest_size_hint)
+}
+
+fn set_default<S, T, A>(dest: &mut S, parent: &mut S, mut access: A)
+where
+    T: Clone,
+    A: FnMut(&mut S) -> &mut Option<T>,
+{
+    if let dest @ None = access(dest) {
```
GuillaumeGomez, line 358: It's super weird. What about:

```suggestion
    if access(dest).is_none() {
```

?

GuillaumeGomez:
So if I understand correctly, when a variant has a `#[template()]` attribute, we generate a struct representing it then call `render`/`render_into()` on this struct, right?

Very hypothetical (and in case I understood everything correctly up to this point): isn't it possible to instead handle the variant like we handle our structs. Trying to explain in code:

```rust
#[template(src = "boo.txt")]
enum Foo {
    A,
    #[template(src = "I-am-b.txt")]
    B(u32, u32),
    #[template(src = "not-boo.txt")]
    C { a: u32 },
}

impl Render for Foo {
    fn render(&self) -> Result<String> {
        match self {
            // First the variants with the template
            B(_0, _1) => {
                // this one needs a struct binding :'(
            }
            C { a } => {
                // We can re-use the code for structs where we can say that this is a struct with a field named `a` and tada.
            }
        }
    }
}
```

Not sure if it's a good idea though... That would allow to skip a type creation for one case. But then `self.a` wouldn't work. So I think overall I just had a terrible idea. ^^'

rinja_derive/src/integration.rs
Kijewski: It's the same as `let dest = access(dest); if dest.is_none() {` in one line. I found it nifty. :) But changed.

Kijewski:
Yep, I implemented it this way so that `self.x` still works. I'd let the compile/optimizer decide if it's good idea to inline everything into one big function or if the implementations for every variant should stay.

GuillaumeGomez:
Please tell me when it's ready for final review. ;)

(also git conflict)

book/src/creating_templates.md
```diff
@@ -118,6 +118,85 @@ recognized:
   struct HelloTemplate<'a> { ... }
   ```
 
+## Templating `enum`s
+
+You can add derive `Template`s for `struct`s and `enum`s.
+If you add `#[template()]` only to the item itself, both item kinds work exactly the same.
+But with `enum`s you also have the option add a specialized implementation to one, some,
```
GuillaumeGomez, line 125:
```suggestion
But with `enum`s you also have the option to add a specialized implementation to one, some,
```

book/src/creating_templates.md
```diff
@@ -118,6 +118,85 @@ recognized:
...
+    #[template(source = "pi * {{radius}}^2")]
+    Circle { radius: f32 },
+}
+```
+
+As you can see with the `ext` attribute, `enum` variants inherit most settings of the `enum`:
+`config`, `escape`, `ext`, `syntax`, and `whitespace`.
+Not inherited are: `block`, and `print`.
+
+If there is no `#[template]` annotation for a `enum` variant,
```
GuillaumeGomez, line 168:
```suggestion
If there is no `#[template]` annotation for an `enum` variant,
```

book/src/creating_templates.md
```diff
@@ -118,6 +118,85 @@ recognized:
...
+then the `enum` needs a default implementation, which will be used if `self` is this variants.
```
GuillaumeGomez, line 169:
```suggestion
then the `enum` needs a default implementation, which will be used if `self` is this variant.
```

rinja_derive/src/generator.rs
```diff
@@ -25,9 +25,12 @@ pub(crate) fn template_to_string(
     input: &TemplateInput<'_>,
     contexts: &HashMap<&Arc<Path>, Context<'_>, FxBuildHasher>,
     heritage: Option<&Heritage<'_, '_>>,
-    target: Option<&str>,
+    tmpl_kind: TmplKind,
 ) -> Result<usize, CompileError> {
-    let ctx = &contexts[&input.path];
+    if tmpl_kind == TmplKind::Struct {
+        buf.write("const _: () = { extern crate rinja as rinja;");
```
GuillaumeGomez, line 31: Is the `as rinja` part useful?

rinja_derive/src/generator.rs
```diff
@@ -161,12 +165,12 @@ impl<'a, 'h> Generator<'a, 'h> {
...
-        buf.write(format_args!(
-            "\
-                rinja::Result::Ok(())\
-            }}\
-            const SIZE_HINT: rinja::helpers::core::primitive::usize = {size_hint}usize;",
-        ));
+        buf.write("rinja::Result::Ok(()) }");
+        if tmpl_kind == TmplKind::Struct {
+            buf.write(format_args!(
+                "const SIZE_HINT: rinja::helpers::core::primitive::usize = {size_hint}usize;",
```
GuillaumeGomez, line 171: Seems a bit strange that the number is marked as usize twice. Why not:

```suggestion
                "const SIZE_HINT: rinja::helpers::core::primitive::usize = {size_hint};",
```

?

rinja_derive/src/integration.rs
```diff
@@ -223,3 +232,291 @@ fn string_escape(dest: &mut String, src: &str) {
...
+    enum_ast: &DeriveInput,
+    mut enum_args: Option<PartialTemplateArgs>,
+    vars_args: Vec<Option<PartialTemplateArgs>>,
+    has_default_impl: bool,
+) -> Result<usize, CompileError> {
+    let Data::Enum(enum_data) = &enum_ast.data else {
+        unreachable!();
+    };
+
+    buf.write("const _: () = { extern crate rinja as rinja;");
```
GuillaumeGomez, line 247: Same question here about the `as rinja` part.

rinja_derive/src/integration.rs
```diff
@@ -223,3 +232,291 @@ fn string_escape(dest: &mut String, src: &str) {
...
+            let mut buf = rinja::helpers::alloc::string::String::new();\
+            let _ = buf.try_reserve(size_hint);\
+            self.render_into_with_values(&mut buf, __rinja_values)?;\
+            rinja::Result::Ok(buf)\
+        }}",
+    ));
+
+    buf.write(format_args!(
+        "\
+        const SIZE_HINT: rinja::helpers::core::primitive::usize = {biggest_size_hint}usize;\
```
GuillaumeGomez, line 362: Same about having `usize` twice.

GuillaumeGomez:
Apart from the small nits, it looks good to me, great work! Can't wait for this to be merged. :3

rinja_derive/src/generator.rs
Kijewski: The same line `extern crate rinja as rinja;` already exists at other places outside of this PR's changes.

Right now it's a placeholder. I plan to add a `crate_` argument to `#[template]`. That would make it easier to derive Templates in a macro.

rinja_derive/src/generator.rs
Kijewski: Consistency: `quote!`'d numbers append the type. E.g. the parser tests use `quote!` and would fail otherwise.

Kijewski:
Thank you for the proofreading! Fixed the typos.

rinja_derive/src/generator.rs
GuillaumeGomez: Makes sense. :+1

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the replies. Still not sure to understand the use case of `extern as` but not a blocker so let's merge. Thanks a lot for the PR!

PR #257, opened 2024-11-19
parser: shrink the size of `WithSpan` to one register

Kijewski:
To be rebased after <https://github.com/rinja-rs/rinja/pull/214> is merged.


rinja_parser/src/lib.rs
```diff
@@ -134,19 +134,70 @@ impl<'a> Ast<'a> {
 /// in the code generation.
 pub struct WithSpan<'a, T> {
     inner: T,
-    span: &'a str,
+    span: Span<'a>,
+}
```
GuillaumeGomez, line 138: To ensure the size doesn't grow up, might be worth adding a compile-time `size_of` check of `WithSpan` and of `Span`.

Kijewski:
Oh, I did not expect a fuzzer error, but it makes me happy that it is actually working!

> `Offset::offset_to` only accepts slices of `self`

Oops, I should have read the documentation of the method more properly. :)

rinja_parser/src/lib.rs
Kijewski: I don't think it matters too much, as the PR fixes another problem, too:

If the input span is shrunk at the end of the input, e.g. `let i = i.trim()`, then the error messages will be broken afterward, because only the lengths are compared. This PR fixes that, too. This was actually my original intention with, because I had to refactor some code a few times in order not to break the error messages. That the Span is smaller afterwards is just a happy co-incidence. :)

Kijewski:
Well, the changes in the parser are not too big. No reason to wait.

GuillaumeGomez (approved):

GuillaumeGomez:
You even added the fuzzing error, nice!

> Well, the changes in the parser are not too big.

As the reviewer, I disagree. :p

GuillaumeGomez:
Please merge once fuzzer passed.

PR #258, opened 2024-11-19
Canonicalize include paths before emitting

Kijewski:
This allows `CARGO_MANIFEST_DIR` to be a relative path, which can be useful in non-cargo build systems.

---

Cherry-picked from <https://github.com/djc/askama/pull/1107>.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -172,11 +172,10 @@ impl<'a> Generator<'a> {
                 Source::Source(_) => path != &*self.input.path,
             };
             if path_is_valid {
-                let canonical_path = path.canonicalize().unwrap();
-                let include_path = canonical_path.to_str().unwrap();
                 buf.write(format_args!(
                     "const _: &[rinja::helpers::core::primitive::u8] =\
-                        rinja::helpers::core::include_bytes!({include_path:#?});",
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (dismissed):

Kijewski:
Drawback: Before this PR, the crate root could reside in a non-UTF-8 path without issues, any only everything inside it had to have a sane encoding. With the PR, the whole path has to be sane.

But I guess that's okay. As a user, I don't think I would expect things to be working it my path was broken.

rinja_derive/src/generator.rs
```diff
@@ -172,10 +172,11 @@ impl<'a> Generator<'a> {
                 Source::Source(_) => path != &*self.input.path,
             };
             if path_is_valid {
-                let path = path.to_str().unwrap();
+                let canonical_path = path.canonicalize().unwrap();
+                let include_path = canonical_path.to_str().unwrap();
```
GuillaumeGomez, line 176: Do we actually need to convert it to string? Can't we just use its `Debug` impl?

Kijewski: Good idea! No, it seems to work just fine without it. And instead of panicking if the path could not be resolved, we can simply use the relative path. I think that's quite elegant. :)

GuillaumeGomez (approved):

PR #260, opened 2024-11-20
derive: some clean-ups and optimizations

Kijewski:
Minor changes extracted from the `enum` implementation. Please see the individual commits.

### DIFF
--- a/rinja_derive/src/input.rs
+++ b/rinja_derive/src/input.rs
@@ -7,6 +7,7 @@ use std::str::FromStr;
 use std::sync::{Arc, OnceLock};
 
 use mime::Mime;
+use parser::node::Whitespace;
 use parser::{Node, Parsed};
 use proc_macro2::Span;
 use rustc_hash::FxBuildHasher;
@@ -282,7 +283,7 @@ pub(crate) struct TemplateArgs {
     ext_span: Option<Span>,
     syntax: Option<String>,
     config: Option<String>,
-    pub(crate) whitespace: Option<String>,
+    pub(crate) whitespace: Option<Whitespace>,
     pub(crate) template_span: Option<Span>,
     pub(crate) config_span: Option<Span>,
 }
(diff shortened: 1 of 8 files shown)


rinja_derive/src/input.rs
```diff
@@ -307,15 +307,15 @@ impl TemplateArgs {
...
                 }
-                Some((_, PartialTemplateArgsSource::Source(s))) => {
+                Some(PartialTemplateArgsSource::Source(s)) => {
                     (Source::Source(s.value().into()), Some(s.span()))
                 }
                 #[cfg(feature = "code-in-doc")]
-                Some((ident, PartialTemplateArgsSource::InDoc(_))) => {
-                    source_from_docs(&ident, &args.meta_docs, ast)?
+                Some(PartialTemplateArgsSource::InDoc(s)) => {
+                    source_from_docs(s.span(), &args.meta_docs, ast)?
```
GuillaumeGomez, line 319: Nice find!

testing/tests/ui/rinja-block.stderr
```diff
@@ -7,10 +7,10 @@ error: unknown node `fail`
   | ^^^^^^^^^^^^^^^^^^^^^^
 
 error: when using `in_doc = true`, the struct's documentation needs a `rinja` code block
-  --> tests/ui/rinja-block.rs:15:25
+  --> tests/ui/rinja-block.rs:15:34
    |
 15 | #[template(ext = "txt", in_doc = true)]
```
GuillaumeGomez, line 12: Shouldn't it underline the whole `in_doc = true`?

GuillaumeGomez (dismissed):

GuillaumeGomez:
Apart from my one comment, looks all good to me, thanks! Please feel free to merge even without fixing my comment (but if you do, please open an issue so we don't forget about it and I'd love to fix it 😄 ).

testing/tests/ui/rinja-block.stderr
Kijewski: In the other cases we also only underline the value part, the part after the equals sign. I actually came to like that, because the problem is with the value, not with the key. If you write `in_doc = false`, then everything is fine (except that in here there is not source).

GuillaumeGomez: Hmmmmmmm. Then the message needs to be reworded a bit to make it obvious that the value is the value. Like:

> when using `in_doc` with the `true` value, the struct's documentation needs a `rinja` code block

Or something like that.

Kijewski: I changed the message a bit. Do you think it's more clear now?

GuillaumeGomez: Indeed, thanks!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for the improvements!

PR #262, opened 2024-11-21
fix: normalize macro args on call (fixes #259)

### LINKED ISSUE #259: macros fail on invocation when arguments are rust keywords
**I am using git main**

I can define this macro without issue

```handlebars
{% macro foo(type="text") %}
    <input type="{{ type }}"/>
{% endmacro foo %}

{# or this one #}

{% macro foo(type) %}
    <input type="{{ type }}"/>
{% endmacro foo %}
```
But calling them produces an error (error shown below).
```handlebars
{% call foo("input") %}
{% call foo(type="input") %}
```

```
error: expected identifier, found keyword `type`
  --> [..path]
   |
30 | #[derive(Template, Debug)]
   |          ^^^^^^^^ expected identifier, found keyword
   |
   = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)

error: proc-macro derive produced unparsable tokens
  --> [..path]
   |
30 | #[derive(Template, Debug)]
   |          ^^^^^^^^

error[E0609]: no field `r#type` on type `&[Struct with Template derived]`
  --> [..path]
   |
30 | #[derive(Template, Debug)]
   |          ^^^^^^^^ unknown field
   |
   = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)

```

Did not see any issues created in regards to this, or mention in the docs that it should be avoided. Just wanted to point it out in case it does not align with intentions.

As a user, I feel the error should at least come from the definition, and not the invocation. It does seem like the keyword is properly escaped in at least some capacity (the error references `r#type`).

Current workaround, naming the argument something else

chrisp60:
I believe this should resolve #259.

Is it correct that `this.locals` should store only normalized identifiers, since there is some assumption that they must be written (and thus valid rust)?

Please feel free to correct me if that is not the case. 

Edit: there is also a failing ui test that seems unrelated. Can dig into it deeper if needed, but must return to the day job for now

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -848,8 +848,11 @@ impl<'a> Generator<'a> {
                             ("", "")
                         };
                         value.write(this.visit_expr_root(ctx, expr)?);
-                        buf.write(format_args!("let {arg} = {before}{value}{after};"));
-                        this.locals.insert_with_default(Cow::Borrowed(arg));
+                        // We need to normalize the arg to write it, thus we need to add it to
+                        // locals in the normalized manner
+                        let normalized_arg = normalize_identifier(arg);
+                        buf.write(format_args!("let {} = {before}{value}{after};", normalized_arg));
+                        this.locals.insert_with_default(Cow::Borrowed(normalized_arg));
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks, nicely done!

PR #264, opened 2024-11-22
Allow variables inheritance in block

### LINKED ISSUE #246: Variables defined in base template, fail to be accessed in partials included by extended templates
Given the following example:
~~~~
## base.txt
{% let variable = 42 %}
{% block extended %}{% endblock %}
~~~~

~~~~
## extended.txt
{% extends "base.txt" %}
{% block extended %}
WORKS for extended: {{ variable }}
{% include "partial.txt" %}
{% endblock %}
~~~~

~~~~
## partial.txt
FAILS for extended: {{ variable }}
WORKS for includes: {{ variable }}
~~~~

~~~~
## includes.txt
{% let variable = 42 %}
WORKS for includes: {{ variable }}
{% include "partial.txt" %}
~~~~

The `extended.txt` template fails for Rinja with the following error, meanwhile succeeds with Askama (i.e. Rinja regression when migrating from Askama):
~~~~
error[E0609]: no field `variable` on type `&RinjaTestTemplate`
~~~~

Meanwhile, `includes.txt` works for both Askama and Rinja.

----

Even more strangely, if the `partial.txt` is an empty file, thus containing no reference to the variable, just by including it in `extended.txt` triggers the error.

Meanwhile, if the inclusion is missing, `extended.txt` compiles correctly.

----

For some reason the Rinja parser / generator behaves differently for variables defined in a base template, than variables defined in non-inherited templates, but only when that variable is used in partials (i.e. `{% include "partial.txt" %}`).

Thus, at least the scoping rules seem to be applied ambiguously.

Moreover, the current Rinja documentation supports the case, by implying that `{% include ... %}` statements should behave as if they wouldn't be there, and the contents of the included template would be spliced at the place of inclusion:

> The include statement lets you split large or repetitive blocks into separate template files. Included templates get full access to the context in which they're used, including local variables like those from loops:

----

Where would such a use-case be needed:
* one wants to create multiple pages, that each share a common outer layout, thus one creates a `base` template;
* the `base` template is then extended for each `extended` pages;
* the `extended` pages would like to include some common `widget` that are developed as includes (i.e. the `partial.txt` template in this issue);
* the `base` and `widget` "communicate" based on variables, but don't want to involve the `extended` pages, thus the developer defines some variables inside `base` that are to be used in `widget` templates;

At the moment, for the given use-case, I see no workaround that doesn't involve moving and duplicating the variables both in `base` and `extended`.

GuillaumeGomez:
Fixes #246.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -3,7 +3,7 @@ use std::collections::hash_map::{Entry, HashMap};
 use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
-use std::{cmp, hash, mem, str};
+use std::{mem, str};
 
 use parser::node::{
     Call, Comment, Cond, CondTest, FilterBlock, If, Include, Let, Lit, Loop, Macro, Match,
@@ -65,7 +65,7 @@ struct Generator<'a> {
     // The heritage contains references to blocks and their ancestry
     heritage: Option<&'a Heritage<'a>>,
     // Variables accessible directly from the current scope (not redirected to context)
-    locals: MapChain<'a, Cow<'a, str>, LocalMeta>,
+    locals: MapChain<'a>,
     // Suffix whitespace from the previous literal. Will be flushed to the
     // output buffer unless suppressed by whitespace suppression on the next
     // non-literal.
@@ -86,7 +86,7 @@ impl<'a> Generator<'a> {
         input: &'n TemplateInput<'_>,
         contexts: &'n HashMap<&'n Arc<Path>, Context<'n>, FxBuildHasher>,
         heritage: Option<&'n Heritage<'_>>,
-        locals: MapChain<'n, Cow<'n, str>, LocalMeta>,
+        locals: MapChain<'n>,
         buf_writable_discard: bool,
         is_in_filter_block: usize,
     ) -> Generator<'n> {
@@ -304,8 +304,7 @@ impl<'a> Generator<'a> {
     }
 
     fn is_var_defined(&self, var_name: &str) -> bool {
-        self.locals.get(&var_name.into()).is_some()
-            || self.input.fields.iter().any(|f| f == var_name)
+        self.locals.get(var_name).is_some() || self.input.fields.iter().any(|f| f == var_name)
     }
 
     fn evaluate_condition(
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
The original discussion was in https://github.com/djc/askama/issues/989. I don't really see a reason why we didn't allow variables to be inherited. Well, fixed now.

rinja_derive/src/generator.rs
```diff
@@ -1152,16 +1152,28 @@ impl<'a> Generator<'a> {
...
-            MapChain::default(),
+            // FIXME: Fix the broken lifetimes.
+            //
+            // `transmute` is here to fix the lifetime nightmare around `&self.locals` not
+            // living long enough. The correct solution would be to make `MapChain` take two
+            // lifetimes `&'a MapChain<'a, 'b, K, V>`, making the `parent` field take
+            // `<'b, 'b, ...>`. Except... it doesn't work because `self` still doesn't live long
+            // enough here for some reason...
+            MapChain::with_parent(unsafe {
+                mem::transmute::<
```
GuillaumeGomez, line 1168: I'm still very unhappy about this. I was able to improve the situation a lot but still get this last error:

```
error: lifetime may not live long enough
    --> rinja_derive/src/generator.rs:1166:13
     |
84   | impl<'a, 'b> Generator<'a, 'b> {
     |      --  -- lifetime `'b` defined here
     |      |
     |      lifetime `'a` defined here
...
1166 |             MapChain::with_parent(&self.locals),
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ argument requires that `'b` must outlive `'a`
     |
     = help: consider adding the following bound: `'b: 'a`
```

I pushed my changes [here](https://github.com/rinja-rs/rinja/compare/master...GuillaumeGomez:rinja:reduce-lifetime-error?expand=1).

I'll try to continue to understand what's wrong. If anyone wants to give it a try though...

Kijewski:
Please have a look at my change whether it's okay for you.

<sup>I won't admit that it took me like 6 hours in today until I came up with a solution I liked.</sup>

GuillaumeGomez:
So `Heritage` was the issue! It was one of my next targets for finding out what went wrong. Less generics is a very good idea!

Book build is failing for some reason though? Seems unrelated to this PR.

GuillaumeGomez:
(also I cannot approve my own PR haha)

Kijewski (approved):
:+1:

PR #266, opened 2024-11-24
Second round of optimizations extracted from PR to implement `enum` variants

Kijewski:


rinja_derive/src/input.rs
```diff
@@ -547,9 +547,9 @@ fn extension(path: &Path) -> Option<&str> {
     }
 }
 
-#[derive(Debug, Hash, PartialEq)]
+#[derive(Debug, Clone, Hash, PartialEq)]
 pub(crate) enum Source {
-    Path(String),
+    Path(Arc<str>),
```
GuillaumeGomez, line 548: Why an `Arc` and not an `Rc` btw? Do we need `Source` to be `Send`?

Kijewski: The other variant `Source::Source` needs an `Arc` to be compatible with cached file sources. `Source::Path` does not, but I think making both variants the same is still the better call. The Clone and Drop implementations are half the size, which I daresay outperforms any drawbacks the use of atomic numbers bring.

GuillaumeGomez: Right, forgot about the static. That answers my question. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #268, opened 2024-11-24
derive standalone: opt out of native `proc-macro` support

Kijewski:
We know that there is none, because `rinja_derive_standalone` is not run inside of rust, but, as the name says, standalone. This makes the playground WASM data a little smaller.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice improvement, thanks!

Issue #267, opened 2024-11-24
match blocks generate code that might normally not compile, leading to use of variable bindings named like enum variants


freelon:
Hi everybody,

I used the `{% match xxx %}` syntax recently and ended up in a nasty debugging session because the derived Template implementation is not checked against the `bindings_with_variant_name` lint (see the example below). I had not qualified my enum variants in the template leading to always the first match expression evaluation to true.

As far as I could find out this cannot easily be changed (as a macro-derive lib I cannot specify that my generated code should be checked), but maybe a warning in the docs to always qualify variants might be helpful for others. Especially, because the initial example with an `Option` type does not use qualifiers, which is not necessary because the `Option` variants are implicitly imported.

Example code:
```
#[derive(Template)]
#[template(path = "hello.html")]
struct Foo {
    bar: Bar,
}

enum Bar {
    Hello,
    World,
}
```
```
{% match bar %}
{% when Hello %}
hello
{% when World %}
world
{% endmatch %}
```
Which expands to roughly the following:
```
    match &self.bar {
        Hello => writer.write_str("hello"),
        World => writer.write_str("world"),
    }
```
which is by default rejected by the compiler if written like that in a source file, because `Hello` and `World` are seen as variables and not enum variants, thus always evaluating to `true`.

I'm using version `0.3.5` and the current stable docs (same with the docs on the `master` branch).

Long story short, unless there is a way to enforce the lint in the generated code, maybe warn in the docs (I can provide a PR if you agree).

Kijewski:
The code that is generated by proc-macros is expected to be "sane", and to some lints we have just no way to opt in to, like `bindings_with_variant_name`.

Please have a look at <https://github.com/rinja-rs/rinja/pull/269>. It makes your example a compile error, which is at least better than wrong matches. The error message is not perfect, and the cannot change it, but should guide the user to match on `Foo::Bar` instead of just `Bar`.

freelon:
Nice, looks good to me. I also thing a compile error is the best way to go. And the compiler message is helpful enough.

Thanks for the quick response!

PR #269, opened 2024-11-24
derive: a safe-guard to disable bindings a unit struct was expected

Kijewski:
This does not fix the error, as rinja is not able to tell was the actually intended to do, but rust gives you a guide what the error is, and the error is not silently ignored.

Related issue: <https://github.com/rinja-rs/rinja/issues/267>.

### DIFF
--- a/rinja_derive/src/generator.rs
+++ b/rinja_derive/src/generator.rs
@@ -2311,6 +2311,7 @@ impl<'a, 'h> Generator<'a, 'h> {
             }
             Target::Path(path) => {
                 self.visit_path(buf, path);
+                buf.write("{}");
             }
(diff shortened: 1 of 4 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
I knew that you could initialize a unit struct with `{}` but I didn't know we could match on unit variants like this as well. Nice discovery, thanks!

Kijewski:
Yeah, for tuple structs you can even do this: `let Tpl { 0: a, 1: b } = Tpl(1, 2);`. Makes makes things easier, and I even use it in the enum variant implementation. :)

PR #271, opened 2024-11-25
Remove integrations and `MIME_TYPE` guessing

### LINKED ISSUE #263: Making "mime magic" optional?
Thank you for creating rinja, it seems quite promising and I'm considering switching to it right now!

I like to keep my dependency tree as slim as possible, so it's great that things like `config` are optional features and help reduce the number of dependencies.

I was wondering about the extension-based mime detection though. I don't really think I need it. The last line in my axum request handlers is something like `Ok(Html(template.render()?))` and that seems completely fine. I don't think I'll need the MIME detection or axum integration that seems to make use of it.

If I could drop the `mime`, `mime_guess` and `unicase` dependencies in exchange for not having that MIME detection feature, I'd consider that a good deal.

What do you think about making all the MIME-related things optional? Would you consider a PR or maybe it would be too much of a `#[cfg]` mess?

Another approach might be dropping the MIME detection altogether, changing MIME_TYPE in the Template trait to `Option<&'static str>` and set it to `None` by default. The integrations would use "text/html; charset=utf-8" in that case and there'd be a `#[template(mime="image/svg+xml")]` param for those who serve SVG content from templates and need the MIME_TYPE to be set for integrations to work correctly. This seems like a painful migration for those who rely on the current auto-detect behavior though, as it might not be obvious that the behavior is gone and they need to add the param or risk having their application break because SVG files are being served with HTML content-type.

Or maybe – and this seems quite brazen, feel free to ignore – drop the integrations along with mime handling? Looking at download counts, the overwhelming number of users seem to use rinja without any integrations. The crate itself has 50k downloads a month, while integrations only have a few hundred max, according to lib.rs. Like 300 for the axum integration.

Also, I won't be mad if you read this and say "nope, we like that feature, closing". :-) I know not everyone dislikes "unimportant" dependencies as much as I do, especially in proc-macros!

Kijewski:
Resolves #263.

Ping @mo8it.


GuillaumeGomez (approved):

GuillaumeGomez:
So much code removed. :3

PR #275, opened 2024-11-25
refactor(parser): Be explicit about `i`nput state

epage:
Instead of parsers returning a `j` or `new_i` and conditionally chaining back into `i` later, this has parsers always return `i`, requiring explicit updating of `i`.  The one time `j` is still used to just because we can't mix a `let` and a mutation of an existing variable.

This is needed for Winnow 0.5 as `i` changes from `Fn(&str) -> (I, O)` to `Fn(&mut &str) -> O`, requiring the explicit updating that this PR does.

imo this also improves the visibility of control flow and so I feel like this is justifiable on its own, in addition to reducing my merge conflicts. I'll likely need to do more of this as I continue my 0.5 work or as other people make changes to the parser.


Kijewski (approved):
Looks good, thank you!

PR #276, opened 2024-11-26
derive: less `format!()`, shorter context on error

Kijewski:


rinja_derive/src/lib.rs
```diff
@@ -244,28 +245,13 @@ struct CompileError {
...
-    ) -> Self {
+    fn new<S: fmt::Display>(msg: S, file_info: Option<FileInfo<'_>>, span: Option<Span>) -> Self {
         let msg = match file_info {
             Some(file_info) => format!("{msg}{file_info}"),
             None => msg.to_string(),
         };
         Self { msg, span }
     }
-
-    fn no_file_info<S: ToString>(msg: S, span: Option<Span>) -> Self {
```
GuillaumeGomez, line 263: I'm a bit surprised by this removal. I found it more explicit (and easier to `grep`) than having to know what each argument is about...

testing/tests/ui/excessive_nesting.stderr
```diff
@@ -1,6 +1,6 @@
...
+       "%}{%if 1%}{%if 1%}{%if 1%}{%if 1%}{%if 1%}\n    {%if 1%}{%if 1%}{%if 1%}{%if 1%}{"...
```
GuillaumeGomez, line 3: That is a very good improvement!

rinja_derive/src/lib.rs
Kijewski: Why would you need to grep this function? Because IDEs tell you the argument name, I think having multiple functions is seldom needed, esp. if it's only so few. And because it's not JavaScript or Python, strong typing comes to your rescue if you missorder the arguments. :)

rinja_derive/src/lib.rs
```diff
@@ -312,14 +298,13 @@ impl fmt::Display for FileInfo<'_> {
...
         } else {
-            let file_path = match std::env::current_dir() {
-                Ok(cwd) => strip_common(&cwd, self.path),
-                Err(_) => self.path.display().to_string(),
-            };
-            write!(f, "\n --> {file_path}")
+            write!(f, "\n --> {}", match std::env::current_dir() {
+                Ok(cwd) => fmt_left!(move "{}", strip_common(&cwd, self.path)),
+                Err(_) => fmt_right!("{}", self.path.display()),
+            },)
```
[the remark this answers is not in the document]
Kijewski: Oops, there were quite a few in the code.

rinja_derive/src/lib.rs
GuillaumeGomez: In case I want to know which functions still don't have file information for their error. Also, I don't use an IDE so I don't have this information. ^^'

Kijewski: I un-picked the commit.

GuillaumeGomez (approved):

Issue #272, opened 2024-11-25
Blocks defined in partial included in base template, fails to be overriden in extended templates


cipriancraciun:
Given the following example:
~~~~
## base.txt

{% block block_in_base %}
block_in_base: from base!
{% endblock %}

{% include "partial.txt" %}
~~~~
~~~~
## partial.txt

{% block block_in_partial %}
block_in_partial: from partial!
{% endblock %}
~~~~
~~~~
## extended.txt

{% extends "base.txt" %}

{% block block_in_base %}
block_in_base: from extended!
{% endblock %}

{% block block_in_partial %}
block_in_partial: from extended!
{% endblock %}
~~~~

Namely:
* we have a `base` and `extended` template,
* the `base` template indirectly defines a block `block_in_partial`, via the `{% include "partial.txt" %}`,
* `extended` although overrides the contents of the `block_in_partial`,
* when `extended` is rendered, the overriden contents fails to output;

Observed outcome of rendering `extended`:
~~~~
block_in_base: from extended!
block_in_partial: from partial!
~~~~
Expected outcome of rendening `extended`:
~~~~
block_in_base: from extended!
block_in_partial: from extended!
~~~~

----

For some reason the parser / generator fails to register the `block_in_partial` (included in `base`) as an extendable block.

Moreover, the current Rinja documentation supports the case, by implying that `{% include ... %}` statements should behave as if they wouldn't be there, and the contents of the included template would be spliced at the place of inclusion:

> The include statement lets you split large or repetitive blocks into separate template files. Included templates get full access to the context in which they're used, including local variables like those from loops:

----

What are potential use-cases:
* sometimes, especially for HTML, the base template can become too large, and the user wants to split it into multiple includes for easier development;  moreover, the developer needs a small part of these included files to be controlled by the extending template (like for example the title, some meta in head, footers, headers, etc.);
* or perhaps there are multiple `base` templates (for example one for mobile, one for HTML-only, one for modern browses, etc.) which reuse some common sections that are extracted in partials;  but the partials need some placeholders to include content controlled by extending templates;

cipriancraciun:
After running `git bisect`, it appeares that the regression was introduced in Askama at the following commit (before Rinja `v0.2` tag):
https://github.com/djc/askama/commit/5cad82f38e800a42717284f20e7e0923add1e32f

GuillaumeGomez:
The official jinja implementation returns for what you provided:

```
block_in_base: from base!
[...]

PR #279, opened 2024-11-28
Fix `block_in_include` and add explanations on what is going on

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/272.

I added extra explanations so we don't misunderstand what's going on again in the future. :)


GuillaumeGomez:
@Kijewski Any idea what's going on with the fuzzing?

Kijewski:
> @Kijewski Any idea what's going on with the fuzzing?

The fuzzing crate has no tests (well, the whole crate is a test, I'd say), which nextest made a suppressable error. Fixed in <https://github.com/rinja-rs/rinja/pull/284>.

Kijewski:
I not happy with this answer. It documents the current implementation, which is wrong, IMHO. The the last block definition should win, or better yet, the one defined in the outmost scope.

GuillaumeGomez:
It's not up to us. If we want to follow the jinja spec, then this is how it works. However I think we could greatly improve the experience for users by emitting a warning in this case.

Kijewski (approved):

Kijewski:
Very odd, I though the previous values were the values rendered by jinja2. I did test it, but maybe on an older jinja version?

I added a python test case that verifies that your changes are right.

Kijewski (approved):

GuillaumeGomez:
Comparing jinja and rinja output is a very good idea, thanks for that!

Gonna open an issue to emit a warning somehow.

Happy new year! 🥳

PR #284, opened 2024-12-10
Fix clippy warnings in rust nightly 1.85

Kijewski:


GuillaumeGomez:
Ah annoying. With newer rust version, the error output changed a bit (hence why CI is failing).

Kijewski:
I added (and removed) a gate for different rust versions before: <https://github.com/rinja-rs/rinja/commit/5955ba23c172ef6bbc0116b9ab0d6f068b86a688>. But duplicating everything felt kinda wrong. Should we simply ignore nightly for now in the CI?

GuillaumeGomez:
I think so yes. Or just run some tests, not ui ones.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #131, opened 2024-08-13
ci: update and clean up

Kijewski:

PR #285, opened 2024-12-13
Use cargo reference manifest section order

Kijewski:
In [#131] I added a check that ensures that the section in `Cargo.toml` (and there values) are in a specific order. I invented the order quite randomly, in the order which I thought might be useful.

This PR changes the order to be the same as in Cargo's reference about [The Manifest Format]. Also some missing softlinks were added.

[#131]: <https://github.com/rinja-rs/rinja/pull/131>
[The Manifest Format]: <https://doc.rust-lang.org/cargo/reference/manifest.html>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,3 @@
 [workspace]
-members = [
-    "rinja",
-    "rinja_derive",
-    "rinja_parser",
-    "testing",
-]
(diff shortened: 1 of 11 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
I don't think the order matters much but at least now it's more beautiful. :laughing:

PR #286, opened 2024-12-13
 Make rinja `#![no_std]` compatible 

### LINKED ISSUE #15: Add no_std feature?
I think the generated code can easily be `#![no_std]`. We only _need_ `std` to `impl std::error::Error for rinja::Error`. We _need_ `alloc` for a few filters. Both use cases can be feature-gated. Then I think, we can add the features ↓ and be useful to a marginally wider audience.

```toml
default = […, "std"]
std = ["alloc"]
alloc = []
```

What do you think?

Kijewski:
Resolves #15.

- [x] add tests for a full no-std environment
- [x] add tests for an `alloc`-only environment

The change is only so big because I split up the built-in filter implementation into two files: full no-std, and with alloc.

### DIFF
--- a/rinja/Cargo.toml
+++ b/rinja/Cargo.toml
@@ -11,7 +11,7 @@ license = "MIT OR Apache-2.0"
 workspace = ".."
 readme = "../README.md"
 edition = "2021"
-rust-version = "1.80"
+rust-version = "1.81"
 
 [package.metadata.docs.rs]
 features = ["full"]
@@ -21,20 +21,21 @@ rustdoc-args = ["--generate-link-to-definition", "--cfg=docsrs"]
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["config", "urlencode"]
+default = ["config", "std", "urlencode"]
+alloc = ["rinja_derive/alloc", "serde?/alloc", "serde_json?/alloc", "percent-encoding?/alloc"]
 full = ["default", "code-in-doc", "serde_json"]
 code-in-doc = ["rinja_derive/code-in-doc"]
 config = ["rinja_derive/config"]
 serde_json = ["rinja_derive/serde_json", "dep:serde", "dep:serde_json"]
+std = ["alloc", "rinja_derive/std", "serde?/std", "serde_json?/std", "percent-encoding?/std"]
 urlencode = ["rinja_derive/urlencode", "dep:percent-encoding"]
 
 [dependencies]
 rinja_derive = { version = "=0.3.5", path = "../rinja_derive" }
 
-num-traits = { version = "0.2.6", optional = true }
-percent-encoding = { version = "2.1.0", optional = true }
-serde = { version = "1.0", optional = true }
-serde_json = { version = "1.0", optional = true }
+percent-encoding = { version = "2.1.0", optional = true, default-features = false }
+serde = { version = "1.0", optional = true, default-features = false }
+serde_json = { version = "1.0", optional = true, default-features = false, features = [] }
 
 itoa = "1.0.11"
 
--- a/rinja/src/filters/alloc.rs
+++ b/rinja/src/filters/alloc.rs
@@ -1,14 +1,9 @@
-use std::cell::Cell;
-use std::convert::Infallible;
-use std::fmt::{self, Write};
-use std::ops::Deref;
-use std::pin::Pin;
+use alloc::string::String;
+use core::fmt::{self, Write};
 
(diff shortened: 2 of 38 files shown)


rinja/Cargo.toml
```diff
@@ -21,20 +21,21 @@ rustdoc-args = ["--generate-link-to-definition", "--cfg=docsrs"]
...
 urlencode = ["rinja_derive/urlencode", "dep:percent-encoding"]
 
 [dependencies]
 rinja_derive = { version = "=0.3.5", path = "../rinja_derive" }
 
-num-traits = { version = "0.2.6", optional = true }
-percent-encoding = { version = "2.1.0", optional = true }
-serde = { version = "1.0", optional = true }
-serde_json = { version = "1.0", optional = true }
+percent-encoding = { version = "2.1.0", optional = true, default-features = false }
```
GuillaumeGomez, line 36: One less dependency, yeay! \o/

rinja/src/filters/alloc.rs
```diff
@@ -0,0 +1,657 @@
...
+///     Example { value: (3, 4) }.to_string(),
+///     "<div>(3, 4)</div>"
+/// );
+/// # }
+/// ```
+///
+/// Compare with [fmt](./fn.fmt.html).
+pub fn format() {}
+
+/// Replaces line breaks in plain text with appropriate HTML
```
GuillaumeGomez, line 97: Missing a dot at the end of this sentence.

rinja/src/filters/alloc.rs
```diff
@@ -0,0 +1,657 @@
...
+    fn linebreaks(s: &str) -> String {
+        let linebroken = s.replace("\n\n", "</p><p>").replace('\n', "<br/>");
+        alloc::format!("<p>{linebroken}</p>")
+    }
+
+    let mut buffer;
+    Ok(HtmlSafeOutput(linebreaks(try_to_str!(s => buffer))))
+}
+
+/// Converts all newlines in a piece of plain text to HTML line breaks
```
GuillaumeGomez, line 131: Here too.

rinja/src/filters/alloc.rs
```diff
@@ -0,0 +1,657 @@
...
+pub fn linebreaksbr(s: impl fmt::Display) -> Result<HtmlSafeOutput<String>, fmt::Error> {
+    fn linebreaksbr(s: &str) -> String {
+        s.replace('\n', "<br/>")
+    }
+
+    let mut buffer;
+    Ok(HtmlSafeOutput(linebreaksbr(try_to_str!(s => buffer))))
+}
+
+/// Replaces only paragraph breaks in plain text with appropriate HTML
```
GuillaumeGomez, line 161: I think I'm seeing a repeated behaviour here. :rofl: 

Well, I suppose it's not too important so stopping comments for this from here.

rinja/src/filters/alloc.rs
```diff
@@ -0,0 +1,657 @@
...
+            match c.is_lowercase() {
+                true => output.push(c),
+                false => output.extend(c.to_lowercase()),
+            }
+        }
+    }
+    Ok(output)
+}
+
+#[cfg(all(test, feature = "alloc"))]
```
GuillaumeGomez, line 531: I think it can just be:

```suggestion
#[cfg(test)]
```

since this file is already imported only with `#[cfg(feature = "alloc")]`.

rinja/src/filters/alloc.rs
Kijewski: I kept the comments as they are in the current master, I just split the file in two. I think it's easier to follow the individual commits than looking at the whole diff.

rinja/src/filters/alloc.rs
Kijewski: Ah, yes. Replaced.

rinja/src/filters/alloc.rs
GuillaumeGomez: I'll make a typo pass later on then. ;)

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks a lot! Just one thing I thought about: we should mention these feature in the book and the crate(s?) docs as well.

Kijewski:
Yes, I guess it would be a good idea to add an explanation what our features do. To a more seasoned rust developer, the feature `std` and `alloc` might be self-explanatory, but `urlencode` and when I need `serde_json` might be not.

PR #289, opened 2024-12-15
Canonicalize paths to ensure includes behave as expected when folders are involved

### LINKED ISSUE #288: [Bug] `#[derive(Template)]` panics when using `include` with nested templates
I've built a small test to reproduce the bug, which you can see [here](https://github.com/rinja-rs/rinja/compare/master...LukeMathWalker:rinja:includes).

What I think is happening:

- The base template includes the fragment using a path prefix with the folder that contains it (i.e. `included/`)
- The prefixed path is not valid when used from the leaf template, where the correct path prefix would be `../included/`
- The leaf template is resolving the includes in the base template as if the base template belonged to the same folder of the leaf template, which isn't the case

I don't have a fix (yet), but putting this up here for visibility in case others bump into it.

LukeMathWalker:
Fixes #288

### DIFF
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -407,7 +407,14 @@ mod tests {
     }
 
     fn assert_eq_rooted(actual: &Path, expected: &str) {
-        let mut root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+        let mut root = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+            .canonicalize()
+            .unwrap();
+        if root.ends_with("rinja_derive_standalone") {
+            root.pop();
+            root.push("rinja_derive");
+        }
+
         root.push("templates");
         let mut inner = PathBuf::new();
         inner.push(expected);
--- a/rinja_derive/src/config.rs
+++ b/rinja_derive/src/config.rs
@@ -182,27 +182,40 @@ impl Config {
         start_at: Option<&Path>,
         file_info: Option<FileInfo<'_>>,
     ) -> Result<Arc<Path>, CompileError> {
-        if let Some(root) = start_at {
-            let relative = root.with_file_name(path);
-            if relative.exists() {
-                return Ok(relative.into());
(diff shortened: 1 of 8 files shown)


LukeMathWalker:
I'm panicking if the canonicalization fails—an alternative would be to return a compilation error. Tell me what you'd prefer.

rinja_derive/src/config.rs
```diff
@@ -185,14 +185,14 @@ impl Config {
         if let Some(root) = start_at {
             let relative = root.with_file_name(path);
             if relative.exists() {
-                return Ok(relative.into());
+                return Ok(relative.canonicalize().unwrap().into());
             }
```
Kijewski, line 189: Thank you so much for finding the issue, finding the source of the error, and presenting a solution all in time-span of less than an hour!

I wonder if it would better to change the order of the test (untested pseudo-code):

```rust
if let Ok(absolute) = root.with_file_name(path).canonicalize() {
    if absolute.exists() {
        return Ok(absolute.into());
    }
}
```

Then we don't have to care if and why the canonalizing the path didn't succeed.

LukeMathWalker: Thank you for the work on `rinja`!

Re: order—we definitely could, but I wonder if that may lead to confusion in case the canonicalization fails but the file exists? No error would be reported, the user would see the file being where it is expected to be yet `rinja` would behave as if it isn't there.

Kijewski: Yes, I think you're right. It's better to tell the user that there is a problem instead of ignoring it. I like explicit error messages better than a panic, because then fallback `Template` implementation with an empty body is generated to suppress subsequent error messages. Could you please add a check like:

```rust
relative.canonicalize().map_err(|err| CompileError::no_file_info(format_args!("could not canonicalize path {relative:?}: {err}"), None))?
```

I don't know when the canonicalization could fail, though, when it was successfully deemed to exists. If the resolved path is excessively long? Well, in the end it does not matter, I guess. :)

LukeMathWalker: Done!

Kijewski (approved):
Thank you!

PR #290, opened 2024-12-16
derive: use `file_info` in `find_template`

Kijewski:


rinja_derive/src/config.rs
```diff
@@ -182,38 +182,32 @@ impl Config {
         start_at: Option<&Path>,
         file_info: Option<FileInfo<'_>>,
     ) -> Result<Arc<Path>, CompileError> {
-        fn _find_template(config: &Config, path: &str, start_at: Option<&Path>) -> Option<PathBuf> {
+        let path = 'find_path: {
             if let Some(root) = start_at {
                 let relative = root.with_file_name(path);
                 if relative.exists() {
-                    return Some(relative);
+                    break 'find_path relative;
```
GuillaumeGomez, line 189: Wait wat. :o

I didn't know you could use `break` on a block.

GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

rinja_derive/src/config.rs
Kijewski: [Labeled blocks](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#break-from-labeled-blocks) are great. You still find some instances of `loop { break { … } }` in older code, simulating the feature. :)

PR #293, opened 2024-12-21
Improve reserved variable name error message

GuillaumeGomez:
Something I realized just after merging https://github.com/rinja-rs/rinja/pull/292. ^^'


Kijewski (approved):
Good idea to mention the reason!

Issue #176, opened 2024-09-15
`if X && let Some(bla)` condition isn't parsed correctly


GuillaumeGomez:
Currently it fails with:

```
error: unknown node `Some`
```

Kijewski:
Currently only `{% if let x = y && z %}` is implemented. On other places a binding is not understood by the parser.

GuillaumeGomez:
I was expecting something like this indeed. :)

PR #296, opened 2024-12-23
Add full support for if let chain feature

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/176.

Interestingly enough, the most difficult part was actually to ensure that variables creation was done in the right order.

### DIFF
--- a/rinja_parser/src/node.rs
+++ b/rinja_parser/src/node.rs
@@ -397,7 +397,7 @@ impl<'a> Cond<'a> {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct CondTest<'a> {
     pub target: Option<Target<'a>>,
     pub expr: WithSpan<'a, Expr<'a>>,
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
Just thought that the "auto bool" cast might be missing. Gonna check later on.

GuillaumeGomez:
And done.

rinja_parser/src/node.rs
```diff
@@ -420,7 +420,19 @@ impl<'a> CondTest<'a> {
                 ws(|i: &mut _| Target::parse(i, s)),
                 ws('='),
             )),
-            ws(|i: &mut _| Expr::parse(i, s.level.get(), false)),
+            ws(|i: &mut _| {
+                let start = *i;
+                let mut expr = Expr::parse(i, s.level.get(), false)?;
+                if let Expr::BinOp(_, _, ref mut right) = expr.inner {
+                    if matches!(right.inner, Expr::Var("set" | "let")) {
+                        *i = &start[start.len() - i.len() - 3..];
```
Kijewski, line 428:
```suggestion
                        *i = right.span.as_suffix_of(start).unwrap();
```

Kijewski:
Except for my one comment, the implementation looks good! A much smaller change than I would have anticipated.

rinja_parser/src/node.rs
GuillaumeGomez: Oh that's fancy!

GuillaumeGomez:
Applied suggestion. :)

Kijewski (approved):
:+1:

PR #297, opened 2024-12-25
parser: use drop guards to track nesting level

Kijewski:
Not for all operations the nesting level was incremented when needed and/or the un-incremented nesting level was used in subfunction calls. Binary operators such as `*` did not properly increment the nesting level.

This PR changes `Level` in such a way that it can be used to keep track of the nesting level when used in a loop. It is now impossible to accidentally refer to an old nesting level value.

Resolves <https://issues.oss-fuzz.com/issues/385256115>.
Resolves <https://github.com/rinja-rs/rinja/issues/30>.


rinja_parser/src/lib.rs
```diff
@@ -962,35 +947,75 @@ impl<'a> SyntaxBuilder<'a> {
...
+    #[inline(never)]
+    fn _fail(i: &str) -> ParseErr<'_> {
+        winnow::error::ErrMode::Cut(ErrorContext::new(
+            "your template code is too deeply nested, or the last expression is too complex",
+            i,
+        ))
+    }
+}
+
+#[must_use]
```
GuillaumeGomez, line 985: Nice. :+1:

rinja_parser/src/lib.rs
```diff
@@ -962,35 +947,75 @@ impl<'a> SyntaxBuilder<'a> {
...
+struct LevelGuard<'l> {
```
GuillaumeGomez, line 986: Please add some code explanations about why we have this type (like what you wrote in the PR description). Future us will thank us. :wink:

rinja_parser/src/lib.rs
```diff
@@ -962,35 +947,75 @@ impl<'a> SyntaxBuilder<'a> {
...
+
+impl Drop for LevelGuard<'_> {
+    fn drop(&mut self) {
+        self.level.0.set(self.level.0.get() + self.count);
+    }
+}
+
+impl LevelGuard<'_> {
+    fn nest<'a>(&mut self, i: &'a str) -> ParseResult<'a, ()> {
+        if let Some(new_level) = self.level.0.get().checked_sub(1) {
```
GuillaumeGomez, line 999: Shouldn't it be `checked_add` here? Or I completely misunderstand this algorithm...

Kijewski: I'm counting down from 128 instead of up to 128. Much simpler to debug. :)

GuillaumeGomez: Ah I see. Makes sense. Not obvious just reading the code. ^^'

rinja_parser/src/lib.rs
Kijewski: Good call. :) Added.

rinja_parser/src/lib.rs
Kijewski: I added an explanation to `Level`'s doc.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #298, opened 2024-12-27
 parser: replace closures with currying 

Kijewski:
Rebased on <https://github.com/rinja-rs/rinja/pull/297>.

IMHO the code more readable with fewer closures.

### DIFF
--- a/rinja_parser/src/expr.rs
+++ b/rinja_parser/src/expr.rs
@@ -12,8 +12,8 @@ use winnow::stream::Stream as _;
 use crate::node::CondTest;
 use crate::{
     CharLit, ErrorContext, Level, Num, ParseErr, ParseResult, PathOrIdentifier, Span, StrLit,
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
I'm not a big fan of this approach. I like the changes in the second commit but not in the first. Reading closure makes it much easier to both read and actually know what's going on whereas with the `Unbound` trait, you actually need to go check where this `bind` method comes from to understand what it's doing.

Kijewski:
I mean, you can always grep for `fn bind`. And `bind` is phrase used in multiple languages for partial-applications. IMHO the invocations are self-explanatory, the current closures aren't.

GuillaumeGomez:
It's generic, whereas closures are directly part of the `winnow` API and are easy to update if needed. IMO it's code obfuscation, not really an improvement.

Kijewski:
It would be too much work to rebase the PR.

Extracted the last commit into <https://github.com/rinja-rs/rinja/pull/322>.

Issue #283, opened 2024-12-09
thread 'rustc' has overflowed its stack


tigerros:
I have a recursive macro (which is not infinite) and it's causing the error in the title. Maybe my code is just bad, but I can't be sure since rustc crashes when I try to check. Here's the repo: https://github.com/tigerros/dia-status-rustc-bug. Here's my rustc version:
```
rustc 1.83.0 (90b35a623 2024-11-26)
binary: rustc
commit-hash: 90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf
commit-date: 2024-11-26
host: x86_64-pc-windows-msvc
release: 1.83.0
LLVM version: 19.1.1
```

Kijewski:
The macro `render_items` calls itself. A macro call in rinja is not the same as a function call in rust, as the macro body gets inlined.

But this is indeed a bug. Either we should catch this error and present a proper error message instead of causing a stack overflow, or we should relax the rules and turn macro body into proper functions. The latter option is most likely the far more difficult solution to implement.

workingjubilee:
Bug was reported to https://github.com/rust-lang/rust/issues/134092 but I am classifying it as programmer error. If you have any requests however for how this class of error is reported then we can entertain such requests, it just is not a problem that unbounded recursion can cause a stack overflow in the compiler. At that moment, you _are_ the compiler (or at least, we are running your macro as a dynamic library).

Kijewski:
Yeah, I should probably have been more explicit in my answer: It's a bug in rinja, not rust itself.

PR #299, opened 2024-12-29
derive: keep track of called macros

Kijewski:
Recursive macro calls, direct and indirect, would cause a stackoverflow.

This PR lets the macro call handler keep track of the stack of called macros we are currently in, so we can abort with an error message instead of panicking.

Resolves <https://github.com/rinja-rs/rinja/issues/283>.

### DIFF
--- a/rinja_derive/src/generator/node.rs
+++ b/rinja_derive/src/generator/node.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::collections::hash_map::{Entry, HashMap};
+use std::fmt::Write;
 use std::mem;
 
 use parser::node::{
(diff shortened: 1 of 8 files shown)


rinja_derive/src/generator/node.rs
```diff
@@ -616,14 +617,28 @@ impl<'a> Generator<'a, '_> {
...
+            (*def, mctx)
         } else {
             let def = ctx.macros.get(name).ok_or_else(|| {
                 ctx.generate_error(format_args!("macro {name:?} not found"), call.span())
             })?;
-            (def, ctx)
+            (*def, ctx)
         };
 
+        if self.seen_macros.iter().any(|(s, _)| std::ptr::eq(*s, def)) {
```
GuillaumeGomez, line 628: I have a few cases where this is problematic:

```jinja
{% macro some_macro(add) %}
    {% if add < 10 %}
        {{ add }}
        {% call next::some_macro(add + 1) %}
    {% endif %}
{% endmacro %}
```

I think we should instead count the recursion instead of forbid it.

Kijewski: Your example causes a stackoverflow in the current implementation. The condition is only checked at runtime, and the macro body is always entered. No counting needed, once we are in a loop, we'll always be in a loop.

GuillaumeGomez: Oh I see. Hum... Should we transform macros into functions?

Kijewski: I think that might be a good idea for a follow up PR. When we encounter a loop, we can transform the call into a function call instead of returning an error. I would not always turn macro into functions, though.

GuillaumeGomez: Fair enough.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks! Gonna open an issue for a better handling of recursive macros.

PR #304, opened 2025-01-04
parser: fix excessive filter blocks

Kijewski:
A `{% filter %}` block could have an unbound number of filters chained one after another. This chain would wind up as a nested structure on the heap. When dropped, this would cause a stackoverflow.

Resolves <https://issues.oss-fuzz.com/issues/386547457>.  
The bug is a regression introduced in <https://github.com/rinja-rs/rinja/pull/297>.


fuzzing/fuzz/src/parser.rs
```diff
@@ -36,7 +39,50 @@ impl<'a> super::Scenario<'a> for Scenario<'a> {
...
+                block_end,
+                expr_start,
+                expr_end,
+                comment_start,
+                comment_end,
+            } = **syntax;
+            Cow::Owned(format!(
+                "\
+SyntaxBuilder {{
+        name: \"test\",
```
GuillaumeGomez, line 64: The indent is a bit broken here. ;)

Kijewski: It's actually intentional. The `syntax` should be indented by two tabs, except on the first line, where it comes directly after `let syntax = `.

GuillaumeGomez (approved):
Apart from my comment, seems all good to me. Feel free to ignore it if you want. Thanks!

PR #310, opened 2025-01-09
Remove "Using integrations" part in book

GuillaumeGomez:
Follow-up of https://github.com/rinja-rs/rinja/pull/271.

We forgot to remove this part.


Kijewski (approved):
Good catch!

Issue #48, opened 2024-07-04
Add a `render_with_values` method?


GuillaumeGomez:
It could allow to create variables with given values before generating the template. Could be useful if there are variables outside of the type.

GuillaumeGomez:
Not possible, however we can add a new API to the `Template` proc-macro to allow to add values at compile-time.

ronnodas:
Do you mean it's not possible because the type of the value passed needs to be available at compile time? Because right now, given a template type `T` and a value type `V` you can create a wrapper struct (oversimplified)
```rust
#[derive(Template(...))]
struct WithValue {
    template: &T,
    values: &V,
}
```
 except the fields of `T` are not available without prefixing `template.`. I was imagining an extension to the `Template` derive that would do this automatically and be used like:
```rust
#[derive(Template(with_value = V, ...))]
struct T {
   ...
}
```
Making it generic over `V` seems harder, although one possibility is to essentially replace the type `V` with `impl ValueForT` for a

GuillaumeGomez:
The problem is that we only have access to the fields of the current struct, no other information. So for this feature to work, if it's at runtime, it means that all unknown fields like `{{ x }}` will need to be generated differently. Let's say we take a hashmap into the `render_with_values()` method. Then for each unknown variable, we'd replace them with getting a value from the hashmap. I'm not sure if it's a good idea in this case.

The other approach I suggested is to instead list values in the template proc-macro directly. Like that, we can still generate the best possible code. However, since it's in a proc-macro, the kind of values which can be used will be limited to booleans, integers, strings, tuples and maybe slices. So I'm not sure if it's a good idea either. It's a tricky feature.

GuillaumeGomez:
I have a use case for such a method in `docs.rs` where we have a `csp_nonce` variable which is only generated when rendering the item. So it's a "dead" field until then. Now I'm thinking that we could maybe introduce a "run-time" kind of variable/struct. Like `rinja_vars.get::<TYPE>("VAR_NAME")`? What do you think @Kijewski ?

Kijewski:
So, it's like a `Cell<Option<_>>` that is filled when first used?

Sorry, I'm on vacation right now, and I won't give any programming question a deeper thought until next Sunday. :) If you want to, maybe you could tell me / link the lines of code that could use that feature and I'll have a look when I'm back home.

GuillaumeGomez:
[...]

PR #311, opened 2025-01-09
Add support for "runtime" values

GuillaumeGomez:
Fixes https://github.com/rinja-rs/rinja/issues/48.

We can now provide values at runtime. It'll be useful for me for switching `mdBook` to `rinja` (or `askama` soon enough) as they need to allow to parse some files after `mdBook` is compiled if users provide more (not sure it'll be enough though, we'll see).

Funnily enough, it forced me to support the `call::<>` syntax with references and all.

### DIFF
--- a/rinja/src/lib.rs
+++ b/rinja/src/lib.rs
@@ -70,6 +70,7 @@ pub mod filters;
 #[doc(hidden)]
 pub mod helpers;
 mod html;
+mod values;
 
 #[cfg(feature = "alloc")]
 use alloc::string::String;
@@ -83,6 +84,7 @@ pub use rinja_derive::Template;
 pub use crate as shared;
 pub use crate::error::{Error, Result};
 pub use crate::helpers::PrimitiveType;
+pub use crate::values::{NO_VALUES, Value, Values, get_value};
 
 /// Main `Template` trait; implementations are generally derived
 ///
@@ -106,21 +108,50 @@ pub use crate::helpers::PrimitiveType;
 ///
 /// [dynamic methods calls]: <https://doc.rust-lang.org/stable/std/keyword.dyn.html>
 pub trait Template: fmt::Display + filters::FastWritable {
-    /// Helper method which allocates a new `String` and renders into it
+    /// Helper method which allocates a new `String` and renders into it.
+    #[inline]
     #[cfg(feature = "alloc")]
     fn render(&self) -> Result<String> {
+        self.render_with_values(NO_VALUES)
+    }
+
+    /// Helper method which allocates a new `String` and renders into it with provided [`Values`].
+    #[inline]
+    #[cfg(feature = "alloc")]
+    fn render_with_values(&self, values: &dyn Values) -> Result<String> {
         let mut buf = String::new();
         let _ = buf.try_reserve(Self::SIZE_HINT);
-        self.render_into(&mut buf)?;
+        self.render_into_with_values(&mut buf, values)?;
         Ok(buf)
     }
 
-    /// Renders the template to the given `writer` fmt buffer
-    fn render_into<W: fmt::Write + ?Sized>(&self, writer: &mut W) -> Result<()>;
+    /// Renders the template to the given `writer` fmt buffer.
+    #[inline]
+    fn render_into<W: fmt::Write + ?Sized>(&self, writer: &mut W) -> Result<()> {
+        self.render_into_with_values(writer, NO_VALUES)
+    }
 
-    /// Renders the template to the given `writer` io buffer
+    /// Renders the template to the given `writer` fmt buffer with provided [`Values`].
+    fn render_into_with_values<W: fmt::Write + ?Sized>(
+        &self,
+        writer: &mut W,
+        values: &dyn Values,
+    ) -> Result<()>;
+
+    /// Renders the template to the given `writer` io buffer.
+    #[inline]
     #[cfg(feature = "std")]
     fn write_into<W: io::Write + ?Sized>(&self, writer: &mut W) -> io::Result<()> {
+        self.write_into_with_values(writer, NO_VALUES)
+    }
+
+    /// Renders the template to the given `writer` io buffer with provided [`Values`].
+    #[cfg(feature = "std")]
+    fn write_into_with_values<W: io::Write + ?Sized>(
+        &self,
+        writer: &mut W,
+        values: &dyn Values,
+    ) -> io::Result<()> {
         struct Wrapped<W: io::Write> {
             writer: W,
             err: Option<io::Error>,
@@ -138,7 +169,7 @@ pub trait Template: fmt::Display + filters::FastWritable {
         }
 
         let mut wrapped = Wrapped { writer, err: None };
-        if self.render_into(&mut wrapped).is_ok() {
+        if self.render_into_with_values(&mut wrapped, values).is_ok() {
             Ok(())
         } else {
             let err = wrapped.err.take();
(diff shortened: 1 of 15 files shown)


GuillaumeGomez:
Finally fixed CI. Sorry for the noise.

rinja/src/values.rs
```diff
@@ -0,0 +1,66 @@
...
+#[cfg(feature = "std")]
+use std::collections::HashMap;
+
+/// Used to store values of any type at runtime.
+///
+/// If the `std` feature is enabled, it internally uses the [`HashMap`] type, if the `alloc`
+/// feature is enabled, it uses the [`BTreeMap`](alloc::collections::BTreeMap) type, otherwise
+/// it uses an empty type and does nothing.
+#[derive(Default)]
+pub struct Values(#[cfg(feature = "alloc")] HashMap<String, Box<dyn Any>>);
```
Kijewski, line 19: I kinda dislike the use of a HashMap in here. Unless there are quiet a lot key-value pairs, a simple slice `&[(&str, &dyn Any)]` will be excessively much faster. My guess is that a HashMap with many intermediate allocations won't be faster for less than 20 values. (I did not benchmark my guess, though. :) )

Maybe `Values` could be a trait instead?

Kijewski:
I am not 100% sure that I see the advantage over adding a field `values`?

GuillaumeGomez:
Mostly because then, you don't need to add a `values` field to all impacted type. For example in docs.rs, we have a csrp thing we pass when rendering sometimes, but that means all docs.rs types need to have this field, even though so they can't set it and have no use for it.

rinja/src/values.rs
GuillaumeGomez: Not using a map sounds good to me. I suppose by making `Values` a trait, we would let it up to users to implement it on whatever they want? But then the question is: should we implement on some types by default as well (like `&(&str, &dyn Any)` or `HashMap<String, Box<dyn Any>>`)?

Kijewski:
Ah, yes, a csrf token sounds like a good use case.

rinja/src/values.rs
Kijewski: I think it would be useful to implement it for some types, so it works "out of the box". But I would not seal the trait.

GuillaumeGomez: What do you mean by "seal the trait"?

Kijewski:
```rust
mod private { // ← not `pub`, no you can't implement this trait outside of the crate
    pub trait Sealed {}
}

pub mod MyTrait: private::Sealed {}
```

<https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/#what-are-sealed-traits>

GuillaumeGomez: Ah so making the trait impossible to implement outside of the crate. I didn't know there was a name for that. TIL, thanks!

GuillaumeGomez: Big question: making the `Values` trait fully generic over the key creates a lot of issues:
 * DynTemplate is not dyn-compatible anymore
 * We get unsolvable errors like:
     ```
     4  | #[derive(Template)]
   |          ^^^^^^^^
   |          |
   |          expected `&K`, found `&str`
   |          arguments to this method are incorrect
    ```

Would it be ok to force the key to be `AsRef<str>`?

Kijewski: Yes, I think it's a good idea to force the key to be str'ish. Maybe `K: Borrow<str>` would be the best solution? So the interface is the same as `HashMap::get()`, and because `Borrow<str>` implies that `K`'s hash is the same as the borrowed `str`'s hash.

GuillaumeGomez: Sounds good to me.

GuillaumeGomez:
Turned `Values` into a trait as discussed and implemented on some types. Sadly, even with the simplified version, there are still some frustrating limitations. For example, we cannot use `&str` as key for `HashMap` otherwise `Values` is not implemented. And when I tried to remove the `for<'a>` limitation, I encountered some lifetime issues I couldn't solve.

I'm very curious to see if you will be able to see a way to get around them.

Also, since `DynTemplate` cannot use generics, I currently commented out the `_with_values` methods on it.

rinja/src/lib.rs
```diff
@@ -83,6 +84,7 @@ pub use rinja_derive::Template;
 pub use crate as shared;
 pub use crate::error::{Error, Result};
 pub use crate::helpers::PrimitiveType;
+pub use crate::values::{NO_VALUES, Value, Values};
```
GuillaumeGomez, line 87: I would also reexport `get_value` since people can always call `rinja::get_value` in their templates or even in their code (who knows?).

rinja_derive/src/generator/node.rs
```diff
@@ -1167,7 +1167,8 @@ impl<'a> Generator<'a, '_> {
                         idx
                     };
                     lines.write(format_args!(
-                        "(&&rinja::filters::Writable(expr{idx})).rinja_write(__rinja_writer)?;",
+                        "(&&&rinja::filters::Writable(expr{idx})).\
```
GuillaumeGomez, line 1170: Why adding a new reference here btw?

GuillaumeGomez:
I think it's a much simpler approach, love it! Shouldn't have tried to force too much the trait argument approach... My turn now (some docs to update and some edges I'd like to soften). Thanks!

rinja_derive/src/generator/expr.rs
```diff
@@ -584,6 +586,7 @@ impl<'a> Generator<'a, '_> {
                 return Ok(DisplayWrap::Unwrapped);
             }
         }
+        #[allow(clippy::literal_string_with_formatting_args)]
```
GuillaumeGomez, line 589: Oh that's a new lint I added haha.

Kijewski:
> My turn now (some docs to update and some edges I'd like to soften)

I'm happy that you like my approach, and I'm interested to see what you make out of it! :)

rinja_derive/src/generator/expr.rs
Kijewski: Thank you, I like the lint! I forgot more than once to surround the string with `format!(…)`. :)

GuillaumeGomez:
Seems like we broke rustc. :o

rinja_derive/src/generator/node.rs
Kijewski: I added the values as new argument to `WriteWritable`. This way it gets automagically passed on to any `{{ template }}`s. To make this work I need to tell `Template`s other `FastWritable`s apart.

Kijewski:
Haha, oops.

Can you open an issue? I guess you have more experience with that than me. (And I just noticed that it's 20:40 already and that I have not eaten all day.)

GuillaumeGomez:
Off you go then. Enough computer for today for you! Eat well.

And yes, gonna do that. :)

GuillaumeGomez:
While trying the new changes and in particular the new filter, I'm not a bit fan of the incoherence introduced: the argument is not argument an argument but a generic. What do you think if we instead added support for generics on filters? Or simpler (although I like the filter): we don't make this filter.

Kijewski:
Yes, having a type as a function argument is a bit odd. Something like `"key" | value::<u32>` should be readable enough. Should we make the `::` optional, so you can write `"key" | value<u32>`, too?

I wouldn't like a free function `value("key")` very much, because it shadows the method `Self::value()`.

GuillaumeGomez:
> Yes, having a type as a function argument is a bit odd. Something like `"key" | value::<u32>` should be readable enough. Should we make the `::` optional, so you can write `"key" | value<u32>`, too?

I'd prefer to keep the syntax Rust compatible as much as possible (and also I already implemented to parsing of `::<>` so less work hehe).

> I wouldn't like a free function `value("key")` very much, because it shadows the method `Self::value()`.

Me neither. Using `rinja::get_value` sounds good enough to me.

If it sounds good to you, I'll make the changes.

GuillaumeGomez:
Finally done! I special-cased the `rinja::get_value` path. I also needed to add even more generics support (for `Expr::Call`) so I think we support generics pretty much everywhere now. Talk about side-effects :laughing: I also extended them to all them to be recursive which allows things like `HashMap<String, Vec<u32>>`.

I think I did pretty much everything I thought was missing. Thanks a lot for your help getting it to the finish line! :)

GuillaumeGomez:
Hum, I think it should be mostly fine. The git history will be completely broken so I'm gonna have a good time. :')

GuillaumeGomez:
Much smaller change now. Now two commits only and just value related changes. :)

rinja_derive/src/generator/expr.rs
```diff
@@ -676,6 +677,68 @@ impl<'a> Generator<'a, '_> {
...
+            }
+            _ => {
+                return Err(
+                    ctx.generate_error("expected a type name, not an expression", ty.span())
+                );
+            }
+        }
+        Ok(())
+    }
+
```
Kijewski, line 741:
```suggestion
```
The method is not used anymore.

Kijewski:
I wonder why neither rustc nor clippy have a problem this the unused method? It calls itself, but nobody else calls it.

Except for the one comment, it look great!

GuillaumeGomez:
It's super strange indeed... Gonna report that. In the meantime, fixed it.

Kijewski (approved):
:+1:

rinja_derive/src/generator/expr.rs
GuillaumeGomez: It's because it starts with an underscore. XD

GuillaumeGomez:
> I wonder why neither rustc nor clippy have a problem with the unused method? It calls itself, but nobody else calls it.

It's because it starts with an underscore. XD

PR #317, opened 2025-01-17
Add support for generics syntax in parser

GuillaumeGomez:
Extracted the parser generics support add from https://github.com/rinja-rs/rinja/pull/311.

### DIFF
--- a/rinja_derive/src/generator/expr.rs
+++ b/rinja_derive/src/generator/expr.rs
@@ -2,7 +2,8 @@ use std::borrow::Cow;
 
 use parser::node::CondTest;
 use parser::{
-    CharLit, CharPrefix, Expr, Filter, IntKind, Num, Span, StrLit, StrPrefix, Target, WithSpan,
+    Attr, CharLit, CharPrefix, Expr, Filter, IntKind, Num, Span, StrLit, StrPrefix, Target,
+    WithSpan,
 };
 
 use super::{
@@ -38,7 +39,7 @@ impl<'a> Generator<'a, '_> {
             Expr::Var(s) => self.visit_var(buf, s),
             Expr::Path(ref path) => self.visit_path(buf, path),
             Expr::Array(ref elements) => self.visit_array(ctx, buf, elements)?,
-            Expr::Attr(ref obj, name) => self.visit_attr(ctx, buf, obj, name)?,
+            Expr::Attr(ref obj, ref attr) => self.visit_attr(ctx, buf, obj, attr)?,
             Expr::Index(ref obj, ref key) => self.visit_index(ctx, buf, obj, key)?,
             Expr::Filter(Filter {
                 name,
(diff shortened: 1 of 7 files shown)


rinja_parser/src/expr.rs
```diff
@@ -685,12 +724,59 @@ impl<'a> Suffix<'a> {
...
+    pub refs: usize,
+    pub ty: &'a str,
+    pub generics: Vec<WithSpan<'a, TyGenerics<'a>>>,
+}
+
+impl<'i> TyGenerics<'i> {
+    fn parse(i: &mut &'i str, level: Level<'_>) -> ParseResult<'i, WithSpan<'i, Self>> {
+        let start = *i;
+        (
+            ws(identifier_with_refs),
```
Kijewski, line 753: This allows e.g. `&&u32`, but would fail for `&&core::primitive::u32`. I think we need to support paths in here. What do you think.

Kijewski:
I made minor changes to keep track of the nesting level, so `A<A<A<A<A<…>>>>>>` won't cause problems.

rinja_parser/src/expr.rs
GuillaumeGomez: Didn't even think about it but I agree.

GuillaumeGomez:
Funnily enough, I was thinking about this case the other day but it was late and I forgot afterwards. ^^'

GuillaumeGomez:
Anyway, gonna allow to have paths as generics tomorrow.

Kijewski:
Oops, I didn't see that you answered immediately. I'll give the PR back to you, now. :)

rinja_derive/src/generator/expr.rs
```diff
@@ -398,8 +402,10 @@ impl<'a> Generator<'a, '_> {
         buf: &mut Buffer,
         name: &str,
         args: &[WithSpan<'_, Expr<'a>>],
+        generics: &[WithSpan<'_, TyGenerics<'_>>],
         node: Span<'_>,
     ) -> Result<DisplayWrap, CompileError> {
+        ensure_filter_has_no_generics(ctx, name, generics, node)?;
```
GuillaumeGomez, line 408: I'm not a big fan because this call can (and will be) forgotten when we add new filters. I voluntarily moved this check outside of the function to special-case filters which actually were making use of generics so by default, all the others would error.

GuillaumeGomez:
> Oops, I didn't see that you answered immediately. I'll give the PR back to you, now. :)

Oh no, less work for me. :stuck_out_tongue: 

Just waiting for you to tell me what you think about my logic and if that sounds ok to you, I'll change the code and move the check in the "common filter path" instead.

rinja_derive/src/generator/expr.rs
Kijewski: Actually I don't think that the line get be forgotten. Without the line, the argument `generics` is unused, and rust will print a warning.

For functions like `visit_filter()` I like to use a clear dispatch pattern. The brains should be in the dispatch target functions, the dispatcher should be as stupid and as simple as possible. And because an unused argument would render a warning, I don't see any drawbacks in here.

But if you dislike this implementation very much, then I could live is an implementation with more logic inside the dispatcher, too.

GuillaumeGomez: It's a valid argument although I think centralizing checks when possible is a better approach. And also, sometimes you can just prepend with `_` to do your things and forget about it later on (happened to me :sweat_smile:).

GuillaumeGomez:
I made the default syntax configurable and optional. Now, if there is a README, it will have the same behaviour as now. However, if it's coming from the crate-level docs because there is no README, it will apply the same rules as rustdoc: if there is no language specified, then by default it's rust.

GuillaumeGomez:
Sorry, it was meant for a fix I wrote on docs.rs. ^^'

rinja_derive/src/generator/expr.rs
GuillaumeGomez: Yeay. :D

But another day, tonight it's video game time. :)

Kijewski: Hehe, have fun! :)

GuillaumeGomez:
Applied the changes we talked about. I think we're good to go for this part?

Kijewski (approved):
Yes, the implementation looks fine. Thank you!

PR #318, opened 2025-01-21
derive: unify filter argument handling

Kijewski:


rinja_derive/src/generator/expr.rs
```diff
@@ -1065,18 +1049,50 @@ impl<'a> Generator<'a, '_> {
...
+fn unexpected_filter_arguments(
+    ctx: &Context<'_>,
+    name: &str,
+    args: &[WithSpan<'_, Expr<'_>>],
+    node: Span<'_>,
+) -> CompileError {
+    ctx.generate_error(
+        format_args!(
+            "unexpected argument{} in `{name}` filter",
+            if args.len() == 2 { "" } else { "s" },
```
GuillaumeGomez, line 1076:
```suggestion
            "expected 1 argument in `{name}` filter, found {}",
            args.len(),
```

Better than handling plural: not handling plural. :D

Kijewski: Good idea, but the "1" is not true in all cases. :) I reworded the message, and implemented it with less magic. <del>🪄 </del>

GuillaumeGomez:
Conflict. I'll approve once updated otherwise it doesn't work. :')

Kijewski:
Would conflict with #311, too. I'll rebase it after #311 is merged. :)

Kijewski:
I expanded *slightly* on the initial idea.

rinja_derive/src/generator/expr.rs
```diff
@@ -284,11 +282,7 @@ impl<'a> Generator<'a, '_> {
         name: &str,
         args: &[WithSpan<'_, Expr<'a>>],
         generics: &[WithSpan<'_, TyGenerics<'_>>],
-        node: Span<'_>,
     ) -> Result<DisplayWrap, CompileError> {
-        if BUILTIN_FILTERS_NEED_ALLOC.contains(&name) {
-            ensure_filter_has_feature_alloc(ctx, name, node)?;
-        }
```
Kijewski, line 291: Check was in the wrong method. Moved to `_visit_builtin_filter`.

rinja_derive/src/generator/expr.rs
```diff
@@ -303,43 +297,18 @@ impl<'a> Generator<'a, '_> {
         buf: &mut Buffer,
         name: &str,
         args: &[WithSpan<'_, Expr<'a>>],
-        generics: &[WithSpan<'_, TyGenerics<'_>>],
```
Kijewski, line 306: No built-in filter needs generics.

rinja_derive/src/generator/expr.rs
```diff
@@ -256,25 +257,22 @@ impl<'a> Generator<'a, '_> {
...
             "ref" => Self::_visit_ref_filter,
             "safe" => Self::_visit_safe_filter,
-            "urlencode" => Self::_visit_urlencode_filter,
-            "urlencode_strict" => Self::_visit_urlencode_strict_filter,
-            "value" => return self._visit_value(ctx, buf, args, generics, node, "`value` filter"),
-            name if BUILTIN_FILTERS.contains(&name) => {
-                return self._visit_builtin_filter(ctx, buf, name, args, generics, node);
+            "urlencode" | "urlencode_strict" => Self::_visit_urlencode_filter,
+            name if BUILTIN_FILTERS.contains(&name) => Self::_visit_builtin_filter,
+            "value" => {
```
GuillaumeGomez: Why moving `"value"` below the `name` check?

rinja_derive/src/generator/expr.rs
```diff
@@ -256,25 +257,22 @@ impl<'a> Generator<'a, '_> {
             "format" => Self::_visit_format_filter,
             "join" => Self::_visit_join_filter,
             "json" | "tojson" => Self::_visit_json_filter,
-            "linebreaks" => Self::_visit_linebreaks_filter,
-            "linebreaksbr" => Self::_visit_linebreaksbr_filter,
-            "paragraphbreaks" => Self::_visit_paragraphbreaks_filter,
+            "linebreaks" | "linebreaksbr" | "paragraphbreaks" => Self::_visit_linebreaks_filters,
```
GuillaumeGomez: So we're back giving the filter name as argument? That's one extra argument we could skip but if you feel like it's better this way, then let's go with it.

rinja_derive/src/generator/expr.rs
```diff
@@ -249,32 +249,164 @@ impl<'a> Generator<'a, '_> {
...
-            "safe" => Self::_visit_safe_filter,
-            "urlencode" | "urlencode_strict" => Self::_visit_urlencode_filter,
-            name if BUILTIN_FILTERS.contains(&name) => Self::_visit_builtin_filter,
-            "value" => {
-                return self._visit_value(ctx, buf, args, generics, node, "`value` filter");
-            }
-            _ => {
-                return self._visit_custom_filter(ctx, buf, name, args, generics);
-            }
+        type Filter = for<'a> fn(
```
GuillaumeGomez: Considering the amount of code it adds, what's the perf gain?

Kijewski:
```text
$ cargo bench --bench filters
noop                    time:   [44.123 µs 44.205 µs 44.283 µs]
                        change: [+0.8450% +1.2321% +1.6350%] (p = 0.00 < 0.05)
                        Change within noise threshold.

no_filters              time:   [51.914 µs 52.539 µs 53.235 µs]
                        change: [+0.0880% +0.9181% +1.8833%] (p = 0.05 < 0.05)
                        Change within noise threshold.

few_filters             time:   [76.035 µs 76.535 µs 76.984 µs]
                        change: [+1.2029% +1.8326% +2.4187%] (p = 0.00 < 0.05)
                        Performance has regressed.

all_filters             time:   [96.863 µs 97.574 µs 98.471 µs]
                        change: [-8.9618% -8.4126% -7.7142%] (p = 0.00 < 0.05)
                        Performance has improved.

some_filters_twice      time:   [70.936 µs 71.177 µs 71.383 µs]
                        change: [-8.5514% -8.1406% -7.7478%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The first two tests are only to measure that my CPU is probably running hotter on the second run. (It shows +216.0°C, but I'm fairly confident that the measurement is incorrect. :D)

GuillaumeGomez:
I'm shared here: the code complexity increases a lot, and although % improvement is noticeable, we're talking in µs. I don't think it's noticeable for a user and therefore, I don't think it's worth the extra complexity.

Kijewski:
Back to a simple `match`. The timings are actually about the same.

rinja_derive/src/generator/expr.rs
GuillaumeGomez: You still didn't answer to this one. :wink:

Kijewski: I moved the filters that `return` to the bottom, to make it easier to follow the code. In the last/current implementation it's strictly alphabetical again. :)

GuillaumeGomez: In this case I don't think it's the right way to do it. In the (unlikely) case we added (by mistake) `value` into `BUILTIN_FILTERS`, then we'd get no way of noticing that the real `value` filter is not called until the tests failed.

But anyway, in the current case I think the code can be improved a bit (adding suggestion comment below). :stuck_out_tongue:

rinja_derive/src/generator/expr.rs
```diff
@@ -256,25 +257,22 @@ impl<'a> Generator<'a, '_> {
...
-            name if BUILTIN_FILTERS.contains(&name) => {
-                return self._visit_builtin_filter(ctx, buf, name, args, generics, node);
+            "urlencode" | "urlencode_strict" => Self::_visit_urlencode_filter,
+            name if BUILTIN_FILTERS.contains(&name) => Self::_visit_builtin_filter,
+            "value" => {
+                return self._visit_value(ctx, buf, args, generics, node, "`value` filter");
+            }
+            _ => {
+                return self._visit_custom_filter(ctx, buf, name, args, generics);
             }
```
GuillaumeGomez:
```suggestion
            "value" => {
                return self._visit_value(ctx, buf, args, generics, node, "`value` filter");
            }
            name => if BUILTIN_FILTERS.contains(&name) {
                Self::_visit_builtin_filter
            } else {
                return self._visit_custom_filter(ctx, buf, name, args, generics);
            },
```

This allows to remove one arm and have the `contains` call to be done after all other strings have been checked.

Kijewski:
(Mostly) implemented in other PRs.

PR #319, opened 2025-01-21
Make filters `|fmt` and `|format` available with feature `"alloc"`, but without `"std"`

Kijewski:


GuillaumeGomez (approved):

GuillaumeGomez:
Nice catch!

PR #324, opened 2025-01-28
[WIP] Dynamic rendering mode

Kijewski:


Kijewski:
I started the PR before #255, that's why there is a conflict.

Could you have a first cursory look if you like the idea, @GuillaumeGomez? IMHO that could be a killer-feature. :)

How to use:
* Run the axum-example: `cargo r`
* Don't stop the process!
* Open <http://127.0.0.1:8080/>
* Change the background color to red.
* Recompile (in a new tab): `cargo b`
* Hit F5.

Values most likely won't work, because I don't know how to serialize them.

GuillaumeGomez:
What's this feature about btw?

Kijewski:
Live-updating the templates, one of he most requested unimplemented features in askama: <https://github.com/rinja-rs/askama/issues/491>. (Instead of manually compiling, something like `cargo watch` should be used.)

GuillaumeGomez:
:open_mouth: 

Wow. Ok, let me get through it.

rinja/src/dynamic/child.rs
```diff
@@ -0,0 +1,207 @@
...
+use std::{eprintln, format};
+
+use linkme::distributed_slice;
+use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader, Stdin, Stdout, stdin, stdout};
+use tokio::spawn;
+use tokio::sync::{Mutex, oneshot};
+
+use super::{DYNAMIC_ENVIRON_KEY, MainRequest, MainResponse, Outcome};
+
+const PROCESSORS: usize = 4;
```
GuillaumeGomez, line 16: Any reason why four?

rinja/src/dynamic/child.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[inline(never)]
+pub(crate) fn run_dynamic_main() {
+    std::env::set_var(DYNAMIC_ENVIRON_KEY, "-");
+
+    let mut entries: Vec<_> = DYNAMIC_TEMPLATES.iter().map(|entry| entry.name()).collect();
+    entries.sort_unstable();
+    eprintln!("templates implemented by subprocess: {entries:?}");
```
GuillaumeGomez, line 24: `println!` instead maybe? It's not an error after all.

rinja/src/dynamic/child.rs
```diff
@@ -0,0 +1,207 @@
...
+    for window in entries.windows(2) {
+        if let &[a, b] = window {
+            if a == b {
+                eprintln!("duplicated dynamic template {a:?}");
```
GuillaumeGomez, line 28: Same here.

rinja/src/dynamic/child.rs
```diff
@@ -0,0 +1,207 @@
...
+        line.push(b'\n');
+        if let Err(err) = serde_json::to_writer(&mut line, &MainResponse { callid, outcome }) {
+            let err = format!("could not serialize response: {err}");
+            return done(Err(std::io::Error::new(ErrorKind::InvalidData, err))).await;
+        }
+        line.push(b'\n');
+
+        let is_done = {
+            let mut stdout = stdout.lock().await;
+            stdout.write_all(line).await.is_err() || stdout.flush().await.is_err()
```
GuillaumeGomez, line 124: For it to be done, it needs to return an error? Seems a bit strange. At the very least, a code comment would be appreciated here.

rinja/Cargo.toml
```diff
@@ -26,11 +26,20 @@ harness = false
...
 serde = { version = "1.0", optional = true, default-features = false }
-serde_json = { version = "1.0", optional = true, default-features = false, features = [] }
+serde_json = { version = "1.0", optional = true, default-features = false }
 
-itoa = "1.0.11"
+# needed by feature "dynamic"
+linkme = { version = "0.3.31", optional = true }
+notify = { version = "8.0.0", optional = true }
+parking_lot = { version = "0.12.3", optional = true, features = ["arc_lock", "send_guard"] }
+tokio = { version = "1.43.0", optional = true, features = ["macros", "io-std", "io-util", "process", "rt", "sync", "time"] }
```
Kijewski, line 42: * `linkme` and `notify` are absolutely needed.
* I tried to implement the PR without an `async` runtime, but well, tokio is just too well implemented to invent the wheel twice.
* `parking_lot` could maybe be replace with an `std::sync::Barrier` of some sort. It's far from trivial, though.

rinja/src/dynamic/child.rs
Kijewski: STDIN & STDOUT are reserved from communication with the parent process.

rinja/src/dynamic/child.rs
Kijewski: [](https://xkcd.com/221/)  
No. :) Maybe the processor count would be better?

GuillaumeGomez: Why do we need multiple processors?

rinja/src/dynamic/child.rs
Kijewski: If the connection to the parent process was interrupted, the child process can exit. Severing the connection is the first step in [killing a subprocess](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.start_kill).

rinja/src/dynamic/child.rs
GuillaumeGomez: Sounds like a code comment is needed. :)

rinja/src/dynamic/child.rs
Kijewski: Without it the templates are rendered one after another.

On my not-so-recent AMD Ryzen 7 2700X system, with four workers and in release mode, the rendering takes about the same time in dynamic mode as non-dynamic: 68µs vs 65µs in he axum example.

GuillaumeGomez:
I think I'll need more technical explanations. From what I can gather, we run a server watching files that then sends serialized JSON data to a client which re-renders the needed template files.

However, what are the clients/servers here exactly, how are they rebuilding templates is still something I cannot figure out. ^^'

rinja/src/dynamic/child.rs
GuillaumeGomez: I see, another code comment. :D

rinja/src/dynamic/child.rs
Kijewski: Yeah, possibly even a small UML'like actor diagram.

Kijewski:
> If you just want to have changes to the template HTML file apply without recompilation, can't we just update the derived render function to use std::fs::read instead of include_bytes in debug builds?

The short explanation for the project reads

> A template rendering engine based on Jinja, generating type-safe Rust code at compile time. 

Do you see the problem?

PR #330, opened 2025-01-31
chore: Upgrade to Winnow 0.7 

epage:
~~I am not thrilled with the fact that annotations of some kind (I used
no-op `map_err`s here) are needed for the errors and will be digging
into this to better understand why.  The code in `toml_edit` is very
similar and yet it doesn't need them.~~

Those `map_err`s were needed to workaround `impl From for ErrMode`.  `ErrMode` is now optional, so its unclear when calling a generic error if its returning a `ErrorContext` or a `ErrMode<ErrorContext>`.  See winnow-rs/winnow#735

### DIFF
--- a/rinja_parser/Cargo.toml
+++ b/rinja_parser/Cargo.toml
@@ -23,7 +23,7 @@ harness = false
 [dependencies]
 memchr = "2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-winnow = "0.6.23"
+winnow = "0.6.26"
 
 [dev-dependencies]
 criterion = "0.5"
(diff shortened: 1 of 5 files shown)


GuillaumeGomez:
Nice thanks!

@Kijewski Any idea what's going on with the failing CI?

Kijewski:
Looks like MicroSoft made DevSkim depend on a newer .net version, but they forgot to tell the script to install the never version, too: <https://github.com/microsoft/DevSkim-Action/issues/30>.

Kijewski (approved):
Thank you!

PR #332, opened 2025-02-01
parser: add missing level guard in `if let`

Kijewski:
Resolves <https://issues.oss-fuzz.com/issues/393401467>


Kijewski:
The default value for `overflow_delimited_expr = true` was reverted to `false`: <https://github.com/rust-lang/rust/pull/136312>. The reformatting should be done in its own PR, as it touches close to every file.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #337, opened 2025-02-08
 derive: implement template attribute `blocks`

### LINKED ISSUE #328: Generate methods to render each Block independently?
My use case with [HTMX](https://htmx.org) is that a response may either include the full page (extending `base.html`) or just a single block `content` depending on a request header. Right now, using Minijinja, this is implemented centrally in a custom middleware which decides what to render. 

I'd love to switch to Rinja, but right now this is not possible without having two versions of every struct, modifying every endpoint to either return the full- or the partial template struct, with the latter one annotated with `#[attribute(block = "content")]`. 

Would you consider extending the `Template` derive macro to be able to generate methods on the struct which allow rendering each block individually? I'm thinking of something like:

```jinja2
{% block title %}Hello {{title}}{% endblock %}
{% block content %}...{% endblock %}
```

```rust
#[derive(Template)]
#[attribute(file = "foo.html", generate_block_methods = "true")]
struct Foo {
    title: String
}

// Generated

impl Template for Foo {..}

impl Foo {
    fn render_block_title(&self) -> Result<String> { ... }
    fn render_block_content(&self) -> Result<String> { ... }
}
```

The methods could even be used in the existing `render` methods (possibly annotated `#[inline]`) to avoid generating redundant code.

Another approach could be extending `Template` and `DynTemplate` with a `fn render_block(&self, block: &str) -> ...` method.

Kijewski:
The new attribute `blocks` is feature-gated because it depends on `syn/full`, which we don't need otherwise.

The implementation is quite big (8 files changed, 269 insertions(+), 9 deletions(-)), but also reasonably self-contained, so I'm still happy. :)

Todos:

- [ ] add documentation

Resolves #328.

### DIFF
--- a/rinja/Cargo.toml
+++ b/rinja/Cargo.toml
@@ -44,7 +44,7 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["config", "std", "urlencode"]
-full = ["default", "code-in-doc", "serde_json"]
+full = ["default", "blocks", "code-in-doc", "serde_json"]
 
 alloc = [
     "rinja_derive/alloc",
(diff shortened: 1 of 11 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

GuillaumeGomez:
Arf, didn't realize the doc was missing, was too excited about the changes. ><

PR #340, opened 2025-02-08
 book: document `blocks`

Kijewski:


rinja_derive/src/lib.rs
```diff
@@ -156,7 +206,6 @@ pub fn derive_template(input: TokenStream12) -> TokenStream12 {
         Err(err) => {
             let msgs = err.into_iter().map(|err| err.to_string());
             let ts = quote! {
-                span =>
```
Kijewski, line 159: The change only matters for our rinja playground. In actual code, `derive_template` would never be called if the input `struct` does not contain valid rust syntax.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #341, opened 2025-02-08
derive: fail if requested template block is missing

### LINKED ISSUE #336: `#[template(block = "something_that_is_missing")]` does **not** fail to compile
Is this intentional? I would have assumed that this was a compile-time error. Instead the output is empty. I think we should *at least* render a warning, because this current behavior can hide spelling errors.

<https://rinja-rs.github.io/play-rinja/?saved=eJwrM4wxUI4uSc0tyEksSdVIrShRsFVQyijJzVHSUUjKyU_OBvFLUotLlDRjuYpLikqTSxQ8UnNy8sPzi3JSrGMMwvPz1EsUklIVilNT8_QALE0anA..>

Kijewski:
Previously it was only tested if the block exists if there were any blocks. If there weren't any, then an empty result would be rendered.

Fixes #336.

### DIFF
--- a/rinja_derive/src/lib.rs
+++ b/rinja_derive/src/lib.rs
@@ -284,21 +284,23 @@ fn build_template_item(
 
     let ctx = &contexts[&input.path];
     let heritage = if !ctx.blocks.is_empty() || ctx.extends.is_some() {
-        let heritage = Heritage::new(ctx, &contexts);
-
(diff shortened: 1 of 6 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #346, opened 2025-02-09
Implement `Values` for all collection types

Kijewski:


GuillaumeGomez:
Could you add a test for each newly added one please?

Kijewski:
Oops, good call! Turns out that it's not possible to use a `HashSet`, `BTreeSet` or `BinaryHeap` as `Values`, because `Value` is not `Ord` or `Hash`.

GuillaumeGomez:
But it works on `HashMap`, so I'm a bit surprised it wouldn't work on sets.

GuillaumeGomez:
Well in any case, this is already an improvement so let's merge it, thanks!

GuillaumeGomez (approved):

PR #349, opened 2025-02-09
Rename `Escaper` argument `fmt` into `dest`

Kijewski:
I was confused for a moment why an escaper would accept an `fmt::Formatter` or `fmt::Arguments`.


GuillaumeGomez (approved):

GuillaumeGomez:
That was indeed not great. Thanks!

PR #358, opened 2025-02-24
Fix links in integrations.md v0.3.x. fixes #357 

### LINKED ISSUE #357: Dead links in the "readthedocs" doc website
Hello,

In the [doc website](https://rinja.readthedocs.io) the top-right link `Suggest an edit` points to the `master` branch of the code.

Problem:

Some files don't exist in the `master` branch but do in previous `stable` releases of the crate.

How to reproduce:

In the [Integrations](https://rinja.readthedocs.io/en/stable/integrations.html) page, all the `the example` links are broken, but also the `Suggest an edit` link! They all points to the `master` branch where the files they point to don't exist anymore.

Suggested solution:

I'm not sure about the authoring workflow for rinja, but since I was visiting the `stable` version of the docs (selected on the bottom left), links should point to the same `stable` version of the code, not `master`.

DjebbZ:
Points to the `v0.3.5` instead of `master`. 

Details in #357

### DIFF
--- a/book/src/integrations.md
+++ b/book/src/integrations.md
@@ -8,7 +8,7 @@ In your template definitions, replace `rinja::Template` with
 Enabling the `with-rocket` feature appends an implementation of Rocket's
 `Responder` trait for each template type. This makes it easy to trivially
 return a value of that type in a Rocket handler. See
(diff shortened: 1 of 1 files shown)


Kijewski:
Thank you for your PR! The fix should not be needed anymore, as everything should be moved by now.

PR #360, opened 2025-03-11
ci: ignore typos in librustdoc sources

Kijewski:
We copied the sources verbatim, so it's no use to fix any typos.

Also, "typ" is not really a typo in here, because "type" would not have been a valid identifier.


GuillaumeGomez:
Thanks!

PR #363, opened 2025-03-11
Add bleeding edge API documentation to rtd

Kijewski:
Resolves <https://github.com/askama-rs/askama/issues/362>.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,6 +6,7 @@ build:
   os: ubuntu-lts-latest
   tools:
     python: "3.12"
+    rust: "latest"
   commands:
     # Generate "book/theme/index.hbs" as "skeleton" of the generated pages.
     - book/update-theme.py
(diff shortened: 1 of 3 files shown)


Kijewski:
@GuillaumeGomez, I disabled the option to force-push or ignore the CI in [the settings](https://github.com/askama-rs/askama/settings/branch_protection_rules/51395530). (After accidentally force-pushing to the old repo.) I hope that's OK. I there is a dire need to ignore the CI or do a force-push, then it's just one click to disable the guard.

GuillaumeGomez:
I think it's a good idea. Force push is a very big power to have, some might even say too big. :laughing: 

Anyway, looks good to me, thanks!

Kijewski:
I forgot to update the API hook, so the version was not built automagically, but triggering the build manually worked: <https://askama.readthedocs.io/en/latest/doc/askama/index.html>. :)

PR #365, opened 2025-03-12
Added support for typst in default text escapers

xabi00:
I did this exact Pull Request in what today is named 'askama-old' repository. These changes were successfully merged; however, they occurred after Rinja's fork, which means they are missing from the current codebase.

### DIFF
--- a/askama_derive/src/config.rs
+++ b/askama_derive/src/config.rs
@@ -373,7 +373,7 @@ static DEFAULT_ESCAPERS: &[(&[&str], &str)] = &[
         ],
         "Html",
     ),
-    (&["md", "none", "txt", "yml", ""], "Text"),
+    (&["md", "none", "txt", "typ", "yml", ""], "Text"),
 ];
 
 #[cfg(test)]
@@ -656,7 +656,7 @@ mod tests {
                     "askama::filters::Html".into()
                 ),
                 (
-                    str_set(&["md", "none", "txt", "yml", ""]),
+                    str_set(&["md", "none", "txt", "typ", "yml", ""]),
                     "askama::filters::Text".into()
                 ),
             ]


Kijewski:
Sorry, but I don't think this is a good addition. Typst has multiple characters that need escaping depending on the context, at least newlines, but possibly also `$`, `-` and `=` as far as I can tell.

askama_derive/src/config.rs
```diff
@@ -373,7 +373,7 @@ static DEFAULT_ESCAPERS: &[(&[&str], &str)] = &[
         ],
         "Html",
     ),
-    (&["md", "none", "txt", "yml", ""], "Text"),
+    (&["md", "none", "txt", "typ", "yml", ""], "Text"),
```
Kijewski, line 376: On the contrary, I would say that we should remove `md` and `yml` from the list.

GuillaumeGomez:
I see, didn't expect that characters needed to be escaped.

Kijewski:
I don't know a lot of typst, but a user could do "malicious" things like injecting images `#image(…)` to exfiltrate data. The other examples like `-` (`<li>`), `$` (enter math mode) are not that bad, but they could break formatting.

xabi00:
Those characters don't need to be handled by askama because they belong to the typst (or markdown) compiler. Also, the special characters used in typst don't conflict with the special characters in askama.

Kijewski:
> Those characters don't need to be handled by askama because they belong to the typst (or markdown) compiler. Also, the special characters used in typst don't conflict with the special characters in askama.

Then you could also say that `<>` don't need escaping in HTML, because they belong to the HTML parser. :-/

xabi00:
But html is handled different to text

Kijewski:
Yeah, but typst is not text. @GuillaumeGomez, do you see that differently?

Maybe I'm too pedantic, but IMHO that default is not safe.

GuillaumeGomez:
I consider it "fine". There is no conflict with jinja syntax so as such so in my opinion, the rest needs to be handled by users, not us.

xabi00:
I'm working on fixing the typos check and wanted to get your input. Do you prefer to ignore typos globally by adding 'typ' to the `extend-words` section in `typos.toml`, or would you rather use the `extend-ignore-re` to ignore just the specific lines related to extensions?

Kijewski:
> I consider it "fine". There is no conflict with jinja syntax so as such so in my opinion, the rest needs to be handled by users, not us.

Sorry, I veto this PR. If there was a conflict with jinja syntax was never even a question.

From a security perspective, this change is not fine, as it allows code injection. If we claim to handle typst files, then it is up to us to prevent that. That is exactly the same, without any difference at all, if auto-escape we chevrons in an HTML input `"<script>"`, or the hash in a typst input `"#image()"`.

For any format we claim to support, we need to provide a proper auto-escaper for the output of expressions. If we do not, then we cannot claim support for the format.

The user can always change [their configuration](https://askama.readthedocs.io/en/latest/configuration.html#escapers) to not do any escaping in `.typ` files, if they validated the input.

GuillaumeGomez:
That is a valid reason indeed.

xabi00:
I tried what @Kijewski proposed and it worked, so this solution is valid for me

GuillaumeGomez:
So I guess we can close this PR? We should also have an example on how to implement the `Escaper` trait.

Kijewski:
I'll open an issue to add an Escaper example.

PR #373, opened 2025-03-19
fix toml features field in the filters chapter of the book

bionicles:
the features field needs to be a list here for this code to be copypasta-ready

(extremely minor book typo fix)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #378, opened 2025-03-21
Revive `askama_escaped`

Kijewski:
We adopted `askama_escape` with the other askama crates. It was not updated for quite some time, but still gets 25k+ downloads / day Mon to Fri.

This PR re-adds the crate using our current HTML escaping function.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,7 +27,7 @@ jobs:
           set -eu
           for PKG in \
             bench-build examples/actix-web-app examples/axum-app examples/poem-app examples/rocket-app examples/salvo-app examples/warp-app fuzzing \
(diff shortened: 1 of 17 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Not sure it's really worth but I'll let it up to you.

Kijewski:
Well, [it](https://crates.io/crates/askama_escape) seems to be the most downloaded html escaping library, before [html-escape](https://crates.io/crates/html-escape) (which has a bigger scope, more functionality) or [v_htmlescape](https://crates.io/crates/v_htmlescape) (which is faster, but the code cannot be fathomed by the human mind).

Esp. after the multiple rounds of reviews in <https://github.com/askama-rs/askama/pull/351> I think it would be nice to provide the updates back to the community.

PR #384, opened 2025-03-31
`askama_derive` accidentally exposed as a feature

Kijewski:
If you don't use an optional dependency at least once with `dep:…` in a feature, it automatically, implicitly becomes a feature flag. The `dep:` prefix was missing from `askama_derive` for the `derive` feature.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #385, opened 2025-03-31
Track config files

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/383.

### DIFF
--- a/askama_derive/src/tests.rs
+++ b/askama_derive/src/tests.rs
@@ -22,7 +22,18 @@ fn build_template(ast: &syn::DeriveInput) -> Result<String, crate::CompileError>
 // the code we want to check.
 #[track_caller]
 fn compare(jinja: &str, expected: &str, fields: &[(&str, &str)], size_hint: usize) {
-    let generated = jinja_to_rust(jinja, fields).unwrap();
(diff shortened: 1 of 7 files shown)


Kijewski:
Thank you for immediately springing into action! :)

I moved the config path into Config itself instead of its key, I think it's a better fit. And I added a test.

Kijewski (approved):

GuillaumeGomez:
This was my very first implementation. :rofl: 

I didn't put it into `Config` because I thought it made more sense to have it into `ConfigKey`. But either way is fine for me.

askama_derive/src/tests.rs
```diff
@@ -1123,14 +1135,33 @@ fn test_concat() {
...
-            let actual = jinja_to_rust(&src, &[]).unwrap();
+            let actual = jinja_to_rust(&src, &[], "").unwrap();
             let actual = unparse(&actual);
             assert_eq!(expected, actual, "source: {:?}", src);
         }
     }
 }
+
+#[test]
+fn test_with_config() {
```
GuillaumeGomez, line 1151: Completely forgot to add a test, thanks!

GuillaumeGomez:
Thanks for the missing bits!

PR #387, opened 2025-04-03
Deterministic CodeGen

tom-ainc:
# Support  for non-cargo build systems

The use of a `HasherBuilder` that uses a random seed results in indeterminism in the code genera by `askama_derive`. This is fine for Cargo projects, however it causes issues with build systems (Buck2 in our case) that use a crate hashes.

# Solution

I've defined:

```rust
// askama_derive/lib.rs
type BuildHasherType = rustc_hash::FxSeededState;
const BUILD_HASHER = rustc_hash::FxSeededState::with_seed(1234); 
```

And updated all uses of `HashMap` in the crate to use the type alias & `BUILD_HASHER`, makes it easy to conditionally swap back to the random seed `HashBuilder` if you wanted to:

1. Gate the fixed seed hasher behind a feature or vice versa (although it shouldn't make a difference so I wouldn't bother),
2. Use a different hasher entirely if you feel like it...

# Draft

We're currently validating that this patch solves our problem, similar solutions have worked for us in the past for other dependencies, so should do.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -10,21 +10,21 @@ use std::sync::Arc;
 
 use parser::node::{Macro, Whitespace};
 use parser::{
-    CharLit, Expr, FloatKind, IntKind, MAX_RUST_KEYWORD_LEN, Num, RUST_KEYWORDS, StrLit, WithSpan,
+    CharLit, Expr, FloatKind, IntKind, Num, StrLit, WithSpan, MAX_RUST_KEYWORD_LEN, RUST_KEYWORDS,
 };
-use rustc_hash::FxBuildHasher;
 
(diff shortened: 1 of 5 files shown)


Kijewski:
I don't think that the use of our seeded caches makes the output indeterministic. E.g. the list of transcluded files is sorted before it gets printed out: https://github.com/askama-rs/askama/blob/6e987eb50dccb25081113ae92ed2376e0dc5a905/askama_derive/src/generator.rs#L155-L160

If you find an example that indeterministically generates code, then could you please post the differences between the generated sources?

The order in which transcluded files are read or their metadata is evaluated can be arbitrary with the current code, but that should not cause problems, or does it?

tom-ainc:
Just discovered this:

image

So this won't actually change anything... Closing the PR & will carry on looking...

PR #391, opened 2025-04-09
Implement `Values` on tuple

GuillaumeGomez:
When there is only one value, using a tuple seems preferrable. I found the need for it in https://github.com/rust-lang/docs.rs/pull/2795. :)


book/src/runtime.md
```diff
@@ -35,4 +35,21 @@ or by calling directly the `askama::get_value` function:
 ```
 
 If you try to retrieve a value with the wrong type or that you didn't set, you will get an
-`Err(ValueError)`.
+`Err(askama::Error::ValueType)` or a `Err(askama::Error::ValueMissing)`.
+
+Another example with a tuple:
```
Kijewski, line 40:
```suggestion
Another example with a key-value tuple:
```

Kijewski (approved):
Good idea!

Kijewski (approved):

GuillaumeGomez:
Sure, why not.

PR #392, opened 2025-04-10
If using local variable as value when creating a new variable, do not put it behind a reference

### LINKED ISSUE #386: Inconsistent (or incorrect) reference semantics of `{% let %}`
I've ran into this issue when migrating from 0.12.1. The new behavior doesn't seem right, and also I haven't found mentions of it in the 0.13 relnotes.

Consider this template:
```
{%- let a = 1 -%}
{%- if a == 1 %}A{% endif -%}
{%- let b = a -%}
{%- if b == 1 %}B{% endif -%}
{%- let c = a + 0 -%}
{%- if c == 1 %}C{% endif -%}
```

This is valid Askama 0.12 template, however with Askama 0.13 it fails to compile on the second comparison, because `b` is suddenly a reference. This would be the correct Askama 0.13 template instead:
```
{%- let a = 1 -%}
{%- if a == 1 %}A{% endif -%}
{%- let b = a -%}
{%- if *b == 1 %}B{% endif -%}
{%- let c = a + 0 -%}
{%- if c == 1 %}C{% endif -%}
```

At the very least the new behavior is inconsistent, as `let` seems to add a reference level depending on type of the expression, and the rules of that are unclear. If adding a reference was intended behavior, this is inconvenient, as it would require adding `|deref`s to most places where defined variable is used. I'm not sure how Askama works with references under the hood, but from the user perspective I expect `let` to work as an alias to the specified value/expression and have the same reference level, which may be modified explicitly with `&`/`*`/`ref`/`deref`.

GuillaumeGomez:
Fixes #386.

Nothing tricky here, just an extra check. However, it adds more tests to enforce it and added explanations about it in the book. :)

### DIFF
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -965,7 +965,11 @@ impl<'a> Generator<'a, '_> {
         }
 
         self.visit_target(buf, true, true, &l.var);
-        let (before, after) = if !is_copyable(val) {
+        // If it's not taking the ownership of a local variable or copyable, then we need to add
+        // a reference.
+        let (before, after) = if !matches!(**val, Expr::Var(name) if self.locals.get(name).is_some())
+            && !is_copyable(val)
+        {
             ("&(", ")")
         } else {
             ("", "")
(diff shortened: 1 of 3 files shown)


Kijewski (approved):
Thank you, esp. for adding a documentation!

PR #396, opened 2025-04-11
Bump version number to 0.13.1

Kijewski:
The parser had no changes, so I kept its version number.

### DIFF
--- a/askama_derive/Cargo.toml
+++ b/askama_derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "askama_derive"
-version = "0.13.0"
+version = "0.13.1"
 description = "Procedural macro package for Askama"
 homepage = "https://github.com/askama-rs/askama"
 repository = "https://github.com/askama-rs/askama"
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "askama"
-version = "0.13.0"
+version = "0.13.1"
 description = "Type-safe, compiled Jinja-like templates for Rust"
 keywords = ["markup", "template", "jinja2", "html"]
 categories = ["template-engine"]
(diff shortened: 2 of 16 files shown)


askama_derive/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "askama_derive"
-version = "0.13.0"
+version = "0.13.1"
```
GuillaumeGomez, line 3: The generator has a breaking change for the variables, so we can't make it part of a small release.

Kijewski: We always ignored semantic versioning in `askama_derive` and `askama_parser` to have the same version number for all related crates.

To make that possible `askama` pins to a specific release of `askama_derive` which pins to a specific release of `askama_parser`.

GuillaumeGomez: Well, if you make a minor version, the same code is still supposed to compile. It's not the case here, so we need to exclude this change from this release.

Kijewski: https://github.com/askama-rs/askama/pull/107#issuecomment-2262919494?

The change won't break older releases, because they were pinned to a specific generator release, too.

GuillaumeGomez: I'm confused. I wonder if we're talking about the same thing.

So let's check! If I understand you correctly, you're saying that as long as they don't update to `0.13.1`, it will build, right?

That's how I understood and hence why I answered that making a minor version update of a dependency shouldn't make compilation fail.

Now if it's not what you meant, I'd like extra explanations please. :)

Kijewski: Oops, I misunderstood you. :) Yeah, the borrow semantics were changed, which is a breaking change.

GuillaumeGomez: So instead I suggest to make a `0.13` branch with only non-breaking changes in it. Like that we can backport them and continue to make small releases. Does it sound good to you?

Kijewski: Yes, sounds good. :)

Kijewski:
I'll re-create the PR.

PR #397, opened 2025-04-11
Pass variables to sub-templates more reliably even if indirectly

Kijewski:


Kijewski:
Since we are doing a breaking release, I think this would be a good addition. It changes the signature of `askama::filters::FastWritable`, which is public.

askama/src/helpers.rs
```diff
@@ -267,9 +268,13 @@ impl<L: fmt::Display, R: fmt::Display> fmt::Display for Concat<L, R> {
 
 impl<L: FastWritable, R: FastWritable> FastWritable for Concat<L, R> {
     #[inline]
-    fn write_into<W: fmt::Write + ?Sized>(&self, dest: &mut W) -> crate::Result<()> {
-        self.0.write_into(dest)?;
-        self.1.write_into(dest)
+    fn write_into<W: fmt::Write + ?Sized>(
+        &self,
+        dest: &mut W,
+        values: &dyn Values,
```
GuillaumeGomez, line 274: Shouldn't we have a `write_into_with_values` instead?

testing/tests/values.rs
```diff
@@ -88,3 +88,23 @@ fn test_value_function_getter() {
     values.insert("a".to_string(), Box::new(false));
     assert_eq!(V.render_with_values(&values).unwrap(), "");
 }
+
+#[test]
+fn test_value_in_subtemplates() {
```
GuillaumeGomez, line 93: We should take the habit to explain what the test is checking. Do you mind adding a code comment please?

Kijewski: Yup, good call! Added.

Kijewski:
I removed the filter `|wordcount`. It's not (easily) possible change the filter to have access to runtime values, which makes it a usage hazard. And actually, I cannot think of a good usecase for this filter.

askama/src/filters/alloc.rs
```diff
@@ -376,39 +512,61 @@ pub fn trim<T: fmt::Display>(s: T) -> Result<String> {
...
+                prefix = format!("{: >1$}", "", width);
             }
+            dest.write_str(&prefix)?;
         }
-        Ok(indented)
+        dest.write_str(line)?;
     }
-    indent(format_args!("{s}"), width)
+    Ok(())
 }
```
Kijewski, line 570: `|indent` is entirely re-implement, please pay extra attention.

askama/src/filters/alloc.rs
```diff
@@ -500,32 +655,37 @@ pub fn wordcount(s: impl fmt::Display) -> Result<usize, fmt::Error> {
...
     }
-    Ok(output)
+}
+
+fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
+    for word in s.split_inclusive(' ') {
+        flush_capitalize(dest, word)?;
+    }
+    Ok(())
 }
```
Kijewski, line 689: `|title` is entirely re-implement, please pay extra attention.

askama/src/helpers.rs
Kijewski: Puh, I don't know. It would not be useful for the trait to have an implementation without `values`, so I think using this shorter name is fine.

GuillaumeGomez: No I meant: is there a case where there is no `values` and therefore we could have a method without this argument like we do for the other trait?

askama/src/filters/alloc.rs
```diff
@@ -454,31 +454,6 @@ pub fn capitalize(s: impl fmt::Display) -> Result<String, fmt::Error> {
     capitalize(try_to_str!(s => buffer))
 }
 
-/// Count the words in that string.
```
GuillaumeGomez, line 457: `wordcount` exists to get closer to parity with official `jinja`. I think we should keep it. Also, what would be the issue to have `values` passed as argument?

askama/src/helpers.rs
GuillaumeGomez: And now that I read more of the PR, the question kinda answers itself: it's needed because we always pass it to filters.

askama/src/filters/humansize.rs
```diff
@@ -33,14 +34,18 @@ pub struct FilesizeFormatFilter(f32);
...
+        Ok(self.write_into(f, NO_VALUES)?)
     }
 }
 
 impl FastWritable for FilesizeFormatFilter {
-    fn write_into<W: fmt::Write + ?Sized>(&self, dest: &mut W) -> crate::Result<()> {
+    fn write_into<W: fmt::Write + ?Sized>(
+        &self,
+        dest: &mut W,
+        _: &dyn Values,
```
GuillaumeGomez, line 45: Well, we could name this argument and pass it to `write_into` instead of having `NO_VALUES` repeated all over, no? Not sure if there is a benefit for this approach compared to the one you took though...

askama/src/filters/alloc.rs
```diff
@@ -118,14 +120,34 @@ pub fn format() {}
...
 #[inline]
-pub fn linebreaks(s: impl fmt::Display) -> Result<HtmlSafeOutput<String>, fmt::Error> {
-    fn linebreaks(s: &str) -> String {
-        let linebroken = s.replace("\n\n", "</p><p>").replace('\n', "<br/>");
-        alloc::format!("<p>{linebroken}</p>")
+pub fn linebreaks<S: fmt::Display>(source: S) -> Result<HtmlSafeOutput<Linebreaks<S>>, Infallible> {
+    Ok(HtmlSafeOutput(Linebreaks(source)))
+}
+
+pub struct Linebreaks<S>(S);
```
GuillaumeGomez, line 127: Just to be sure: this commit is just about optimizing the rendering of the filters modified here, right?

askama/src/filters/alloc.rs
```diff
@@ -376,39 +512,61 @@ pub fn trim<T: fmt::Display>(s: T) -> Result<String> {
...
+    if s.len() >= MAX_LEN {
+        return dest.write_str(s);
+    }
+
+    let mut prefix = String::new();
+    for (idx, line) in s.split_inclusive('\n').enumerate() {
+        if idx > 0 {
+            // only allocate prefix if needed, i.e. not for single-line outputs
+            if prefix.is_empty() {
+                prefix = format!("{: >1$}", "", width);
```
GuillaumeGomez, line 563: TIL about this formatting rule.

askama/src/filters/alloc.rs
GuillaumeGomez: Implementation looks fine to me, but it doesn't seem like we have tests for it so please add some. :)

askama/src/filters/alloc.rs
```diff
@@ -603,18 +788,6 @@ mod tests {
         assert_eq!(capitalize("ßß").unwrap().to_string(), "SSß".to_string());
     }
 
-    #[test]
-    fn test_title() {
```
GuillaumeGomez, line 641: Why removing this test?

askama/src/filters/alloc.rs
```diff
@@ -475,32 +655,37 @@ pub fn capitalize(s: impl fmt::Display) -> Result<String, fmt::Error> {
...
+    ) -> crate::Result<()> {
+        let mut buffer = String::new();
+        self.0.write_into(&mut buffer, values)?;
+        Ok(flush_title(dest, &buffer)?)
     }
-    Ok(output)
+}
+
+fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
+    for word in s.split_inclusive(' ') {
```
GuillaumeGomez, line 685: `' '` is a whitespace, but so is `'\n'` or `'\t'` and a few others. Take a look at the [is_whitespace](https://doc.rust-lang.org/nightly/std/primitive.char.html#method.is_whitespace) docs. 😉

askama/src/filters/alloc.rs
```diff
@@ -603,18 +788,6 @@ mod tests {
...
-    #[test]
-    fn test_title() {
-        assert_eq!(&title("").unwrap(), "");
-        assert_eq!(&title(" \n\t").unwrap(), " \n\t");
-        assert_eq!(&title("foo").unwrap(), "Foo");
-        assert_eq!(&title(" foo").unwrap(), " Foo");
-        assert_eq!(&title("foo bar").unwrap(), "Foo Bar");
-        assert_eq!(&title("foo  bar ").unwrap(), "Foo  Bar ");
-        assert_eq!(&title("fOO").unwrap(), "Foo");
-        assert_eq!(&title("fOo BaR").unwrap(), "Foo Bar");
```
GuillaumeGomez, line 649: Also we would need to add a few cases to ensure that `foo\nbar` would render `Foo\nBar`. :)

book/src/filters.md
```diff
@@ -520,6 +520,16 @@ fn main() {
 }
 ```
 
+### Runtime values
+[#runtime-values]: #runtime-values
+
+It is possible to access [runtime values](./runtime.html) in custom filter,
```
GuillaumeGomez, line 526:
```suggestion
It is possible to access [runtime values](./runtime.html) in a custom filter,
```

or:

```suggestion
It is possible to access [runtime values](./runtime.html) in custom filters,
```

GuillaumeGomez:
Wouldn't it make more sense to just provide the `Values` object to all filter call instead of doing it only for `FastWritable`?

askama/src/filters/alloc.rs
Kijewski: It has some in `filter_blocks` and `filter_blocks_ws`. Also, there is the doc test.

askama/src/filters/alloc.rs
Kijewski: Re-added and implemented with access to runtime values, though I'm not sure if anyone will ever use this filter. :)

askama/src/filters/humansize.rs
Kijewski: Yes, makes sense to simply pass the values along. The `write_into` implementations for u32 and &str don't use the value, but the optimizer knows that, too, and can kill the variable itself.

askama/src/filters/alloc.rs
Kijewski: Sorry, I don't understand the question.

askama/src/filters/alloc.rs
Kijewski: Oops, you're right. Changed to `char::is_whitespace`.

askama/src/filters/alloc.rs
Kijewski: Purely by accident.

askama/src/filters/alloc.rs
Kijewski: Right, added.

Kijewski:
> Wouldn't it make more sense to just provide the `Values` object to all filter call instead of doing it only for `FastWritable`?

I want custom filters to remain as simple to use as possible for the average case.

GuillaumeGomez:
> > Wouldn't it make more sense to just provide the `Values` object to all filter call instead of doing it only for `FastWritable`?
> 
> I want custom filters to remain as simple to use as possible for the average case.

Is it though? Erf, It'd be nice if we could have a survey for this or something... I have a few thousands followers on socials so we can actually do it now that I think about it. What do you think?

If you agree, then we need to have a short explanation on the two possibilities:

1. Pass `Values` to all filters
2. If you want to have access to `Values`, you need to return a type which implements `FastWritable`

Kijewski:
Sure, a poll would be great!

Kijewski:
Thinking again about it, no, simply passing the runtime values as a filter argument would solve nothing: The runtime values need to be passed on to the filter source, if it can accept this argument, and we still want to be compatible with any fmt::Display input.

GuillaumeGomez:
The idea I had was to force all filters to have a `Values` argument. Not sure why it wouldn't be compatible with `fmt::Display` though?

Kijewski:
I meant, even if we provide the argument to custom filters, the rest of the changes are still needed. I'm happy to review a different solution, though.

book/src/filters.md
```diff
@@ -535,6 +545,15 @@ fn main() {
 }
 ```
 
+### Runtime values
+[#runtime-values]: #runtime-values
+
+It is possible to access [runtime values](./runtime.html) in custom filters:
```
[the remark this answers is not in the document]
GuillaumeGomez: But an example would still be nice for using `Values` in filters!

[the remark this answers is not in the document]
Kijewski: What's wrong with the example? https://github.com/askama-rs/askama/blob/c6d45e1cdc0f93f24673f726c2df69e9e3205fbe/testing/tests/book_example_runtime_values_in_custom_filters.rs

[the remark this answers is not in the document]
GuillaumeGomez: This example is "advanced" (not sure how to word it better). Since now filters all receive the `Values` argument, they just need to make use of it, no need to return a type implementing `FastWritable`. "Simpler is better" yada yada, you know the drill. :rofl: 

However, it'd be nice to also add a chapter in the book about `FastWritable` (hence why I opened https://github.com/askama-rs/askama/issues/402).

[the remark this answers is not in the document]
GuillaumeGomez: Oh indeed. Sorry about that. ^^'

GuillaumeGomez (approved):

PR #401, opened 2025-04-15
 Implement `first` and `blank` arguments for `|indent`

Kijewski:
For feature parity with Jinja: <https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.indent>

Resolves <https://github.com/askama-rs/askama/issues/193>.

### DIFF
--- a/askama/src/filters/alloc.rs
+++ b/askama/src/filters/alloc.rs
@@ -1,6 +1,6 @@
 use alloc::borrow::Cow;
-use alloc::string::String;
 use alloc::str;
+use alloc::string::String;
 use core::cell::Cell;
 use core::convert::Infallible;
 use core::fmt::{self, Write};
@@ -507,7 +507,13 @@ fn flush_trim(dest: &mut (impl fmt::Write + ?Sized), collector: TrimCollector) -
     dest.write_str(collector.0.trim_end())
 }
 
-/// Indent lines with spaces or a prefix
+/// Indent lines with spaces or a prefix.
+///
+/// The first line and blank lines are not indented by default.
+/// The filter has two optional [`bool`] arguments, `first` and `blank`, that can be set to `true`
+/// to indent the first and blank lines, resp.
+///
+/// ### Example of `indent` with spaces
 ///
 /// ```
 /// # #[cfg(feature = "code-in-doc")] {
@@ -528,6 +534,8 @@ fn flush_trim(dest: &mut (impl fmt::Write + ?Sized), collector: TrimCollector) -
 /// # }
 /// ```
 ///
+/// ### Example of `indent` with prefix a custom prefix
+///
 /// ```
 /// # #[cfg(feature = "code-in-doc")] {
 /// # use askama::Template;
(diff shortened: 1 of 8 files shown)


GuillaumeGomez:
Gonna wait for #397 to be merged before reviewing this one. :wink:

Kijewski:
Rebased.

askama/src/filters/alloc.rs
```diff
@@ -776,6 +816,93 @@ fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
...
+
+#[cfg(feature = "alloc")]
+impl AsIndent for alloc::string::String {
+    #[inline]
+    fn as_indent(&self) -> &str {
+        self
+    }
+}
+
+impl AsIndent for usize {
```
GuillaumeGomez, line 847: Would be nice to support all unsigned, no? And also support signed integers too (with an extra check to ensure it's not negative).

askama/src/filters/alloc.rs
```diff
@@ -776,6 +816,93 @@ fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
...
+/// # use askama::filters::AsIndent;
+/// assert_eq!(4.as_indent(), "    ");
+/// assert_eq!(" -> ".as_indent(), " -> ");
+/// ```
+pub trait AsIndent {
+    /// Borrow `self` as prefix to use.
+    fn as_indent(&self) -> &str;
+}
+
+impl AsIndent for str {
```
GuillaumeGomez, line 832: `CowStr` I suppose too?

askama/src/filters/alloc.rs
Kijewski: Then you could not write a simple `2` to for two spaces, when the trait is implemented for signed and unsigned types.

askama/src/filters/alloc.rs
Kijewski: `Cow<'_, T>` is implemented further down.

askama/src/filters/alloc.rs
GuillaumeGomez: Faire enough!

askama/src/filters/alloc.rs
```diff
@@ -845,17 +976,49 @@ mod tests {
...
+            "hello\n    \n     bar"
+        );
+        assert_eq!(
+            indent("hello\n\n bar", 4, true, false).unwrap().to_string(),
+            "    hello\n\n     bar"
+        );
+        assert_eq!(
+            indent("hello\n\n bar", 4, true, true).unwrap().to_string(),
+            "    hello\n    \n     bar"
+        );
```
GuillaumeGomez, line 1021: Can you add test with string prefixes and basically all supported types in a separate test please?

GuillaumeGomez:
As a follow-up we could also allow to have named arguments in `indent`. :)

askama/src/filters/alloc.rs
Kijewski: Added.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

GuillaumeGomez:
Agreed!

PR #403, opened 2025-04-17
 Add named arguments for filters 

Kijewski:
Todo:
- [x] write tests
- [x] update book


Kijewski:
The book is not updated, yet, but the changes to the code can already be reviewed. The diff is rather big, and should be digested commit-by-commit, but I think the change is worth it.

askama_derive/src/generator/filter.rs
```diff
@@ -516,6 +515,60 @@ impl<'a> Generator<'a, '_> {
...
+
+    fn visit_filter_with_length(
+        &mut self,
+        ctx: &Context<'_>,
+        buf: &mut Buffer,
+        args: &[WithSpan<'a, Expr<'a>>],
+        node: Span<'_>,
+        name: &str,
+    ) -> Result<DisplayWrap, CompileError> {
+        ensure_filter_has_feature_alloc(ctx, name, node)?;
```
GuillaumeGomez: I don't think checking this here is a good idea. What happens if we add more filters that require `alloc` but don't have a `length` argument? Imo this is becoming too specific and not really a good idea for future developments.

askama_derive/src/generator/filter.rs
```diff
@@ -516,6 +515,60 @@ impl<'a> Generator<'a, '_> {
...
+        let [arg, length] = args else {
+            return Err(ctx.generate_error(
+                format_args!("filter `{name}` needs one argument, the `length`"),
```
GuillaumeGomez:
```suggestion
                format_args!("`{name}` filter needs one argument, the `length`"),
```

askama_derive/src/generator/filter.rs
Kijewski: All filters in `BUILTIN_FILTERS` actually need `alloc`. :D I added a commit to fix that: <https://github.com/askama-rs/askama/pull/403/commits/d14126f53628d55edcd30a1c84d3d860368b341f>

GuillaumeGomez: Oh convenient :D

testing/tests/named_filter_arguments.rs
```diff
@@ -0,0 +1,188 @@
+use askama::Template;
+
+#[test]
+fn test_pluralize_two_positional_args() {
```
GuillaumeGomez, line 4: So many missing code comments describing what the test is checking. :stuck_out_tongue:

Kijewski: The function name describes it, doesn't it?

GuillaumeGomez: Yeah true, nitpicking a bit too much here maybe. ^^'

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+
+    for (&arg, collected) in filter_args.iter().zip(collected_args) {
+        if !is_argument_placeholder(collected) {
+            continue;
+        } else if let Some(default) = arg.default_value {
+            *collected = default;
+        } else {
+            return Err(ctx.generate_error(
+                format_args!(
+                    "filter `{name}` argument `{}` is missing{}",
```
GuillaumeGomez, line 853: Missed this one:

```suggestion
                    "`{}` argument is missing when calling `{name}` filter{}",
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+            let Some(idx) = filter_args
+                .iter()
+                .enumerate()
+                .find_map(|(idx, arg)| (arg.name == arg_name).then_some(idx))
+            else {
+                return Err(ctx.generate_error(
+                    match filter_args.len() {
+                        1 => fmt_left!("filter `{name}` does not have any arguments"),
+                        _ => fmt_right!(
+                            "filter `{name}` does not have an argument `{}`; {}",
```
GuillaumeGomez, line 808: There is already a colon in `Display` impl of `ItsArgumentsAre`.

```suggestion
                            "filter `{name}` does not have an argument `{}`{}",
```

Kijewski: Oops, yes.

askama_derive/src/generator/filter.rs
Kijewski: Yes, we should always print the signature.

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+    }
+
+    Ok(())
+}
+
+struct ItsArgumentsAre(&'static [&'static FilterArgument]);
+
+impl fmt::Display for ItsArgumentsAre {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("; it's arguments are: (")?;
```
GuillaumeGomez, line 869: Missed this one too. ^^'

```suggestion
        f.write_str("; its arguments are: (")?;
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+    ctx: &Context<'_>,
+    name: &str,
+    args: &[WithSpan<'_, Expr<'_>>],
+    node: Span<'_>,
+) -> Result<(), CompileError> {
+    for arg in args {
+        if is_argument_placeholder(arg) {
+            return Err(ctx.generate_error(
+                format_args!(
+                    "filter `{}` cannot accept named arguments",
```
GuillaumeGomez, line 896:
```suggestion
                    "`{}` filter cannot accept named arguments",
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+            let Some(idx) = filter_args
+                .iter()
+                .enumerate()
+                .find_map(|(idx, arg)| (arg.name == arg_name).then_some(idx))
+            else {
+                return Err(ctx.generate_error(
+                    match filter_args.len() {
+                        1 => fmt_left!("filter `{name}` does not have any arguments"),
+                        _ => fmt_right!(
+                            "filter `{name}` does not have an argument `{}`{}",
```
GuillaumeGomez, line 808:
```suggestion
                        1 => fmt_left!("`{name}` filter does not have any arguments"),
                        _ => fmt_right!(
                            "`{name}` filter does not have an argument `{}`{}",
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+            arg_idx += 1;
+            (idx, arg)
+        };
+
+        let Some(collected_arg) = collected_args.get_mut(idx) else {
+            return Err(ctx.generate_error(
+                format_args!(
+                    "filter `{name}` accepts at most {} argument{}{}",
+                    filter_args.len() - 1,
+                    if filter_args.len() != 2 { "s" } else { "" },
```
GuillaumeGomez, line 828: The comparison to 2 made me bug for a while. XD

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+                    ItsArgumentsAre(filter_args),
+                ),
+                arg.span(),
+            ));
+        };
+        if !is_argument_placeholder(replace(collected_arg, value)) {
+            return Err(ctx.generate_error(
+                format_args!(
+                    "argument `{}` was already set",
```
GuillaumeGomez, line 837:
```suggestion
                    "`{}` argument was already set",
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+
+fn no_arguments<'a, 'b>(
+    ctx: &Context<'_>,
+    name: &str,
+    args: &'b [WithSpan<'a, Expr<'a>>],
+) -> Result<&'b WithSpan<'a, Expr<'a>>, CompileError> {
+    match args {
+        [arg] => Ok(arg),
+        [_, arg, ..] => Err(ctx.generate_error(
+            format_args!("filter `{name}` does not have any arguments"),
```
GuillaumeGomez, line 759:
```suggestion
            format_args!("`{name}` filter does not have any arguments"),
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+
+fn ensure_no_generics(
+    ctx: &Context<'_>,
+    name: &str,
+    node: Span<'_>,
+    generics: &[WithSpan<'_, TyGenerics<'_>>],
+) -> Result<(), CompileError> {
+    if !generics.is_empty() {
+        return Err(
+            ctx.generate_error(format_args!("unexpected generics on filter `{name}`"), node)
```
GuillaumeGomez, line 664:
```suggestion
            ctx.generate_error(format_args!("unexpected generics on `{name}` filter"), node)
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+            &FilterArgument {
+                name: "format",
+                default_value: None,
+            },
+        ];
+
+        ensure_filter_has_feature_alloc(ctx, "fmt", node)?;
+        let [source, fmt] = collect_filter_args(ctx, "fmt", node, args, ARGUMENTS)?;
+        let Expr::StrLit(ref fmt) = **fmt else {
+            return Err(ctx.generate_error(r#"use filter fmt like `value|fmt("{:?}")`"#, node));
```
GuillaumeGomez, line 536:
```suggestion
            return Err(ctx.generate_error(r#"use `fmt` filter like `value|fmt("{:?}")`"#, node));
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+                self.visit_str_lit(buf, fmt);
+                if args.len() > 1 {
+                    buf.write(',');
+                    self.visit_args(ctx, buf, &args[1..])?;
+                }
+                buf.write(')');
+                return Ok(DisplayWrap::Unwrapped);
+            }
+        }
+        Err(ctx.generate_error(r#"use filter format like `"a={} b={}"|format(a, b)`"#, node))
```
GuillaumeGomez, line 515:
```suggestion
        Err(ctx.generate_error(r#"use `format` filter like `"a={} b={}"|format(a, b)`"#, node))
```

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+    }
+
+    fn visit_fmt_filter(
```
GuillaumeGomez, line 518: Since this is the same code as `format`, shouldn't it be merged into one function and we only pass the filter name for the error messages?

askama_derive/src/generator/filter.rs
```diff
@@ -0,0 +1,921 @@
...
+                .iter()
+                .find_map(|(extensions, path)| {
+                    extensions
+                        .contains(&Cow::Borrowed(name))
+                        .then_some(path.as_ref())
+                })
+                .ok_or_else(|| {
+                    ctx.generate_error(
+                        format_args!(
+                            "invalid escaper '{name}' for `escape` filter. {}",
```
GuillaumeGomez, line 480:
```suggestion
                            "invalid escaper `{name}` for `escape` filter. {}",
```

book/src/filters.md
```diff
@@ -18,9 +18,29 @@ Enable it with Cargo features (see below for more information).
 ## Built-In Filters
 [#built-in-filters]: #built-in-filters
 
+Built-in filters that take (optional) arguments can be called with named arguments, too.
+The order of the named arguments does not matter, but named arguments must come after
+positional (i.e. unnamed) arguments.
+
+E.g. the filter `pluralize` takes two optional arguments: `sg` and `pl` (singular and plural),
```
GuillaumeGomez, line 25: Btw, why not using the full name instead of `sg`/`pl`?

askama_derive/src/generator/filter.rs
Kijewski: It's different from `format`. It's `{{ var | fmt("{:?}") }}` vs `{{ "{var1} {var2}" | format(var1 = 1, var2 = 2) }}`.

askama_derive/src/generator/filter.rs
Kijewski: Just imagine that rust's arrays are 1-based. :D

book/src/filters.md
Kijewski: The words are just too long, and `sg` and `pl` are universally understood abbreviations, I think.

GuillaumeGomez: It's the first time I ever saw them, so I'm definitely not convinced. ^^'

Kijewski: How about <https://en.wikipedia.org/wiki/List_of_glossing_abbreviations>, <https://en.wiktionary.org/wiki/singular>, <https://en.wiktionary.org/wiki/plural>, <https://en.wikipedia.org/wiki/Goose>?

GuillaumeGomez: Thanks!

GuillaumeGomez (approved):

PR #404, opened 2025-04-17
Deprecation messages for askama integration crates

Kijewski:
Many users don't depend on `askama` directly, but use e.g. `askama_axum` instead. They may never notice that a new askama version was released, because e.g. dependabot won't tell them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,11 @@
 [workspace]
-members = ["rinja", "rinja_derive", "rinja_parser"]
+members = [
+    "askama_actix",
+    "askama_axum",
+    "askama_rocket",
+    "askama_warp",
+    "rinja",
+    "rinja_derive",
+    "rinja_parser",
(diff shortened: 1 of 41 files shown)


GuillaumeGomez:
Shouldn't it be better to do a `0.0.x` release and add the deprecated mention there?

Kijewski:
No. If the new version number is less than the current number, then no-one will ever notice the upload. Crates.io shows the highest non-pre-release, not the last upload. Dependabot does not care about the upload order, either, only about the version number.

If the new release was version compatible to the current version, then we would needlessly break things for users who do a simple `cargo update`, so it has to be version incompatible.

It's exactly the same as for the rinja* v0.4 releases.

GuillaumeGomez (approved):

GuillaumeGomez:
Ah I see. I keep forgetting about that... All good then, thanks!

PR #405, opened 2025-04-18
Add `unique` filter

GuillaumeGomez:
Part of the missing jinja filters.

Implementation-wise, it's the best I could come up with.


Kijewski:
The filter must be added to the `BUILTIN_FILTERS` list, otherwise it won't be accessible. I am confused as to why the doc test of `fn unique()` passed? Could you please add a wrong assertion to it to test if it is actually invoked?

I didn't come up with a better implementation. I tried if it was possible to insert the element into a `HashMap<I::Item, ()>` and return a reference to the newly added key (→ [`VacantEntry::insert_entry()`](https://docs.rs/rustc-std-workspace-std/latest/std/collections/hash_map/struct.VacantEntry.html#method.insert_entry)), but I couldn't get the lifetimes to work. :(

askama/src/filters/std.rs
```diff
@@ -0,0 +1,61 @@
...
+        // To prevent cloning the data, we need to use `Rc`, like that we can clone `elem` as
+        // key of the `HashSet` and return it.
+        let elem = Rc::new(elem);
+
+        if set.insert(Rc::clone(&elem)) {
+            Some(elem)
+        } else {
+            None
+        }
+    })
```
Kijewski, line 36: If we increase the msrv to 1.83, then we could use `VacantEntry::insert_entry()`. This way the `Rc` only needs to be cloned if the key is actually absent from the list ↓. What do you think? Would it be worth it?

```rust
let mut set = HashMap::new();

it.into_iter().filter_map(move |elem| {
    // To prevent cloning the data, we need to use `Rc`, like that we can clone `elem` as
    // key of the `HashSet` and return it.
    if let Entry::Vacant(entry) = set.entry(Rc::new(elem)) {
        Some(Rc::clone(entry.insert_entry(()).key()))
    } else {
        None
    }
})
```

GuillaumeGomez:
> The filter must be added to the `BUILTIN_FILTERS` list, otherwise it won't be accessible. I am confused as to why the doc test of `fn unique()` passed? Could you please add a wrong assertion to it to test if it is actually invoked?
> 
> I didn't come up with a better implementation. I tried if it was possible to insert the element into a `HashMap<I::Item, ()>` and return a reference to the newly added key (→ [`VacantEntry::insert_entry()`](https://docs.rs/rustc-std-workspace-std/latest/std/collections/hash_map/struct.VacantEntry.html#method.insert_entry)), but I couldn't get the lifetimes to work. :(
> 
> (Also, it's odd that `HashMap` is in `std`, not `alloc`.)

I tried a lot of approaches, even to have a struct containing the hashmap and the iterator etc. But performance wise, it was always terrible hence why I used `Rc` in the end. With generators, this code will be much better. :3

And yep, it's odd but I think it's to keep only the minimum necessary inside `alloc` I guess?

askama/src/filters/std.rs
GuillaumeGomez: Huuuuuum... Ok let's go!

GuillaumeGomez:
It should fail CI this time.

GuillaumeGomez:
Seems like the version wasn't updated in the fuzzer but can't figure out where the rust version is defined for it... Any idea @Kijewski ?

Kijewski:
Hm, I didn't find a switch for that. Let's see if adding a "rust-toolchain.toml" works: <https://github.com/askama-rs/askama/pull/405/commits/27f7357c462cf0f04099d18dcb9bcd461b310e00>

Kijewski:
Looks like <https://github.com/google/oss-fuzz/pull/13237> worked! Needs rebasing because of <https://github.com/askama-rs/askama/pull/403>.

askama/src/filters/std.rs
```diff
@@ -0,0 +1,62 @@
...
+/// # use askama::{Template, filters};
+/// #[derive(Template)]
+/// #[template(ext = "html", source = "{% for elem in example|unique %}{{ elem }},{% endfor %}")]
+/// struct Example<'a> {
+///     example: Vec<&'a str>,
+/// }
+///
+/// assert_eq!(
+///     Example { example: vec!["a", "b", "a", "c"] }.to_string(),
+///     "a,b,c"
```
GuillaumeGomez, line 22: This test is supposed to fail btw.

Kijewski: Look like my usage of `cargo nextest` is wrong. Because of [you](https://blog.guillaume-gomez.fr/articles/2024-08-17+Doctests+-+How+were+they+improved%3F#How-they-were-improved) I think using `nextest` is not needed anymore. :) You could please try if replacing `cargo nextest run` with `cargo test` in the CI config causes a failure as it should?

GuillaumeGomez: Gladly. :D

.github/workflows/rust.yml
```diff
@@ -205,12 +205,9 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
-      - uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-nextest
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --all-targets --features full
-      - run: cargo nextest run --all-targets --no-tests=warn --features full
+      - run: cargo test run --all-targets --no-tests=warn --features full
```
Kijewski, line 210: `%s/cargo test run/cargo test/`

askama/src/filters/std.rs
jorgehermo9: Hi, sorry for jumping in.
nextest does not support doctests https://github.com/nextest-rs/nextest/issues/16

What people usually do is to run nextest as you did before in CI, and add another step just for doctests with `cargo test --doc`

PR #407, opened 2025-04-19
book: add page about `FastWritable`

Kijewski:
Resolves <https://github.com/askama-rs/askama/issues/402>.

### DIFF
--- a/book/src/performance.md
+++ b/book/src/performance.md
@@ -24,6 +24,27 @@ monomorphised code. On average, expect `.to_string()` to be 100% to 200% slower 
 [`.to_string()`]: <https://doc.rust-lang.org/stable/std/string/trait.ToString.html#tymethod.to_string>
 [`format!()`]: <https://doc.rust-lang.org/stable/std/fmt/fn.format.html>
 
+## Faster Rendering of Custom Types
+
+Every type that implements [`fmt::Display`] can be used in askama expressions: `{{ value }}`.
+Rendering with `fmt::Display` can be slow, though, because it uses [dynamic methods calls] in its
+[`fmt::Formatter`] argument. To speed up rendering (by a lot, actually),
+askama adds the trait [`FastWritable`]. For any custom type you want to render,
+it has to implement `fmt::Display`, but if it also implements `FastWritable`,
+then – using [autoref-based specialization] – the latter implementation is automatically preferred.
+
+To reduce the amount of code duplication, you can let your `fmt::Display` implementation call
+your `FastWritable` implementation:
+
+```rust
+{{#include ../../testing/tests/book_example_performance_fmt_call_fast_writable.rs}}
+```
+
+[`fmt::Display`]: <https://doc.rust-lang.org/stable/std/fmt/trait.Display.html>
+[`fmt::Formatter`]: <https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html>
+[`FastWritable`]: <./doc/askama/filters/trait.FastWritable.html>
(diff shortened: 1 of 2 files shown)


Kijewski:
Should we move `FastWritable` to the project root? It's not really related to filters. It's only in there, because `|escape` can be invoked explicitly, too.

GuillaumeGomez:
Would make more sense indeed.

book/src/performance.md
```diff
@@ -24,6 +24,26 @@ monomorphised code. On average, expect `.to_string()` to be 100% to 200% slower
 [`.to_string()`]: <https://doc.rust-lang.org/stable/std/string/trait.ToString.html#tymethod.to_string>
 [`format!()`]: <https://doc.rust-lang.org/stable/std/fmt/fn.format.html>
 
+## Faster Rendering of Custom Types
+
+Every type that implements [`fmt::Display`] can be used in askama expressions: `{{ value }}`.
+Rendering with `fmt::Display` can be slow, though, because is uses [dynamic methods calls] in its
```
GuillaumeGomez, line 30:
```suggestion
Rendering with `fmt::Display` can be slow, though, because it uses [dynamic methods calls] in its
```

book/src/performance.md
```diff
@@ -24,6 +24,26 @@ monomorphised code. On average, expect `.to_string()` to be 100% to 200% slower
...
+[`fmt::Formatter`] argument. To speed up rendering (by a lot, actually),
+askama adds the trait [`FastWritable`]. For any custom type you want to render,
+it has to implement `fmt::Display`, but if it also implements `FastWritable`,
```
GuillaumeGomez, line 33: Might be worth explaining how askama does that, what do you think?

testing/tests/book_example_performance_fmt_call_fast_writable.rs
```diff
@@ -0,0 +1,52 @@
...
+
+// In a real application, please have a look at
+// https://github.com/kdeldycke/awesome-falsehood/blob/690a070/readme.md#human-identity
+struct Name<'a> {
+    forename: &'a str,
+    surname: &'a str,
+}
+
+impl fmt::Display for Name<'_> {
+    // Because the method simply forwards the call, it should be `inline`.
```
GuillaumeGomez, line 14: Normally, the compiler is supposed to do that on its own, especially for small functions. So might be worth saying that normally it's not really necessary because rustc does it on its own but that it doesn't hurt to have it.

Kijewski: Not across crates units, or does it?

book/src/performance.md
Kijewski: Thanks!

book/src/performance.md
Kijewski: Makes sense! I added a link. [Autoref-based specialization](https://lukaskalbertodt.github.io/2019/12/05/generalized-autoref-based-specialization.html) is such a strange language feature that others should know it, too. :)

testing/tests/book_example_performance_fmt_call_fast_writable.rs
GuillaumeGomez: Hum, not sure about that...

book/src/performance.md
GuillaumeGomez: It's indeed a very nice feature! :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #413, opened 2025-04-20
Add throughput to derive benchmark

Kijewski:


GuillaumeGomez:
What is this PR doing btw?

Kijewski:
When you run `cargo bench` in `askama_derive_standalone`, then the change is not only expressed in `time / iterations`, but also in `len(input) / time`. It's more or less simply the reciprocal value, but I find it sometimes easier to grasp (bigger value = better).

GuillaumeGomez:
Ah I see. Good idea, thanks!

GuillaumeGomez (approved):

PR #415, opened 2025-04-22
Add `|titlecase` as alias for `|title`

Kijewski:
Resolves <https://github.com/askama-rs/askama/issues/408>.

### DIFF
--- a/askama/src/filters/alloc.rs
+++ b/askama/src/filters/alloc.rs
@@ -820,6 +820,31 @@ fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
     Ok(())
 }
 
+/// Return a title cased version of the value. Alias for the [`|title`](title) filter.
+///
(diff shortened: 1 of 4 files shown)


askama/src/filters/alloc.rs
```diff
@@ -820,6 +820,31 @@ fn flush_title(dest: &mut (impl fmt::Write + ?Sized), s: &str) -> fmt::Result {
     Ok(())
 }
 
+/// Return a title cased version of the value. Alias for the [`|title`](title) filter.
+///
+/// ```
+/// # #[cfg(feature = "code-in-doc")] {
```
GuillaumeGomez, line 826: I think we never run these doc tests because of this `cfg`.

Kijewski: Hm, you're right. I (hope I) fixed that in <https://github.com/askama-rs/askama/pull/416>. I cherry-picked this PR over there.

PR #417, opened 2025-04-22
Fix `unique` filter implementation

GuillaumeGomez:
I also removed the "code-in-doc" `cfg` from doctests to be sure that they are run all the time.

Also fixed the wrong implementation of the `unique` filter (it was supposed to return a `Result`).

### DIFF
--- a/askama/src/filters/std.rs
+++ b/askama/src/filters/std.rs
@@ -1,3 +1,4 @@
+use core::convert::Infallible;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use std::hash::Hash;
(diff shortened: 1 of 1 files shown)


Kijewski (changes requested):
`#[cfg(feature = "code-in-doc")] {` is needed for tests that use ```` ```jinja ```` block

GuillaumeGomez:
Why using `jinja` blocks btw?

Kijewski:
Because we had no filter tests whatsoever, and when they were added, they were added using this format.

GuillaumeGomez:
Do you mind if I migrate them to `source` instead?

Kijewski:
Yes, I do. The problem is not even related to `code-in-doc`, and the fix is readily available in <https://github.com/askama-rs/askama/pull/416>.

GuillaumeGomez:
My issue with the current doctests is that they require a feature to run, which is not great. I won't force if you prefer it this way but I'd rather having them run all the time if we run tests, even if the `code-in-doc` feature is not available.

GuillaumeGomez:
Just to confirm: do you want to keep them like this before I make the change? :smile:

GuillaumeGomez:
A debate for another day then. Reverting these changes then!

GuillaumeGomez:
Removed other changes.

Kijewski (approved):

PR #419, opened 2025-04-26
generator: remove invisible group from output

Kijewski:
The generated `TokenStream` gets parsed correctly by `rustc`, but fails to be parsed by `syn` in some circumstances. E.g. the template syntax `test {#` lets the playground crash.

This PR removes the invisible group.


GuillaumeGomez (dismissed):

Kijewski:
Fixed the linting errors. :) Needs to be re-approved.

GuillaumeGomez (approved):

PR #421, opened 2025-04-27
Add missing `#[inline]` annotations

Kijewski:
Manually selected from the list of warnings found with [`#![deny(clippy::missing_inline_in_public_items)]`](https://rust-lang.github.io/rust-clippy/master/#missing_inline_in_public_items)

### DIFF
--- a/askama/src/values.rs
+++ b/askama/src/values.rs
@@ -7,11 +7,15 @@ use crate::Error;
 pub const NO_VALUES: &dyn Values = &();
 
 /// Try to find `key` in `values` and then to convert it to `T`.
+#[inline]
 pub fn get_value<T: Any>(values: &dyn Values, key: impl AsRef<str>) -> Result<&T, Error> {
-    let Some(src) = values.get_value(key.as_ref()) else {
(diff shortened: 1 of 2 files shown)


askama/src/values.rs
```diff
@@ -7,11 +7,15 @@ use crate::Error;
...
-    let Some(src) = values.get_value(key.as_ref()) else {
-        return Err(Error::ValueMissing);
-    };
+    values
+        .get_value(key.as_ref())
+        .ok_or(Error::ValueMissing)
+        .and_then(convert_value)
+}
 
+fn convert_value<T: Any>(src: &dyn Any) -> Result<&T, Error> {
```
GuillaumeGomez, line 18: What's the advantage of creating this function btw?

Kijewski: The generated byte code of this function is huge, so I would not want to inline it.

GuillaumeGomez: :+1:

GuillaumeGomez:
Thanks!

PR #422, opened 2025-04-30
feat: add caller pattern from jinja 

pollend:
ref: https://github.com/askama-rs/askama/issues/282#issuecomment-2840581249

I adjusted this quite heavily given the discussion in issue 282. This both adds the caller macro pattern and changes the syntax to align closer to jinja.

`{% call super() %}`  is now `{{ super() }}`

https://jinja.palletsprojects.com/en/stable/templates/#call
https://jinja.palletsprojects.com/en/stable/templates/#child-template

```jinja
{% macro render_dialog(title, class='dialog') -%}
    <div class="{{ class }}">
        <h2>{{ title }}</h2>
        <div class="contents">
            {{ caller() }}
        </div>
    </div>
{%- endmacro %}

{% call render_dialog('Hello World') %}
    This is a simple dialog rendered by using a macro and
    a call block.
{% endcall %}

```

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -85,9 +85,9 @@ struct Generator<'a, 'h> {
     /// Used in blocks to check if we are inside a filter block.
     is_in_filter_block: usize,
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
-    seen_macros: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
-    /// Set of callers to forward into the macro.
-    current_caller: Option<&'a Call<'a>>,
+    seen_callers: Vec<(&'a Call<'a>, &'a Macro<'a>, Option<FileInfo<'a>>)>,
+    /// the active caller within the macro.
+    active_caller: Option<&'a Call<'a>>,
 }
 
 impl<'a, 'h> Generator<'a, 'h> {
@@ -112,8 +112,8 @@ impl<'a, 'h> Generator<'a, 'h> {
                 ..Default::default()
             },
             is_in_filter_block,
-            seen_macros: Vec::new(),
-            current_caller: None,
+            seen_callers: Vec::new(),
+            active_caller: None,
         }
     }
 
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -87,7 +87,7 @@ struct Generator<'a, 'h> {
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
     seen_macros: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
     /// Set of callers to forward into the macro.
-    seen_callers: Vec<&'a Call<'a>>,
+    current_caller: Option<&'a Call<'a>>,
 }
 
 impl<'a, 'h> Generator<'a, 'h> {
@@ -113,7 +113,7 @@ impl<'a, 'h> Generator<'a, 'h> {
             },
             is_in_filter_block,
             seen_macros: Vec::new(),
-            seen_callers: Vec::new(),
+            current_caller: None,
         }
     }
 
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -86,6 +86,8 @@ struct Generator<'a, 'h> {
     is_in_filter_block: usize,
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
     seen_macros: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
+    /// Set of called macros we are currently associated with a call.
+    seen_call_macro: Vec<&'a Macro<'a>>,
 }
 
 impl<'a, 'h> Generator<'a, 'h> {
@@ -111,6 +113,7 @@ impl<'a, 'h> Generator<'a, 'h> {
             },
             is_in_filter_block,
             seen_macros: Vec::new(),
+            seen_call_macro: Vec::new(),
         }
     }
 
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -9,7 +9,7 @@ use std::path::Path;
 use std::str;
 use std::sync::Arc;
 
-use parser::node::{Macro, Whitespace};
+use parser::node::{Call, Macro, Whitespace};
 use parser::{
     CharLit, Expr, FloatKind, IntKind, MAX_RUST_KEYWORD_LEN, Num, RUST_KEYWORDS, StrLit, WithSpan,
 };
@@ -86,8 +86,8 @@ struct Generator<'a, 'h> {
     is_in_filter_block: usize,
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
     seen_macros: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
-    /// Set of called macros we are currently associated with a call.
-    seen_call_macro: Vec<&'a Macro<'a>>,
+    /// Set of callers to forward into the macro.
+    seen_callers: Vec<&'a Call<'a>>,
 }
 
 impl<'a, 'h> Generator<'a, 'h> {
@@ -113,7 +113,7 @@ impl<'a, 'h> Generator<'a, 'h> {
             },
             is_in_filter_block,
             seen_macros: Vec::new(),
-            seen_call_macro: Vec::new(),
+            seen_callers: Vec::new(),
         }
     }
 
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -612,6 +612,7 @@ impl<'a> Generator<'a, '_> {
             scope,
             name,
             ref args,
+            ws2,
             ..
         } = **call;
 
@@ -776,7 +777,7 @@ impl<'a> Generator<'a, '_> {
             buf.write('}');
             Ok(size_hint)
         })?;
-        self.prepare_ws(ws1);
+        self.prepare_ws(ws2);
         self.seen_callers.pop();
         self.active_caller = None;
         Ok(size_hint)
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -1127,10 +1127,14 @@ impl<'a> Generator<'a, '_> {
                 ctx: &Context<'a>,
                 expected: usize,
                 found: usize,
+                name: &str,
             ) -> Result<(), CompileError> {
                 if expected != found {
                     Err(ctx.generate_error(
-                        format!("expected `{}` argument, found `{}`", expected, found),
+                        format!(
+                            "expected {expected} argument{} in `{name}`, found {found}",
+                            if expected != 1 { "s" } else { "" }
+                        ),
                         s.span(),
                     ))
                 } else {
@@ -1138,11 +1142,11 @@ impl<'a> Generator<'a, '_> {
                 }
             }
             if ***path == Expr::Var("super") {
-                check_num_args(s, ctx, 0, expr_args.len())?;
+                check_num_args(s, ctx, 0, expr_args.len(), "super")?;
                 return self.write_block(ctx, buf, None, ws, s.span());
             } else if ***path == Expr::Var("caller") {
                 let def = self.active_caller.ok_or_else(|| {
-                    ctx.generate_error(format_args!("block is not defined for caller"), s.span())
+                    ctx.generate_error(format_args!("block is not defined for `caller`"), s.span())
                 })?;
                 self.active_caller = None;
                 self.handle_ws(ws);
@@ -1151,7 +1155,7 @@ impl<'a> Generator<'a, '_> {
                     buf.write('{');
                     this.prepare_ws(def.ws1);
                     let mut value = Buffer::new();
-                    check_num_args(s, ctx, def.caller_args.len(), expr_args.len())?;
+                    check_num_args(s, ctx, def.caller_args.len(), expr_args.len(), "caller")?;
                     for (index, arg) in def.caller_args.iter().enumerate() {
                         match expr_args.get(index) {
                             Some(expr) => {
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -636,7 +636,11 @@ impl<'a> Generator<'a, '_> {
             (*def, ctx)
         };
 
-        if self.seen_callers.iter().any(|(_, s, _)| std::ptr::eq(*s, def)) {
+        if self
+            .seen_callers
+            .iter()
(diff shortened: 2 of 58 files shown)


pollend:
@GuillaumeGomez I think this should be it umm any clue about test coverage?

askama_derive/src/generator/node.rs
```diff
@@ -1092,17 +1091,108 @@ impl<'a> Generator<'a, '_> {
...
+            ..
+        } = &**s
+        {
+            if ***path == Expr::Var("super") {
+                return self.write_block(ctx, buf, None, ws, s.span());
+            } else if ***path == Expr::Var("caller") {
+                let def = self.seen_callers.last().ok_or_else(|| {
+                    ctx.generate_error(format_args!("block is not defined for caller"), s.span())
+                })?;
+                //self.flush_ws(ws); // Cannot handle_ws() here: whitespace from macro definition comes first
```
GuillaumeGomez, line 1113: Maybe remove the code and only leave the code comment?

askama_parser/src/node.rs
```diff
@@ -878,38 +878,85 @@ impl<'a> Import<'a> {
...
 
 impl<'a> Call<'a> {
     fn parse(i: &mut &'a str, s: &State<'_, '_>) -> ParseResult<'a, WithSpan<'a, Self>> {
-        let start = *i;
+        let start_s = *i;
+        let parameters = |i: &mut _| -> ParseResult<'_, Option<Vec<&str>>> {
+            let args = opt(preceded(
+                '(',
+                (
+                    opt(terminated(separated(0.., ws(identifier), ','), opt(','))),
```
GuillaumeGomez, line 897: This code doesn't handle literals like strings or integers or even tuples, etc. Alo, for `caller`, no need to support arguments I suppose, however `super` can have named arguments iiuc. So maybe call `Expr::arguments` instead?

book/src/template_syntax.md
```diff
@@ -914,9 +914,54 @@ Then if you don't pass a value for this argument, its default value will be used
...
 {# We specify both `arg1` and `arg2` so no default value is used #}
-{% call heading(1, 2) %}
+{% call heading(1, 2) %} {% endcall %}
+```
+
+### Call
+
+you can specify a block for a call to be passed down as used as apart of a macro. For this case 
+a block is denoted by `{% call(args) macro() %}`, ending with `{% endcall %}`. the callee is denoted by
+`{{caller()}}`.
```
GuillaumeGomez, line 926:
```suggestion
You can use the content in the call block directly inside the macro by using `{{ caller() }}`:
```

book/src/template_syntax.md
```diff
@@ -914,9 +914,54 @@ Then if you don't pass a value for this argument, its default value will be used
...
+{%- endmacro %}
+
+{% call() render_dialog('Hello World') %}
+    This is a simple dialog rendered by using a macro and
+    a call block.
+{% endcall %}
+
+```
+
+here is an example with a call block using arguments:
```
GuillaumeGomez, line 945:
```suggestion
Here is an example with a call block using arguments:
```

book/src/template_syntax.md
```diff
@@ -914,9 +914,54 @@ Then if you don't pass a value for this argument, its default value will be used
...
+
+{% call(user) dump_users(list_of_user) %}
+    <dl>
+        <dt>Realname</dt>
+        <dd>{{ user.realname }}</dd>
+        <dt>Description</dt>
+        <dd>{{ user.description }}</dd>
+    </dl>
+{% endcall %}
+
```
GuillaumeGomez, line 964:
```suggestion
{% endcall %}
```

GuillaumeGomez: Hum, weird suggestion... Well, please remove the empty line at the end of the block.

GuillaumeGomez (changes requested):
Seems like a very good and solid start! I think the syntax should allow to have default values etc, but can be done in a follow-up. For now though, please support more than just identifiers as arguments when calling it.

askama_parser/src/node.rs
pollend: what would the syntax look like you would only declare arguments to be received right? this is for the calle?

GuillaumeGomez: Oh right, sorry I thought it was about the call, not the definition. Please ignore this comment then. ^^'

testing/tests/calls.rs
```diff
@@ -106,6 +106,76 @@ mod test_double_attr_arg {
...
+}
+
+#[test]
+fn test_caller_args() {
+    #[derive(Template)]
+    #[template(
+        source = r#"
+{%- macro test() -%}
+{{-  caller("test") -}}
+{{-  caller("test") -}}
```
GuillaumeGomez, line 166: Maybe replace `"test"` with `1` for this one? Like that we cover integer literals as well. :)

pollend:
probably need a lot more test where to start?

GuillaumeGomez:
Hard to say. ^^'

I have a few ideas:

* What happens if you call `caller()` inside the `call` block?
* What happens if you use named arguments in `caller` like `caller(blue="a")`?
* Can you call a macro `caller`? (That should emit an error as it should not be allowed!)

pollend:
umm looks like i made a mistake somewhere

image

pollend:
the space handling was wrong didn't understand it the first pass on this. I made some more adjustments.

testing/tests/ui/caller_arguments.stderr
```diff
@@ -0,0 +1,59 @@
...
+   |
+5  |       source = r#"
+   |  ______________^
+6  | |     {% macro test() %}
+7  | |         {{- caller("a", "b") -}}
+8  | |     {%- endmacro -%}
+...  |
+11 | |     {%- endcall -%}
+12 | |     "#,
+   | |______^
```
pollend, line 14: this does not look quite correct it should be the call

askama_derive/src/generator.rs
```diff
@@ -87,7 +87,7 @@ struct Generator<'a, 'h> {
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
     seen_macros: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
     /// Set of callers to forward into the macro.
-    seen_callers: Vec<&'a Call<'a>>,
+    current_caller: Option<&'a Call<'a>>,
```
pollend, line 90: this is only scoped to the nested callee so this doesn't need to be a stack right. when I do call into another caller i store this onto the stack.

pollend:
having to wrap every macro in `{% call .. %}` and `{% endcall %}` seems kind of annoying. would be nice to support this syntax as well: https://jinja.palletsprojects.com/en/stable/templates/#import

pollend:
I added more test cases.

GuillaumeGomez:
> having to wrap every macro in `{% call .. %}` and `{% endcall %}` seems kind of annoying. would be nice to support this syntax as well: https://jinja.palletsprojects.com/en/stable/templates/#import we can do that later

It's somewhat supported. Can be revisited though.

book/src/template_syntax.md
```diff
@@ -914,9 +914,51 @@ Then if you don't pass a value for this argument, its default value will be used
...
```
GuillaumeGomez, line 961: Please remove this empty line.

book/src/template_syntax.md
```diff
@@ -839,15 +839,15 @@ You can then call it with `{% call name(args) %}`:
 
 {% endmacro %}
 
-{% call heading(s) %}
+{% call heading(s) %} {% endcall %}
```
GuillaumeGomez, line 842:
```suggestion
{% call heading(s) %}{% endcall %}
```

book/src/template_syntax.md
```diff
@@ -839,15 +839,15 @@ You can then call it with `{% call name(args) %}`:
...
 ```
 
 You can place macros in a separate file and use them in your templates by using `{% import %}`:
 
 ```jinja
 {%- import "macro.html" as scope -%}
 
-{% call scope::heading(s) %}
+{% call scope::heading(s) %} {% endcall %}
```
GuillaumeGomez, line 850:
```suggestion
{% call scope::heading(s) %}{% endcall %}
```

book/src/template_syntax.md
```diff
@@ -865,19 +865,19 @@ You can also specify arguments by their name (as defined in the macro):
 
 {% endmacro %}
 
-{% call heading(bold="something", arg="title") %}
+{% call heading(bold="something", arg="title") %} {% endcall%}
```
GuillaumeGomez, line 868:
```suggestion
{% call heading(bold="something", arg="title") %}{% endcall%}
```

book/src/template_syntax.md
```diff
@@ -914,9 +914,51 @@ Then if you don't pass a value for this argument, its default value will be used
 
 ```jinja
 {# We only specify `arg1`, so `arg2` will be "something" #}
-{% call heading(1) %}
+{% call heading(1) %} {% endcall %}
```
GuillaumeGomez, line 917:
```suggestion
{% call heading(1) %}{% endcall %}
```

book/src/template_syntax.md
```diff
@@ -914,9 +914,51 @@ Then if you don't pass a value for this argument, its default value will be used
...
 {# We specify both `arg1` and `arg2` so no default value is used #}
-{% call heading(1, 2) %}
+{% call heading(1, 2) %} {% endcall %}
```
GuillaumeGomez, line 919:
```suggestion
{% call heading(1, 2) %}{% endcall %}
```

testing/tests/calls.rs
```diff
@@ -106,6 +106,73 @@ mod test_double_attr_arg {
...
+{%- endmacro -%}
+{%- call() test() -%}
+nested
+{%- endcall -%}
+"#,
+        ext = "html"
+    )]
+    struct CallerEmpty {}
+    let x = CallerEmpty {};
+    assert_eq!(x.render().unwrap(), "nested");
```
GuillaumeGomez, line 125: Can you add the following test to ensure backlines are not removed around "nested" please?

```jinja
{%- macro test() -%}
{{- caller() -}}
{%- endmacro -%}
{%- call() test() %}
nested
{% endcall -%}
```

pollend: is this expected??

```
\nnested\n\n
```

askama_derive/src/generator/node.rs
```diff
@@ -1092,17 +1092,109 @@ impl<'a> Generator<'a, '_> {
...
+                                            normalized_arg
+                                        ));
+                                        this.locals
+                                            .insert_with_default(Cow::Borrowed(normalized_arg));
+                                    }
+                                }
+                            }
+                            None => {
+                                return Err(ctx.generate_error(
+                                    format_args!("missing `{arg}` argument"),
```
GuillaumeGomez, line 1168:
```suggestion
                                    format_args!("missing `{arg}` argument in `caller`"),
```

testing/tests/ui/caller_arguments.stderr
GuillaumeGomez: Strange indeed, seems like the span is created from the start of the arguments and not of `caller`.

testing/tests/calls.rs
GuillaumeGomez: No it's not.

GuillaumeGomez:
Still need to add tests I mentioned [here](https://github.com/askama-rs/askama/pull/422#issuecomment-2851313596).

testing/tests/calls.rs
pollend: umm i made a mistake somewhere :?

GuillaumeGomez: Likely. Good Luck with whitespace handling. :wink: 

If you can't figure it out, I can take a look later on.

pollend:
> Still need to add tests I mentioned [here](https://github.com/askama-rs/askama/pull/422#issuecomment-2851313596).

any clue what this test looks like i'm not quite following
- Can you call a macro caller? (That should emit an error as it should not be allowed!)

GuillaumeGomez:
Ah sorry, I meant a macro *named* `caller` (forgot a word).

pollend:
- if you have fewer arguments in the caller then the number passed it should it produce an error? 
- should we make caller/super reserved words that can't be used?
- I don't think i deal with named arguments i could just have it immediately throw an error or should i just add it.

GuillaumeGomez:
> if you have fewer arguments in the caller then the number passed it should it produce an error?

Yes, number of arguments should match.

> should we make caller/super reserved words that can't be used?

Yep!

> I don't think i deal with named arguments i could just have it immediately throw an error or should i just add it.

Can be added in a follow-up, but then please add an issue and a test which fails on it (with a code comment in the test linking to the issue).

pollend:
> > if you have fewer arguments in the caller then the number passed it should it produce an error?
> 
> Yes, number of arguments should match.
> 
> > should we make caller/super reserved words that can't be used?
> 
> Yep!
> 
> > I don't think i deal with named arguments i could just have it immediately throw an error or should i just add it.
> 
> Can be added in a follow-up, but then please add an issue and a test which fails on it (with a code comment in the test linking to the issue).

umm does Expr even supported named arguments?

testing/tests/calls.rs
pollend: yea ... can't quite figure it out

GuillaumeGomez: Just came back from conference. Give me some (random amount of) time to take a look at it and try to figure out what's wrong.

pollend: Thanks for letting me know. Conference? Sounds like fun did you present anything.

GuillaumeGomez: Nah, it was rustweek/rust all-hands. It was time to meet with my teams mates and to talk with a loooot of people. Social battery is empty now. :')

GuillaumeGomez:
Pushed the fix, it was very small: `endcall` whitespace information was `ws2`, not `ws1`. So at the end of the function, it was supposed to be using `ws2`, not `ws1`. That's pretty much it, you were very clone. Waiting for CI then merging. Thanks a lot for this PR!

GuillaumeGomez (dismissed):

GuillaumeGomez:
Ah woups, seems like a recent update broke things, fixing it.

GuillaumeGomez:
@Kijewski: The book build breakage is expected or is it a new dark magic bug? XD

pollend:
> @Kijewski: The book build breakage is expected or is it a new dark magic bug? XD

403 is a permission error

Kijewski:
I re-ran the test, and it works now. The download is retried up to 3 times, but I guess 1 second in between is a bit short.

The examples use macros, and need to be updated, too.

GuillaumeGomez (approved):

GuillaumeGomez:
CI is happy. Time to merge! Thanks a lot!

PR #423, opened 2025-04-30
chore: fix tiny wording error in "Calling functions" docs

m4tx:
Currently the doc says:

> If you only provide a function name, askama will assume it's a method. If you want to call a **method**, you will need to use a path instead:

The second "method" doesn't seem like it was intended, as it contradicts the first sentence; it makes much more sense that this was supposed to be a "function" here.


GuillaumeGomez:
Thanks!

PR #425, opened 2025-05-01
Add filters `|default`, `|assigned_or` and `|defined_or`

Kijewski:
Also, `enum Pluralize<S, P>` is renamed into `enum Either<L, R>` and exported.

Cc @joshka.

Resolves <https://github.com/askama-rs/askama/issues/424>.


askama_derive/src/generator/filter.rs
```diff
@@ -643,6 +706,8 @@ fn ensure_filter_has_feature_alloc(
     Ok(())
 }
 
+const FALSE: &WithSpan<'static, Expr<'static>> = &WithSpan::new_without_span(Expr::BoolLit(false));
```
GuillaumeGomez, line 709: Considering it's only used by `visit_default_filter`, maybe it'd be better to move it there?

Kijewski: It's also used in `visit_indent_filter`.

GuillaumeGomez (approved):

GuillaumeGomez:
Btw, no clue about a better for `DefaultUnwrappable` so I'm fine with it.

book/src/filters.md
```diff
@@ -78,6 +78,59 @@ Output:
...
+{{ greeting | default("Hi") }} 
+```
+
+then the output is the default value: `Hi`.
+In this variant, on the left-hand side of the filter must be a variable name.
+
+</li><li>
+
+```jinja
+{{ variable_or_expression | default(default_value, true) }}
```
joshka, line 110: This sort of parameter without a name makes it difficult to understand a template worth reading the template docs. I'd been inclined to either:
1. Replace a naked bool parameter with an enum that self documents (that may look verbose here), or
2. Replace a bool with a second named method. E.g. default_if_undefined might be a good name for the false branch. Alternatively default_if_none / err

joshka:
> Also, it would be possible to implement the trait for integers (testing if != 0), and Strings (testing if != "").

Empty string seems reasonable as that's rationally returned when a string is not entered somewhere, but most of the time a zero value seems like it would be only 0 if you're using that as a flag value, which is not very rusty, so I'd perhaps implement default_if_empty for strings (and str, cow perhaps), but not default_if_zero unless there's an explicit obvious use case where it makes sense.

book/src/filters.md
Kijewski: I added a line that `{{ variable_or_expression | default(default_value, boolean = true) }}` works, too.

Kijewski:
* Renamed trait to `DefaultFilterable` (I think that's better)
* Implemented for `Cow<'_, T>` (I forgot about taat, good call!)
* Implemented for `Wrapping<T>` & `Saturating<T>`
* Implemented for `NonZero*` (infallably `Ok(Some())`) [just so you don't have to update the code if you use NonZeroU32 and u32]

I think implementing the type for integers is fine and useful. In many real-world cases you will find `0` being used as a poor-man's `NULL`. And unless `boolean = true`, it does not matter either way. `DefaultFilterable` is not recursive: `Some(0)` returns `0`; only the `Option` is evaluated, not the contained value.

book/src/filters.md
joshka: I'd still prefer explicit over implicit a bit. This is based on an assumption that the direction of askama is a jinja-like template library with natural rust idioms rather than a jinja compatible template library that adds rust conveniences. I'm not sure whether that assumption is correct however (intuitively I'd lean towards the former as being more useful)

As a single data point. Having not seen the jinja syntax for the `default` filter prior, even with the parameter named `boolean` in the CoPilot suggestion rather than the bare `true`, I still had no idea from reading it what that parameter meant or controlled seeing it for the first time. This suggests to me that it is poorly named in jinja, and something that we could make better pretty easily with good naming here.

askama/src/filters/builtin.rs
```diff
@@ -330,13 +331,13 @@ impl<T: fmt::Display> fmt::Display for Center<T> {
 /// # }
 /// ```
 #[inline]
-pub fn pluralize<C, S, P>(count: C, singular: S, plural: P) -> Result<Pluralize<S, P>, C::Error>
+pub fn pluralize<C, S, P>(count: C, singular: S, plural: P) -> Result<Either<S, P>, C::Error>
```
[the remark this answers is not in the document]
GuillaumeGomez: Well, version bumping will be needed since we plan to extend `{% call %}` blocks. So it's fine.

askama/src/filters/builtin.rs
```diff
@@ -426,31 +427,270 @@ const _: () = {
...
+    /// e.g. a poisoned state or an unacquirable lock.
+    type Error: Into<crate::Error>;
+
+    /// Return the contained value, if a value was contained, and it's not the default value.
+    ///
+    /// Returns `Ok(None)` if the value could not be unwrapped.
+    fn as_filtered(&self) -> Result<Option<Self::Filtered<'_>>, Self::Error>;
+}
+
+const _: () = {
```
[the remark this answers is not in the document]
GuillaumeGomez: To prevent impacting the upper scope.

askama/src/filters/builtin.rs
```diff
@@ -426,31 +427,270 @@ const _: () = {
...
+            Self: 'a;
+
+        type Error = Infallible;
+
+        #[inline]
+        fn as_filtered(&self) -> Result<Option<&T>, Infallible> {
+            match self {
+                Ok(value) => Ok(Some(value)),
+                Err(_) => Ok(None),
+            }
```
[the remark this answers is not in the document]
GuillaumeGomez: Nice idea!

askama_derive/src/generator/filter.rs
```diff
@@ -627,6 +626,70 @@ impl<'a> Generator<'a, '_> {
...
+
+        let [value, default, boolean] = collect_filter_args(ctx, "default", node, args, ARGUMENTS)?;
+        let var_name = match **value {
+            Expr::Var(var_name) => Some(var_name),
+            _ => None,
+        };
+        let Expr::BoolLit(do_unwrap) = **boolean else {
+            return Err(ctx.generate_error(
+                format_args!(
+                    "the `default` filter takes a boolean literal as it's optional second argument"
```
GuillaumeGomez, line 657:
```suggestion
                    "the `default` filter takes a boolean literal as its optional second argument"
```

askama_derive/src/generator/filter.rs
```diff
@@ -627,6 +626,70 @@ impl<'a> Generator<'a, '_> {
...
+                ),
+                boolean.span(),
+            ));
+        };
+
+        if !do_unwrap {
+            let Some(var_name) = var_name else {
+                return Err(ctx.generate_error(
+                    format_args!(
+                        "the `default` filter requires a variable name as on its left-hand side"
```
GuillaumeGomez, line 667: "as on"?

askama_derive/src/generator/filter.rs
```diff
@@ -627,6 +626,70 @@ impl<'a> Generator<'a, '_> {
...
+        node: Span<'_>,
+    ) -> Result<DisplayWrap, CompileError> {
+        const ARGUMENTS: &[&FilterArgument; 3] = &[
+            FILTER_SOURCE,
+            &FilterArgument {
+                name: "default_value",
+                default_value: None,
+            },
+            &FilterArgument {
+                name: "boolean",
```
GuillaumeGomez, line 644: Why not naming it "do_unwrap" or something along the line to better match what it's doing instead?

askama/src/filters/builtin.rs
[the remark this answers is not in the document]
Kijewski: The type was unreachable before, so its name did not matter.

Kijewski:
So, what do we want to do with the argument "boolean"? It's the name and semantics Jinja uses. The question is: I compatible do we want to be?

If we don't care to be 100% compatible (and we don't), then yes, two filters would be the better option:
1. I would then call the `boolean = false` option `{{ var | defined_or(default_value) }}`.
2. And the `boolean = true` option `{{ expr | assigned_or(default_value) }}`.

Fun fact minijinja does not implement `|default` either.

askama/src/filters/builtin.rs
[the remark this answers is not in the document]
Kijewski: In here I mostly use it like a [`#region`](https://stackoverflow.com/questions/44390454/what-is-the-use-of-region-and-endregion-in-c). Every `DefaultFilterable` is in this scope and one click hides everything.

Kijewski:
Or, what about _both_ options?
* `{{ var | default(default_value, [false]) }}` → `{{ var | defined_or(default_value) }}`
* `{{ expr | default(default_value, [boolean =] true) }}` → `{{ expr | assigned_or(default_value) }}`

The filter `|default` will work as in jinja, same semantics, name argument naming; but you could also use filters with a (hopefully) better naming scheme.

GuillaumeGomez:
Having two filters for the same thing sounds good to me. :+1:

Just please improve the argument name. We can add an error for this specific case to tell that for better naming, `askama` is using `X` name instead for this named argument.

What do you think @joshka?

joshka:
Supporting ninja syntax probably makes it easier to bring templates from elsewhere (and allows LLMs to do fall into the pit of success). So while I dislike the syntax, if there's a better syntax available like suggested above, then I've no complaints about providing compatible syntax like this. I'd even go so far as to suggest that the parameter name is fine in that case (alternatively, if there's some future ability to alias it to give it an additional more descriptive name then that would work too)

Kijewski:
I added `|assigned_or` and `|defined_or` as (better) alternatives to `|defined`.

I don't think that it would be a good idea to implement `|defined` with other arguments than in Jinja.
* EITHER we keep it for Jinja compatibility exactly like in jinja
* OR we drop it, because the other two filters are better
* OR we re-implement the `|default` filter to `compile_error!` with a message to use the other filters.

joshka:
The other names are good.

I like the idea of making it possible to do something useful if someone tries jinja syntax. That makes a lot of sense as it makes it easy to fall into a pit of success.

But I also like the idea of making this a compile error, but wonder if it might be possible to cause a deprecation warning to be emitted for `default` instead? Are there cases where a template might want to need default|undefined type logic in askama? I.e. is there some case where a template would call another template but intentionally not provide some field?

Kijewski:
Blocked by <https://github.com/askama-rs/askama/pull/471>.

----

> Are there cases where a template might want to need default|undefined type logic in askama? I.e. is there some case where a template would call another template but intentionally not provide some field?

I could not think of a use case. In python, the `boolean` argument can be evaluated at runtime, but with our pre-compiled code this is not possible.

Right now, a proc-macro can only emit fatal errors, no warnings. There are open RFCs to add `compile_warning!`-like macros, though. But I guess we could always add a call to ↓ to fake a deprecation message.
```rust
#[deprecated(note = "use `|defined_or` or `|assigned_or` instead of `|default`")]
fn dont_use_default_default() {}
```
Most warnings that come from proc-macros get suppressed, though. I would need to test if deprecation message are one of them.

Kijewski:
We can't assign a span to the warning (→ <https://github.com/askama-rs/askama/issues/420>). Without the span, such a warning is kinda useless IMHO. I think the current implementation is fine as it is. :) We can always add a warning in a subsequent PR. @GuillaumeGomez, @joshka, what do you think? Good enough to be merged?

GuillaumeGomez:
Output looks ok to me.

askama_derive/src/generator/filter.rs
GuillaumeGomez: Question still open. :)

askama/src/filters/builtin.rs
```diff
@@ -426,31 +427,265 @@ const _: () = {
...
+
+/// A type (or a reference to it) that can be used in [`|assigned_or`](assigned_or).
+///
+/// The type is either a monad such as [`Option`] or [`Result`], or a type that has a well defined,
+/// trivial default value, e.g. an [empty](str::is_empty) [`str`] or `0` for integer types.
+#[diagnostic::on_unimplemented(
+    label = "`{Self}` is not `|assigned_or` filterable",
+    message = "`{Self}` is not `|assigned_or` filterable"
+)]
+pub trait DefaultFilterable {
```
joshka, line 451: It could be worth adding a unit tests for the various implementations of this that captures each implementation. E.g.:

```rust
assert_eq!(0_u8.as_filtered(), Ok(None));
assert_eq!(1_u8.as_filtered(), Ok(Some(1));
```

book/src/filters.md
```diff
@@ -27,6 +27,30 @@ which are `singular = ""` and `plural = "s"` by default.
...
+{{ variable_or_expression | assigned_or(fallback) }}
+```
+
+If the variable on the left-hand side is in its "default" state, e.g. an empty `""` string,
+a `0` integer, a `None` optional, or an `Err(_)` result, then the `fallback` value is printed.
+Otherwise the value.
+
+```jinja
+{% let greeting = Some("Hello") %}
+{{ greeting.as_ref() | default("Hi", true) }}
```
joshka, line 44: should be 
```suggestion
{{ greeting.as_ref() | assigned_or("Hi") }}
```

book/src/filters.md
```diff
@@ -27,6 +27,30 @@ which are `singular = ""` and `plural = "s"` by default.
 If you are fine with the default empty string for the singular, and you only want to set a
 specific plural, then you can call the filter like `dog{{ count | pluralize(plural = "gies") }}`.
 
+### assigned_or
+[#assigned_or]: #assigned_or
+
+```jinja
+{{ variable_or_expression | assigned_or(fallback) }}
+{{ variable_or_expression | assigned_or(fallback) }}
```
joshka, line 35: duplicate line - needs to be fixed or state what's going on here.

book/src/filters.md
```diff
@@ -78,6 +102,46 @@ Output:
...
+[#default]: #default
+
+```jinja
+{{ variable_or_expression | default(default_value) }}
+{{ variable_or_expression | default(default_value, [[boolean =] true]) }}
+```
+
+This filter works like the Jinja filter of the same name.
+If the second argument is not an boolean `true`, then the filter behaves like [`|defined_or`][#defined_or].
+If it is supplied and `true`, then the filter behaves like [`|assigned_or`][#assigned_or].
```
joshka, line 115: This should maybe describe how this works without having to link to the jinja / the other filters.

book/src/filters.md
```diff
@@ -78,6 +102,46 @@ Output:
...
+
+**This filter exists for compatibility with Jinja.
+You should use [`|defined_or`][#defined_or] and [`|assigned_or`][#assigned_or] directly instead.**
```
joshka, line 118:
```suggestion
Askama provides [`|defined_or`][#defined_or] and [`|assigned_or`][#assigned_or] which both better express the intention and should generally usually be used instead of this filter.**
```

testing/tests/ui/default.stderr
```diff
@@ -0,0 +1,88 @@
...
+  |
+6 | #[template(ext = "html", source = "{{ (1 + 1) | default(2) }}")]
+  |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: `default_value` argument is missing when calling `default` filter; its arguments are: (default_value, [boolean])
+ --> DefaultNoDefaultValue.html:1:3
+       "var | default }}"
+  --> tests/ui/default.rs:10:35
+   |
+10 | #[template(ext = "html", source = "{{ var | default }}")]
```
joshka, line 14: This actually suggests a bit of a jinja / rust impedance mismatch here. In Rust, we have a few different operations like `map_or_default` and `unwrap_or_default`. Intuitively, `var | default` when the type of var implements the `Default` trait seems like it would be obvious that it should return default. I know that a user could call `.unwrap_or_default()` or whatever there, but I'm curious whether it might be worth extending this to allow the obvious intuitive thing there. (I'm not fully certain of this, either way, WDYT?)

joshka (commented):
Some extra rview notes having gone through this in a bit more detail.

Kijewski:
@joshka, thank you for your suggestions and proofreading! I will have a look tomorrow.

Kijewski:
Took me a little longer than a day to look into it. 😳

askama_derive/src/generator/filter.rs
Kijewski: The filter only makes sense if it behaves the same as in jinja.

askama/src/filters/default.rs
```diff
@@ -0,0 +1,403 @@
...
+
+            #[inline]
+            fn as_filtered(&self) -> Result<Option<$ty>, Infallible> {
+                Ok(Some(self.get()))
+            }
+        }
+    )* };
+}
+
+impl_for_non_zero!(
```
GuillaumeGomez, line 230: Since we need to provide both the non-zero ty and the integer ty, why not merge this macro with `impl_for_int`? Like that: less code and one less macro definition/expansion.

askama/src/filters/default.rs
```diff
@@ -0,0 +1,403 @@
...
+            false => Ok(None),
+        }
+    }
+}
+
+/// An `f32` has a value if it is [`Normal`][FpCategory::Normal], i.e. it is not zero,
+/// not sub-normal, not infinite and not NaN.
+impl DefaultFilterable for f32 {
+    type Filtered<'a>
+        = Self
```
GuillaumeGomez, line 253: Wow this formatting is super weird. O.o

askama/src/filters/default.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    #[inline]
+    fn as_filtered(&self) -> Result<Option<Self::Filtered<'_>>, Self::Error> {
+        Ok((self.classify() == FpCategory::Normal).then_some(*self))
+    }
+}
+
+/// An `f64` has a value if it is [`Normal`](FpCategory::Normal), i.e. it is not zero,
+/// not sub-normal, not infinite and not NaN.
+impl DefaultFilterable for f64 {
```
GuillaumeGomez, line 267: Considering it's the same implementation as `f32` and that rust already has `f16` and `f128` in nightly, should we put this in a macro so we can more easily implement it on other floating types?

askama_derive/src/generator/filter.rs
GuillaumeGomez: That saddens me, but it's a valid point.

testing/tests/default.rs
```diff
@@ -0,0 +1,594 @@
...
+        }
+        .render()
+        .unwrap(),
+        "42"
+    );
+}
+
+#[test]
+fn test_default_with_str() {
+    // In this test a default value should be returned for an empty string.
```
GuillaumeGomez, line 398: Comments explaining what the test checks is very appreciated, thanks! :)

testing/tests/ui/default.stderr
```diff
@@ -0,0 +1,88 @@
...
+
+error: `fallback` argument is missing when calling `defined_or` filter; its arguments are: (fallback)
+ --> DefinedOrNoDefaultValue.html:1:7
+       "| defined_or }}"
+  --> tests/ui/default.rs:40:35
+   |
+40 | #[template(ext = "html", source = "{{ var | defined_or }}")]
+   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^
+
+error[E0277]: `Mutex<u32>` is not `|assigned_or` filterable
```
GuillaumeGomez, line 65: I didn't know we could do that in rust, it looks super nice! (and will look even better once I add the spans to the generated `TokenStream` :3)

GuillaumeGomez:
Looks all good to me, thanks!

testing/tests/ui/default.stderr
Kijewski: The feature is already stable for over a year: <https://releases.rs/docs/1.78.0/>, but I am not sure if I ever saw it in the wild. I found it [in the standard library](https://doc.rust-lang.org/1.88.0/src/core/future/future.rs.html#32) and simply tested if library developers can use it, too. :)

askama/src/filters/default.rs
Kijewski: Good idea! Made it a macro call.

testing/tests/ui/default.stderr
GuillaumeGomez: A lot of features are, but if I don't use them, I simply forget about them. ^^'

askama/src/filters/default.rs
Kijewski: Yeah, it's not the formatting I would have preferred, but rustfmt insists on adding a linebreak there. :)

joshka (approved):

PR #427, opened 2025-05-01
ci: remove zizmor test

Kijewski:
It was a small script to test common Github action config errors, but IMHO it became more opinionated with every release. No, I won't add a hashsum to every `uses` line, thank you for asking again and again.

Also, it takes 3 minutes to build by now. 3 minutes until the first real test is executed.


GuillaumeGomez (approved):

GuillaumeGomez:
That will make CI much faster. :)

PR #428, opened 2025-05-02
Use `either::Either`

Kijewski:
This PR lets us implement our traits for `either::Either` instead of implementing `Either` ourself in the form of `Pluralize`.

Probably any non-trivial web project will already (transitively) depend on `either`, be it through `writable`/`icu_*`/`fluent`, `bytes-utils`, `sqlx`, `reqwest` …

### DIFF
--- a/askama/Cargo.toml
+++ b/askama/Cargo.toml
@@ -24,6 +24,7 @@ name = "escape"
 harness = false
 
 [dependencies]
+either = { version = "1.15.0", default-features = false }
 itoa = "1.0.11"
 
 # needed by feature "derive"
@@ -61,6 +62,7 @@ serde_json = ["std", "askama_derive?/serde_json", "dep:serde", "dep:serde_json"]
 std = [
     "alloc",
     "askama_derive?/std",
+    "either/std",
     "serde?/std",
     "serde_json?/std",
     "percent-encoding?/std",
--- a/askama/src/either.rs
+++ b/askama/src/either.rs
@@ -1,0 +1,131 @@
+#[cfg(feature = "alloc")]
+use alloc::string::String;
+use core::any::Any;
+use core::fmt;
+
+use either::{Either, for_both};
+
+use crate::{Error, FastWritable, PrimitiveType, Template, Value, Values, filters};
+
+impl<L, R> FastWritable for Either<L, R>
+where
+    L: FastWritable,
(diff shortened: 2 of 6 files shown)


GuillaumeGomez:
Not sure to understand what's the benefit for us.

Kijewski:
Not re-inventing the wheel, and higher usability for users. An `Either` monad comes up often in a more functional-style programming. (Even the `Option` and `Result` monads are instances of `Either`, but obviously not implemented this way in Rust.)

An example in our code is [`EitherFmt`](https://github.com/askama-rs/askama/blob/73ba176ad5ac07380306b16e107c6ffd5db0232e/askama_derive/src/lib.rs#L549-L598). This is how `Either<L: Template, R: Template>` could come up in real-world code. Currently you would have to add an `enum` manually to implement this type.

And when we have two usages already (Pluralize and Template), then we can simply add the other traits, too.

GuillaumeGomez:
Again, I don't see how it benefits `askama`. `Option`/`Result`/`Cow` and other equivalents are all specialized because it makes reading the code simpler. In the current case, this PR achieves the opposite imo: more code and less readable.

Kijewski:
The first benefit is the removal of these lines: <https://github.com/askama-rs/askama/pull/428/files#diff-8fd031c47a492a35b50329d9b129a854d584ade4c1895379920d09388a5db99eL429-L457>. The second benefit is that it aids compatibility.

A user can already write ↓, but then the `SIZE_HINT` is wrong, and it's eight lines, when zero lines of boilerplate could have done the same:
```rust
#[derive(Template)]
#[template(ext = "html")]
enum EitherTemplate<L: Template, R: Template> {
    #[template(source = "{{self.0}}")]
    Left(L),
    #[template(source = "{{self.0}}")]
    Right(R),
}
```
I don't get how the change makes the code less readable.

GuillaumeGomez:
Sorry, I have a lot of questions, not a big user of "web" crates outside of docs.rs so not aware of much there. So what using `either` is enabling for end users exactly?

> Because it is invariantly `3` in this case. Any expression is `3` unless the generator knows better what the value is.

Fair enough. So basically, all enums implemented by `askama` directly get more accurate `SIZE_HINT`.

Kijewski:
Not pursued.

PR #430, opened 2025-05-02
Add new `reject` filter

GuillaumeGomez:
This one is interesting because whether you give it a path or a value, it will not call the same rust function. Found it interesting so implemented it.

### DIFF
--- a/askama/src/filters/builtin.rs
+++ b/askama/src/filters/builtin.rs
@@ -455,6 +455,63 @@ impl<S: FastWritable, P: FastWritable> FastWritable for Pluralize<S, P> {
     }
 }
 
+/// Returns an iterator without filtered out values.
+///
+/// ```
+/// # use askama::Template;
+/// #[derive(Template)]
+/// #[template(
+///       ext = "html",
+///       source = r#"{% for elem in strs|reject("a") %}{{ elem }},{% endfor %}"#,
+/// )]
+/// struct Example<'a> {
+///     strs: Vec<&'a str>,
+/// }
+///
+/// assert_eq!(
+///     Example { strs: vec!["a", "b", "c"] }.to_string(),
+///     "b,c,"
+/// );
+/// ```
+pub fn reject<T: PartialEq>(
+    it: impl IntoIterator<Item = T>,
+    filter: T,
+) -> Result<impl Iterator<Item = T>, Infallible> {
+    Ok(it.into_iter().filter(move |v| v != &filter))
+}
+
+/// Returns an iterator without filtered out values.
+///
+/// ```
+/// # use askama::Template;
+///
+/// fn is_odd(v: &&u32) -> bool {
+///     **v & 1 != 0
+/// }
+///
+/// #[derive(Template)]
+/// #[template(
+///       ext = "html",
+///       source = r#"{% for elem in numbers|reject(self::is_odd) %}{{ elem }},{% endfor %}"#,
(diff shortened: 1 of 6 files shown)


Kijewski:
Should we (try to) implement closures like `{{ x | reject(|x| *x % 2 != 0) }}` to make this filter even more useful?

askama/src/filters/builtin.rs
```diff
@@ -455,6 +455,63 @@ impl<S: FastWritable, P: FastWritable> FastWritable for Pluralize<S, P> {
...
+/// fn main() {
+///     assert_eq!(
+///         Example { numbers: vec![1, 2, 3, 4] }.to_string(),
+///         "2,4,"
+///     );
+/// }
+/// ```
+pub fn reject_with<T: PartialEq, F: Fn(&T) -> bool>(
+    it: impl IntoIterator<Item = T>,
+    callback: F,
```
Kijewski, line 510: Can you please make this `F: FnMut…`, `mut callback: F`?

askama/src/filters/builtin.rs
```diff
@@ -455,6 +455,63 @@ impl<S: FastWritable, P: FastWritable> FastWritable for Pluralize<S, P> {
...
+/// }
+///
+/// assert_eq!(
+///     Example { strs: vec!["a", "b", "c"] }.to_string(),
+///     "b,c,"
+/// );
+/// ```
+pub fn reject<T: PartialEq>(
+    it: impl IntoIterator<Item = T>,
+    filter: T,
```
Kijewski, line 478: Maybe better `filter: &T`, with an additional `&(…)` in the generated code. This way (I think) the type-coercion might work better if the input is already an `&T`.

GuillaumeGomez:
> Should we (try to) implement closures like `{{ x | reject(|x| *x % 2 != 0) }}` to make this filter even more useful?

I thought about it and concluded that we definitely didn't want to implement closures in our parser.

book/src/filters.md
```diff
@@ -467,6 +467,49 @@ will become:
...
+Output will be:
+
+```text
+2,3,
+```
+
+For more control over the filtering, you can use a callback instead. Declare a function:
+
+```jinja
+fn is_odd(value: &&u32) -> bool {
```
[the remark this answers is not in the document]
GuillaumeGomez: The `filter` method. Made me sad too.

askama/src/filters/builtin.rs
GuillaumeGomez: Anyway, added a second one.

Kijewski: No, that's not what I meant, but I confused myself with my comment. Let me think for a second or 86400. :)

Kijewski:
I tried to get rid of the second reference, but I failed. :( Still, I think that the changes I came up with during my tests could be useful. Please have a look.

GuillaumeGomez:
Looks good to me (can't approve if it's on my PR :disappointed:).

Kijewski (dismissed):
I can :)

GuillaumeGomez:
Show off. :stuck_out_tongue:

Kijewski (approved):

Kijewski:
Oops, the filter is not display wrapped. I don't think it would be matter, actually, because `impl Iterator` does not implement `Display` anyway, but better safe than sorry.

Issue #431, opened 2025-05-03
`Cow<'_, str>` doesn't implement `FastWritable`


C0D3-M4513R:
```
error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> app/src/rocket/content/admin/domain.rs:102:22
    |
102 |             content: Cow::Borrowed(VIEW_ADMIN_PANEL_DOMAIN_NO_PERM),
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `str`, which is required by `Cow<'_, str>: FastWritable`
    = help: the trait `FastWritable` is implemented for `Cow<'_, T>`
    = note: required for `Cow<'_, str>` to implement `FastWritable`
note: required by a bound in `domain_base::DomainBase`
   --> app/src/rocket/template/authenticated/domain_base.rs:7:30
    |
7   | pub struct DomainBase<'a, T: FastWritable> {
    |                              ^^^^^^^^^^^^ required by this bound in `DomainBase`
```

`DomainBase:`
```rust
use std::borrow::Cow;
use askama::{FastWritable, Template};
use crate::rocket::auth::session::Permission;

#[derive(Template)]
#[template(path = "authenticated/domain_base.html")]
pub struct DomainBase<'a, T: FastWritable> {
    pub domain: &'a str,
    pub permission: Option<Permission>,
    pub content: T,
}
```

The typebound in https://github.com/askama-rs/askama/blob/73ba176ad5ac07380306b16e107c6ffd5db0232e/askama/src/lib.rs#L418 should probably include `+ ?Sized`?

Edit: also, I just did realize, that I don't need to wrap the content in a `Cow`.
Nevertheless I feel like this should be addressed.

PR #432, opened 2025-05-03
Relax Sized constraint on `impl FastWritable for Cow`

C0D3-M4513R:
Fixes: #431

Please note, that I have not actually tested this commit.
This should however address the issue.

### DIFF
--- a/testing/tests/cow_str_implements_fast_writable.rs
+++ b/testing/tests/cow_str_implements_fast_writable.rs
@@ -3,16 +3,30 @@ use std::fmt::Display;
 
 use askama::{FastWritable, Template};
 
-// This tests, that Cow<'_, str> implements FastWritable.
-// If it didn't, compiling the tests would fail.
-const _: () = {
+#[test]
+fn test_cows() {
+    // This test ensures that Cow is FastWritable.
+    // Every expression needs to implement fmt::Display, even if the FastWritable path
+    // is going to be used.
+
     #[derive(Template)]
-    #[template(source = "{{content|safe}}", ext = "txt")]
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Please add test to ensure it works as expected. Then seems all good to me.

C0D3-M4513R:
I added a test, but for me the test-suite is currently in a failing state.
The ui test from askama_tests doesn't pass, as well as some other doc-tests.

Is that state expected?

Also for some reason escaping is broken? (see todo in the added "test")

testing/tests/cow_str_implements_fast_writable.rs
```diff
@@ -0,0 +1,33 @@
+use std::borrow::Cow;
+
+use askama::{FastWritable, Template};
+
+// This tests, that Cow<'_, str> implements FastWritable.
+// If it didn't, compiling the tests would fail.
+//
+// Since this test fails at compile-time, there isn't actually any need to run it.
+#[allow(dead_code)]
```
GuillaumeGomez, line 9:
```suggestion
#[test]
```

C0D3-M4513R: The test works compile time though. I could even just put it in a const block.

GuillaumeGomez: True, however it also allows us to ensure that it's actually built.

C0D3-M4513R: Why would the test not be built?
With cargo test --workspace it's built.

C0D3-M4513R:
Thank you for the better test code

PR #433, opened 2025-05-12
Ignore fuzzing corpus from source line count

Kijewski:
Github's language statistics for the project say that 43% of askama is assembly code. It is not. The heuristic counts our fuzzing corpus as assembly code, and the corpus is quite big.

This PR excludes the corpus from being counted.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #434, opened 2025-05-15
Make code generator re-usable for other projects

### LINKED ISSUE #377: Please make lib.rs/build_template public
I'am the author of TaiTan-Orm TaiTan-ORM,
And template support is the most important feature.
I formally use rinja template engine, and now use askama instead.

now I can implement a Template Query like this:
```rust
#[derive(askama::Template, Template, Debug)]
#[template(source="SELECT * FROM users WHERE name= :{name} {% if age.is_some() %}  AND age = :{age} {% endif %}", ext="txt")]
pub struct Query {
   name: String,
   age: Option<i64>
}
```
But it is ugly to write both askama::Template and Template,
And it wll be appreciate if there is any possible to make lib.rs/build_template public.

Kijewski:
This PR
* removes the crate `askama_derive_standalone`,
* makes `askama_derive` a normal library, and
* adds the proc-macro crate `askama_macros`,

Before, it was not possible for another crate to re-export `askama::Template` in a useful way, because the generated code assumes that it has access to an `extern crate askama`.

`askama_derive` will export the function `derive_template()` like `askama_derive_standalone` did, but it has an additional argument to accept a `TokenStream` that should contain (an) statement(s) to define the identifier `askama`, e.g. `quote! { extern crate askama; }`.

The new proc-macro crate `askama_macros` now defines the derive-macro `Template` by calling `askama_derive::derive_template()`.

Prior art: [`encase`] → [`encase_derive`] → [`encase_derive_impl`]; [2298a3e].

The PR is best viewed with ignored whitespace changes.

Resolves #377.

[`encase`]: <https://crates.io/crates/encase/0.11.0>
[`encase_derive`]: <https://crates.io/crates/encase_derive/0.11.0>
[`encase_derive_impl`]: <https://crates.io/crates/encase_derive_impl/0.11.0>
[2298a3e]: <https://github.com/teoxoy/encase/commit/2298a3efd5e4d86fee78572804c7aa5b4c806db0>


askama/src/lib.rs
```diff
@@ -81,10 +81,8 @@ use core::ops::Deref;
 use std::io;
 
 #[cfg(feature = "derive")]
-pub use askama_derive::Template;
+pub use askama_macros::Template;
 
-#[doc(hidden)]
-pub use crate as shared;
```
Kijewski, line 87: Not needed since askama v0.10 IIRC.

GuillaumeGomez (approved):

PR #435, opened 2025-05-15
Allow filters with explicit path

Kijewski:
IMHO that makes custom filters much easier to use.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice improvement!

PR #440, opened 2025-05-17
 Fuzz code generator, too 

Kijewski:
This PR adds a fuzzing target for `askama_derive`.

To make sure that the generator does not pull arbitrary files by means of e.g. `{% include %}`, this PR adds the feature `"external-sources"`, which is enabled by default. If it is disabled, the nodes `include`, `import`, `extends`, as well as `#[target(path = ..)]` result in a `CompileError`.

There is no seed corpus, yet. I'll generate it in our [fuzzing-corpus](https://github.com/askama-rs/fuzzing-corpus) after the PR gets merged.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -216,6 +216,7 @@ jobs:
       matrix:
         fuzz_target:
           - all
+          - derive
           - filters
           - html
           - parser
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -238,34 +238,34 @@ jobs:
         env:
           RUSTFLAGS: '-Ctarget-feature=-crt-static'
 
-  Cluster-Fuzz:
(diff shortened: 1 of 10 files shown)


Kijewski:
I assume this change will make the project fail on oss-fuzz, too, until the seed corpus is added: [build.sh#L22-L25](https://github.com/google/oss-fuzz/blob/de1a2503f846789e98c03a771ba65a3abfb37f82/projects/askama/build.sh#L22-L25). Well, I guess that's okay.

GuillaumeGomez (approved):

GuillaumeGomez:
We'll survive. :wink:

Kijewski:
Executing the fuzzer right now for 18 hours on 8 cores (on an ancient Opteron 2384 @ 2700Hz, though). Will update the corpus repo tomorrow evening, and add a PR to re-enable the cluster-fuzz ci job.

PR #444, opened 2025-05-18
derive: prefix more generated variables with `__askama`

Kijewski:
I want to say "prefix *all* generated variables with `__askama`", but I'm sure I overlooked some variables. :)

Rationale: We deny defining variables with this naming scheme, so this change makes sure that the user does not accidentally shadow a generated variable.

Squash before merging.


GuillaumeGomez (approved):

GuillaumeGomez:
Looks like a good idea! 👍

PR #446, opened 2025-05-19
Print invalid generated code during panic

Kijewski:


GuillaumeGomez:
Please add a UI test. :)

GuillaumeGomez:
Oh wait, might be tricky to do that. Hum... Is there a way to add such a test?

GuillaumeGomez (approved):

GuillaumeGomez:
Yeah that's what I figured out. Unless we voluntarily add a failing case in our code generator, not possible. Don't think it's worth it so nothing more we can do here. Very good idea, thanks!

PR #448, opened 2025-05-19
derive: don't suppress fatal errors in integer parsing

Kijewski:
Fixes <https://github.com/askama-rs/askama/issues/445>.

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -476,7 +476,7 @@ fn num_lit<'a>(i: &mut &'a str) -> ParseResult<'a, Num<'a>> {
         }
     };
 
-    let num = if let Ok(Some(num)) = opt(int_with_base.take()).parse_next(i) {
+    let num = if let Some(num) = opt(int_with_base.take()).parse_next(i)? {
         let suffix =
             opt(|i: &mut _| num_lit_suffix("integer", INTEGER_TYPES, start, i)).parse_next(i)?;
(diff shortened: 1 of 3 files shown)


askama_parser/src/lib.rs
```diff
@@ -476,7 +476,7 @@ fn num_lit<'a>(i: &mut &'a str) -> ParseResult<'a, Num<'a>> {
         }
     };
 
-    let num = if let Ok(Some(num)) = opt(int_with_base.take()).parse_next(i) {
+    let num = if let Some(num) = opt(int_with_base.take()).parse_next(i)? {
```
Kijewski, line 479: It took me like 2 hours to delete 4 characters and to insert 1 character. -_-

GuillaumeGomez: 2 hours well spent then. :)

askama_parser/src/tests.rs
```diff
@@ -1415,3 +1415,9 @@ fn underscore_is_an_identifier() {
     assert_eq!(result.unwrap(), "_");
     assert_eq!(input, "");
 }
+
+#[test]
+fn there_is_no_digit_two_in_a_binary_integer() {
```
GuillaumeGomez, line 1420: Should we add test cases for octal and hexa numbers as well?

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #450, opened 2025-05-19
Prevent `crate` to be used as identifier

### LINKED ISSUE #449: `crate` cannot be a raw identifier
Minimal example: `{{ crate }}` gets turned into `write!(f, "{}", self.r#crate)`, but `r#crate` is not a valid identifier.

> ### [Identifiers](https://doc.rust-lang.org/reference/identifiers.html#identifiers)
> 
> RAW_IDENTIFIER : `r#` IDENTIFIER_OR_KEYWORD *<sub>Except `crate`, `self`, `super`, `Self`</sup>*  

----

Fuzzing failure in `derive`.

Artifact:

```text
data:application/octet-steam;base64,///Le3sMS0szZTMzMTxjN319NjEvNjNtMzMzM3U3PGMwLn19BgAMDHt7Yy9jcmF0ZTwzMzg8Yzd9fTZ1bnNhZmUvNjNhM2FlDAwMJWV0WAD/ZmZmZmZmZmZmboxmZoxmZmZmZmbLSww=
```

Formatted test case:

```rust
use askama_derive::derive_template;
use quote::quote;

#[test]
fn test() -> Result<(), syn::Error> {
    let input = quote! {
        #[template(
            ext = "",
            source = "{{\u{c}KK3e331<c7}}61/63m3333u7<c0.}}\u{6}\0\u{c}\u{c}{{c/crate<338<c7}}6unsafe/63a3ae\u{c}\u{c}\u{c}%et"
//                                                                              ^^^^^ invalid token
        )]
        struct ffffn {}
    };
    let output = derive_template(input, import_askama);
    let _: syn::File = syn::parse2(output)?;
    Ok(())
}

fn import_askama() -> TokenStream {
    quote! {
        extern crate askama;
    }
}
```

Generated code:

```rust
match (
    &((&&askama::filters::AutoEscaper::new(&(KK3e331 < self.c7), askama::filters::Text)).askama_auto_escape()?),
    &((&&askama::filters::AutoEscaper::new(&(self.c / self.r#crate < 338 < self.c7), askama::filters::Text)).askama_auto_escape()?),
//                                                         ^^^^^^^ invalid token
) {
    (expr0,expr3,) => {
        (&&&askama::filters::Writable(expr0)).askama_write(__askama_writer, __askama_values)?;
        __askama_writer.write_str("61/63m3333u7<c0.}}    ")?;
        (&&&askama::filters::Writable(expr3)).askama_write(__askama_writer, __askama_values)?;
    }
}
__askama_writer.write_str("6unsafe/63a3ae   %et")?;
```

GuillaumeGomez:
Fixes #449.

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -39,6 +39,14 @@ fn check_expr<'a>(
     allow_underscore: bool,
 ) -> Result<(), ParseErr<'a>> {
     match &expr.inner {
+        // List can be found in rust compiler "can_be_raw" function (although in our case, it's also
+        // used in cases like `match`, so `self` is allowed in this case).
+        Expr::Var(name @ ("crate" | "super" | "Self")) => {
+            Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+                format!("`{name}` cannot be used as an identifier"),
+                expr.span,
+            )))
+        }
         Expr::Var("_") if !allow_underscore => Err(winnow::error::ErrMode::Cut(ErrorContext::new(
             "reserved keyword `_` cannot be used here",
             expr.span,
@@ -53,11 +61,21 @@ fn check_expr<'a>(
                 Ok(())
             }
         }
+        Expr::Path(path) => {
+            if let [name] = path.as_slice() {
+                if !crate::can_be_variable_name(name) {
+                    return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+                        format!("`{name}` cannot be used as an identifier"),
+                        expr.span,
+                    )));
+                }
+            }
+            Ok(())
+        }
         Expr::BoolLit(_)
         | Expr::NumLit(_, _)
         | Expr::StrLit(_)
         | Expr::CharLit(_)
-        | Expr::Path(_)
         | Expr::Attr(_, _)
         | Expr::Filter(_)
         | Expr::NamedArgument(_, _)
@@ -702,7 +720,17 @@ impl<'a> Suffix<'a> {
         preceded(
             ws(('.', not('.'))),
             cut_err((
-                alt((digit1, identifier)),
+                |i: &mut _| {
+                    let name = alt((digit1, identifier)).parse_next(i)?;
+                    if !crate::can_be_variable_name(name) {
+                        Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+                            format!("`{name}` cannot be used as an identifier"),
+                            *i,
+                        )))
+                    } else {
+                        Ok(name)
+                    }
+                },
                 opt(|i: &mut _| call_generics(i, level)),
             )),
         )
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -66,7 +66,7 @@ fn check_expr<'a>(
                 if !crate::can_be_variable_name(name) {
                     return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
                         format!("`{name}` cannot be used as an identifier"),
-                        expr.span,
+                        *name,
                     )));
                 }
             }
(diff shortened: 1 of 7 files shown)


askama_parser/src/expr.rs
```diff
@@ -39,6 +39,10 @@ fn check_expr<'a>(
     allow_underscore: bool,
 ) -> Result<(), ParseErr<'a>> {
     match &expr.inner {
+        Expr::Var("crate") => Err(winnow::error::ErrMode::Cut(ErrorContext::new(
```
Kijewski, line 42: "self", "super" and "Self" need to be rejected, too.

testing/tests/ui/crate_identifier.rs
```diff
@@ -0,0 +1,15 @@
...
+#[template(ext = "html", source = "{{ crate }}")]
+struct Example;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{% if crate == 12 %}{% endif %}")]
+struct Example2;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{% match crate %}{% endmatch %}")]
+struct Example3;
```
Kijewski, line 13: Can you add tests for `{% let crate %}`, `{% let crate = 12 %}` and `{{ self.crate }}`?

Kijewski:
Please add the found fuzzing failure, too:
```rust

#[test]
fn invalid_raw_identifier_crate() -> Result<(), syn::Error> {
    let input = quote! {
        #[template(
            ext = "",
            source = "{{\u{c}KK3e331<c7}}61/63m3333u7<c0.}}\u{6}\0\u{c}\u{c}{{c/crate<338<c7}}6unsafe/63a3ae\u{c}\u{c}\u{c}%et"
        )]
        struct ffffn {}
    };
    let output = derive_template(input, import_askama);
    let _: syn::File = syn::parse2(output)?;
    Ok(())
}
```

askama_parser/src/expr.rs
GuillaumeGomez: `super` too.

testing/tests/ui/crate_identifier.rs
GuillaumeGomez: About `self.crate` (or more precisely, any `.crate`): in this case, it means the user is actually doing a typo because the rust compiler doesn't allow to have a field named `crate`. So I suggest that we don't handle this case ourselves and hope that in the future, we'll have better interconnection between template source and generated rust code errors because otherwise we'll need to greatly improve our name checker (also in the `identifier` function which handles paths like `crate::whatever` where `crate` is valid in this position).

Kijewski: We have to handle this case. No [`proc_macro::Ident`](https://doc.rust-lang.org/stable/proc_macro/struct.Ident.html) can be called "crate", "self", … so the parsing of our generated code will fail.

testing/tests/ui/crate_identifier.rs
```diff
@@ -0,0 +1,63 @@
...
+#[template(ext = "html", source = "{% match Self %}{% endmatch %}")]
+struct Self3;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{% let Self %}")]
+struct Self4;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{% let Self = 12 %}")]
+struct Self5;
```
Kijewski, line 61: Could you please add tests for `{{ self.X }}` and `{{ variable.X }}` for *X* in ["self", "Self", "super", "crate"]?

GuillaumeGomez:
I think I covered all cases. :)

Kijewski:
Good idea to make this a problem of the parser, not the code generator!

I just noticed the `a::b::self` case, and added tests against that. If there are still problems, then the fuzzer will find them. :D

GuillaumeGomez:
Nice catch! (and thanks for the bugfix!) Your changes look good to me. =D

Kijewski (approved):
👍

PR #457, opened 2025-05-22
Comparison operators cannot be chained

Kijewski:
Fixes <https://github.com/askama-rs/askama/issues/454>.

### DIFF
--- a/testing/tests/ui/comparator-chaining.rs
+++ b/testing/tests/ui/comparator-chaining.rs
@@ -1,0 +1,49 @@
+// Comparison operators cannot be chained, so our parser must reject chained comparisons.
+
+use askama::Template;
+
+#[derive(Template)]
+#[template(ext = "txt", source = "{{ a == b != c }}")]
+struct EqNe {
+    a: usize,
+    b: usize,
+    c: usize,
+}
+
+#[derive(Template)]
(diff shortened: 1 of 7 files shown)


testing/tests/ui/comparator-chaining.rs
```diff
@@ -0,0 +1,47 @@
...
+#[template(ext = "txt", source = "{{ a == (b == (c == d == e)) }}")]
+struct ThreeTimesOk2 {
+    a: usize,
+    b: usize,
+    c: usize,
+    d: usize,
+    e: usize,
+}
+
+// Regression test for <https://github.com/askama-rs/askama/issues/454>
```
GuillaumeGomez, line 39: Please also add what this is testing. :)

Kijewski: I see no use in adding another comment besides the file name, the existing comment, and the expected outcome.

GuillaumeGomez: Thanks! Always better to have context, makes it more reliable in case files are renamed, etc.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #460, opened 2025-05-23
Correctly escape and reject macro names

Kijewski:
Partial fix for <https://github.com/askama-rs/askama/issues/459>.

* `crate!` etc are invalid macro names 
*  Escape macro names like `loop!`

Right now, `true` and `false` are never parsed as identifiers, so a macro `true!()` cannot be called from inside of a template. Actually, I think that's okay. Just don't call the macro that.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #462, opened 2025-05-25
Add support for mutable variables

GuillaumeGomez:
Part of https://github.com/askama-rs/askama/issues/452.

For now it only supports variables declared as mutable, not destructuring. Not sure if we want to go down this road or not but if we do, let's do it in a follow-up PR.

### DIFF
--- a/book/src/template_syntax.md
+++ b/book/src/template_syntax.md
@@ -63,6 +63,15 @@ Like Rust, Askama also supports shadowing variables.
 {{ foo }}
 ```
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Hum... I wonder if we shouldn't just keep the shadowing. If the new variable doesn't have the same type, it would fail compilation after all. Shadowing and updating the variable value are kinda the same no?

Kijewski (approved):

PR #463, opened 2025-05-26
Add issue templates

GuillaumeGomez:
It'll make it easier for us to go through issues thanks to some basic labeling.

It will also help people opening bug reports to provide the needed information.


Kijewski (approved):
Good idea!

PR #466, opened 2025-05-29
Refactor `indent` Filter Implementation

### LINKED ISSUE #465: `indent` filter fails to apply indentation for large input strings
**Description:**  
The `askama` `indent` filter does not apply the specified indentation (4 spaces) when the input string exceeds a certain length (e.g., 10,000 characters). Instead, the output string lacks the expected indentation.

**Steps to Reproduce:**  
1. Create a project with the following `Cargo.toml`:
   ```toml
   [package]
   name = "indent-test"
   version = "0.1.0"
   edition = "2024"

   [dependencies]
   askama = { version = "0.14.0", features = ["code-in-doc"] }
   ```

2. Use the following `lib.rs`:
   ```rust
   use askama::Template;

   /// ~~~askama
   /// {{ text | indent(4) }}
   /// ~~~
   #[derive(Template)]
   #[template(ext = "txt", in_doc = true)]
   pub struct IndentTest {
       text: String,
   }

   pub fn run(limit: usize) {
       let rendered = IndentTest {
           text: format!("a\n{}", "b".repeat(limit - 2)),
       }
       .render()
       .unwrap();

       assert_eq!(&rendered[0..10], "a\n    bbbb");
   }

   #[test]
   fn test_9999() {
       run(9999);
   }

   #[test]
   fn test_10000() {
       run(10000);
   }
   ```

3. Run `cargo test`.

Image

**Expected Behavior:**  
The `indent` filter should prepend 4 spaces to each line after the first, regardless of input size. For `test_10000`, the first 10 characters of the rendered output should be `"a\n    bbbb"`.

**Actual Behavior:**  
For inputs with 9,999 characters, the output is correct (`"a\n    bbbb"`). For 10,000 characters, the indentation is missing, resulting in `"a\nbbbbbbbb"`.

**Environment:**  
- Rust edition: v1.86.0
- Askama version: 0.14.0

strickczq:
## Description
This PR refactors the `indent` filter:
1. Rewrote `indent` filter to use `IndentWriter`, eliminating memory allocation, removing `MAX_LEN` restriction, and dropping the `fuzzed_indent_filter` test.
2. Moved `indent` filter from `alloc.rs` to a new `indent.rs` file, as it no longer requires memory allocation, and updated module imports/exports.

Special thanks to @Kijewski for their suggestion in issue #465 to rewrite the filter without intermediate allocations, which inspired this optimization.

## Changes
- Rewrote `indent` filter to use `IndentWriter`, removing memory allocation and `MAX_LEN`.
- Moved `indent` filter to `indent.rs`, removed unused dependencies (`Cow`, `Deref`, `Pin`) from `alloc.rs`, and updated `filters/mod.rs`.

## Testing
- Existing tests for `indent` filter pass.
- Removed `fuzzed_indent_filter` test as it's no longer relevant.

## Related Issues
- Fixes #465


GuillaumeGomez:
Solid start. Almost nothing to be changed, well done!

In addition to my comments, please also update the book "filters" chapter to remove the mention of the `alloc` feature on `indent` filter.

askama/src/filters/indent.rs
```diff
@@ -0,0 +1,355 @@
...
+}
+
+impl<W: fmt::Write> fmt::Write for IndentWriter<'_, W> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if self.indent.is_empty() {
+            return self.dest.write_str(s);
+        }
+
+        for (idx, line) in s.split_inclusive('\n').enumerate() {
+            if (self.first || idx > 0 || !self.is_new_line)
```
Kijewski, line 132: There is only one first line :), so `idx` would need to be tracked across calls. But in here we can simply removed the counter and set `self.first = true` at the end of the loop.

Kijewski (changes requested):
I only found this one logic problem, otherwise the PR looks gereat, thank you!

Kijewski (approved):
Thank you very much for fixing the problem! Good idea to add a test that splits up the input!

PR #468, opened 2025-06-01
parser: macro calls need proper tokens

Kijewski:
With this PR invalid tokens inside a macro call `macro_name!(..)` are rejected. Otherwise we might emit invalid code that cannot be parsed by rust. Also update fuzzing corpus to include derive.

Resolves <https://github.com/askama-rs/askama/issues/459>.  
Resolves <https://issues.oss-fuzz.com/issues/418896502>.

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -7,7 +7,7 @@ use winnow::combinator::{
     alt, cut_err, fail, not, opt, peek, preceded, repeat, separated, terminated,
 };
 use winnow::error::ParserError as _;
-use winnow::stream::Stream as _;
+use winnow::token::one_of;
 
 use crate::node::CondTest;
 use crate::{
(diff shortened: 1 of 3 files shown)


askama_parser/src/expr.rs
```diff
@@ -709,60 +709,118 @@ impl<'a> Suffix<'a> {
...
+                };
+                let close_token = match token {
+                    Token::SomeOther => continue,
+                    Token::Open(group) => {
+                        open_list.push(group);
+                        continue;
+                    }
+                    Token::Close(close_token) => close_token,
+                };
+                let open_token = open_list.pop().unwrap();
```
GuillaumeGomez, line 756: Are we sure there will always be at least one `open_list` item?

Kijewski: Yes, the vector is initialized with one element, and `open_list.is_empty()` is the break condition.

GuillaumeGomez:
Looks all good to me, thanks!

PR #469, opened 2025-06-01
bugfix: resolve with macro with call after caller

pollend:
https://github.com/askama-rs/askama/issues/467


Kijewski (approved):
Great, thank you fixing the bug!

// edit: oops, you weren't the reporter. Still thank you very much! :)

PR #471, opened 2025-06-01
ci: don't collect filter output into `String`

Kijewski:
We don't have to care if the output is big. If the user calls a filter with a big input, then they should know that the output will probably be even bigger.


GuillaumeGomez (approved):

GuillaumeGomez:
Ah nice. Should hopefully allow to have even more fuzzing. :)

PR #479, opened 2025-06-09
 parser: reject illegal string literals

Kijewski:
Resolves <https://github.com/askama-rs/askama/issues/478>.  
Resolves <https://issues.oss-fuzz.com/issues/422814961>.


askama_parser/src/expr.rs
```diff
@@ -788,51 +788,96 @@ impl<'a> Suffix<'a> {
...
                 _ => {
                     return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+                        format!("reserved prefix `{}#`", prefix.escape_debug()),
+                        prefix,
+                    )));
+                }
+            };
+
+            if opt('"').parse_next(i)?.is_some() {
+                // a got a raw string
```
GuillaumeGomez, line 821: nit:

```suggestion
                // got a raw string
```

GuillaumeGomez (commented):
Just a small nit but otherwise looks good to me, nice work!

GuillaumeGomez (approved):

PR #481, opened 2025-06-11
Rename `LoopItem::index` to `LoopItem::index0`

Kijewski:
That's what the element is called in the template syntax.

Also refactor `Generator::visit_attr()` at bit to make it more readable.  
Also add `__askama` prefix to `_len`.

### DIFF
--- a/askama_derive/src/generator/expr.rs
+++ b/askama_derive/src/generator/expr.rs
@@ -402,25 +402,22 @@ impl<'a> Generator<'a, '_> {
         obj: &WithSpan<'a, Expr<'a>>,
         attr: &Attr<'a>,
     ) -> Result<DisplayWrap, CompileError> {
-        if let Expr::Var(name) = **obj {
-            if name == "loop" {
-                if attr.name == "index" {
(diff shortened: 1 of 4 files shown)


askama_derive/src/generator/expr.rs
```diff
@@ -402,25 +402,22 @@ impl<'a> Generator<'a, '_> {
...
-                    return Ok(DisplayWrap::Unwrapped);
-                } else if attr.name == "first" {
-                    buf.write("(__askama_item.index == 0)");
-                    return Ok(DisplayWrap::Unwrapped);
-                } else if attr.name == "last" {
-                    buf.write("__askama_item.last");
-                    return Ok(DisplayWrap::Unwrapped);
-                } else {
-                    return Err(ctx.generate_error("unknown loop variable", obj.span()));
+        if let Expr::Var("loop") = **obj {
```
GuillaumeGomez, line 405: This code is so much better!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #485, opened 2025-06-13
parser: better error message for bare CRs

Kijewski:
@GuillaumeGomez, yeah, the message was not good, even if it is the "official" phrasing. What do you think about this longer message?


GuillaumeGomez (approved):

GuillaumeGomez:
I like it much better. :)

PR #487, opened 2025-06-16
parser: recognize comments in macro calls

Kijewski:
Fixes issue <https://issues.oss-fuzz.com/issues/425145246>


Kijewski:
I was too lazy to file an issue that explains the error found by the fuzzer.

Another option would have been to parse comments, but I don't see when you would need them, we have `{# comments #}` after all. Also, it would enable eye-sores like ↓ which I really don't want to see in actual use:
```jinja
{{ x!(// }}
) }}
```
:)

GuillaumeGomez:
I'm not sure it's a good idea, or at least not the right way to handle it. For bigger (rust) macros, it's very common to have code comments to explain what's going on. And for crates like `uniffi-rs`, which generates code, it could be a big blocker to not be able to have comments in macro calls. So if they write `a!(// haha)`, then it's their problem, not ours.

testing/tests/ui/comments-in-macro-calls.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Template)]
+#[template(ext = "html", source = "{{ e!(/** hello) }}")]
+struct OuterBlockDoc;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{{ e!(//! hello) }}")]
+struct InnerLineDoc;
+
+#[derive(Template)]
+#[template(ext = "html", source = "{{ e!(/*! hello) }}")]
```
GuillaumeGomez, line 26: What happens for this case:

```
{{ e!(//)
}}
```

?

GuillaumeGomez (approved):

testing/tests/ui/comments-in-macro-calls.rs
Kijewski: > error: expected `)` but found `}`

The line comment `//` make the parser skip the closing paren `)`, so the next character it sees is `}`, but it expects `)` to close the macro call.

I added tests to ensure that behavior (and found a typo in my implementation this way, oops).

PR #489, opened 2025-06-16
parser: implement `{% raw %}` without `skip_till`

Kijewski:
The implementation is longer than before, but also more correct: Before, the input `{% raw %}{% endraw }}{% endraw %}` would be a syntax error, now it renders the string `{% endraw }}`.


GuillaumeGomez (approved):

Kijewski:
The dependency is still there, just transitively now. :) I had a look: Our implementation was needed with nom, only winnow made it possible to scan for multiple literals in `skip_until`.

PR #496, opened 2025-06-21
parser: string literal must not be immediately followed by `suffix#`

Kijewski:
Resolves <https://issues.oss-fuzz.com/issues/426509683>.


askama_parser/src/lib.rs
```diff
@@ -680,9 +680,20 @@ fn str_lit<'a>(i: &mut &'a str) -> ParseResult<'a, StrLit<'a>> {
...
     }
 
+    not_suffix_with_hash(i)?;
     Ok(lit)
 }
 
+fn not_suffix_with_hash<'a>(i: &mut &'a str) -> ParseResult<'a, ()> {
+    if let Some(suffix) = opt((identifier, '#').take()).parse_next(i)? {
+        return Err(winnow::error::ErrMode::Cut(ErrorContext::new(
+            "a space is missing to tell suffix and prefix apart",
```
GuillaumeGomez, line 690: Took me a while to understand this error message. We should mention it's about a string literal and also give an example of what we suggest (between parens ``(like `r##"" ##`)``.

Kijewski: The problem is that we cannot really tell what the user wanted to write, only that somewhere a space is missing. Is it a suffixed string or a prefixed string? I hope that if the user sees the error message they know what they wanted to write. :)

GuillaumeGomez: Well, this message still needs to be improved. Open to suggestions. 😆

GuillaumeGomez:
Message is better so let's go. :)

PR #498, opened 2025-06-23
parser: introduce more types for `Expr` variants

Kijewski:
This makes the code prettier and `Expr` smaller.


GuillaumeGomez (approved):

GuillaumeGomez:
It's a good idea, thanks! Feel free to merge with or without the const assert (I can add it in a follow-up PR :wink:).

Kijewski:
Yes, good idea. At least it should not get bigger than it is now, I guess. :)

PR #503, opened 2025-06-29
Fix clippy warnings for rust 1.90

Kijewski:
This PR also cherry-picks #502. The stylized output still works with the new `console` version.


GuillaumeGomez:
Thanks!

Issue #507, opened 2025-06-30
Can not use imports from within call-block of another imported macro


seijikun:
**Describe the bug**
The call-block of an imported macro is unable to use the other imports - because instead of looking at the call-side's import context, it searches for the imports in the file that it was imported from.

**To Reproduce**
[repro.tar.gz](https://github.com/user-attachments/files/20979159/repro.tar.gz)

File `section.jinja`:

```jinja
{% macro render(title) %}
    <section>
        <h1>{{title}}</h1>
        <div>{{caller()}}</div>
    </section>
{% endmacro %}
```

File `element.jinja`:
```jinja
{% macro render() %}
    <div class="myElement">{{caller()}}</div>
{% endmacro %}
```

Root template (`page.jinja`):
```jinja
{% import "section.jinja" as section %}
{% import "element.jinja" as element %}

{% call section::render("sectiontitle") %}
    {# the element scope is not found because it's looking in section's import context #}
    {% call element::render() %}
        elementcontent
    {% endcall %}
    sectioncontent
{% endcall %}
```

**Workaround**
A dirty hack to workaround this, is to add the same `element` import in the `section.jinja` file.

**Askama version**
7610b37

**Rust version**
`rustc 1.90.0-nightly (b03b3a7ec 2025-06-26)`

PR #510, opened 2025-07-03
Fix #505, #507: Treat caller() as variable with support for aliasing

### LINKED ISSUE #505: Support storing caller within variable
Jinja doesn't support this pattern:

```jinja
{% macro inner %}
    inner before
        {{ caller() }}
    inner after
{% endmacro %}

{% macro outer %}
    outer before
    {% call inner() %}
        {# attempt to call outer caller() in inner caller(), which should be "my_awesome_content" #}
        {{ caller() }}
    {% endcall %}
    outer after
{% endmacro %}

{% call outer() %}
   my_awesome_content
{% endcall %}
```

This is [intentional](https://github.com/pallets/jinja/issues/371#issuecomment-271081314).
However, there is a [suggested workaround](https://github.com/pallets/jinja/issues/371#issuecomment-271081497) that looks like this:

```jinja
{% macro outer %}
    outer before
    {# declare variable storing outer caller #}
    {% set caller_ = caller %}
    {% call inner() %}
        {# use outer caller from declared variable #}
        {{ caller_() }}
    {% endcall %}
    outer after
{% endmacro %}
```

With askama, this does currently not work, because the generated code attempts to use the field `caller` in the template's struct - which obviously doesn't (and probably also shouldn't ?) exist. I assume this would need special handling.

Would it be possible to support this?

seijikun:
While implementing #505 over my previous version of fixing #507 (https://github.com/askama-rs/askama/pull/509), I noticed that there is a much better way which actually fixes both at the same time.

The `local` `MapChain` now keeps track of:
- variables (declaration, definition, with reference)
- caller aliases ("variables" assigned from either `caller` or another alias) - this contains the call-site definition, as well as the call site's `Context<'a>` as metadata
- Blockouts - This allows "hiding" a variable from a certain scope onward.

### Basic design:

When a `{% call ... %}` block is entered, I push `caller` as a `LocalMeta::CallerAlias` into the `local` map, together with the call-site `Context<'a>` and `Call<'a>` definition.

When a variable is assigned from another variable  (`Expr::Var(dst) = Expr::Var(src)`), where the `src` variable is a `local` of type `LocalMeta::CallerAlias`, the variable `dst` will also be inserted as caller alias by simply cloning `src`.

When a call expression like `{{ abc() }}` is encountered, whose variable is a `LocalMeta::CallerAlias`, we have found a `{{ caller() }}`, or any kind of: `{{ caller_alias() }}`. As soon as a new `local` scope was created for the call into `caller()`, a `LocalMeta::Negative` is inserted into the LocalMap. With this, the `caller` variable is blocked from access for this scope, and any number of scopes above - until a new entry for `caller` is inserted in one of the upper scopes. This prohibits callers from calling themselves - which is not supported. This was previously handled by tracking `active_caller` - but that was not prepared for multiple layers. Due to this restructuring, the error message changed a bit - but I like the new one better.


askama_derive/src/generator.rs
```diff
@@ -493,47 +490,100 @@ fn compile_time_escape<'a>(expr: &Expr<'a>, escaper: &str) -> Option<Writable<'a
...
             initialized: true,
-        }
+        })
     }
 
-    fn with_ref(refs: String) -> Self {
-        Self {
+    /// Variable referencing another
+    const fn var_with_ref(refs: String) -> Self {
+        Self::Variable(LocalVariableMeta {
```
GuillaumeGomez, line 537: Are we sure in this case the referenced variable was initialized?

askama_derive/src/generator/node.rs
```diff
@@ -931,6 +931,17 @@ impl<'a> Generator<'a, '_> {
             return Ok(());
         };
 
+        // Handle when this statement creates a new alias of a caller variable (or of another alias),
+        if let Target::Name(dstvar) = l.var {
+            if let Expr::Var(srcvar) = **val {
+                if let Some(caller_alias) = self.locals.get_caller(srcvar) {
```
GuillaumeGomez, line 937: Since 1.88, `let_chain` feature is now stable, so we can now write:

```suggestion
        if let Target::Name(dstvar) = l.var
            && let Expr::Var(srcvar) = **val
            && let Some(caller_alias) = self.locals.get_caller(srcvar)
        {
```

askama_derive/src/generator.rs
seijikun: That's a question I can't answer, since this is just a small refactoring to what [is there currently](https://github.com/askama-rs/askama/blob/a3f913af7029e8da224c425d7932dbe35c4f084c/askama_derive/src/generator.rs#L509). Or did you find a mistake in my refactoring?

askama_derive/src/generator/node.rs
seijikun: I specifically refrained from using let-chains because I didn't know which compiler version we were targeting. All the better :D

testing/tests/ui/caller_arguments.stderr
```diff
@@ -95,3 +57,27 @@ error: failed to parse template source
 100 | |     {%- endcall -%}
 101 | |     "#,
     | |______^
+
+error[E0599]: no method named `caller` found for reference `&CallerInCaller` in the current scope
```
GuillaumeGomez, line 61: That doesn't seem like the expected output. ^^'

askama_derive/src/generator/node.rs
GuillaumeGomez: You might need to update the MSRV and the version used in CI but otherwise, go for it!

seijikun: I have to change the rust edition to 2024 to use them, though. Is that okay?

askama_derive/src/generator.rs
GuillaumeGomez: No nevermind, maybe something we'll need to worry about later on. Since I was reviewing your code, I just thought about this case.

askama_derive/src/generator/node.rs
GuillaumeGomez: Yep, go for it! Next release will be a major one anyway so let's put as many things as possible haha.

testing/tests/ui/caller_arguments.stderr
seijikun: This is the error message by the official `jinja2` engine:
```
No caller defined
```

What would you suggest to do?

askama_derive/src/generator/node.rs
```diff
@@ -1105,21 +1116,27 @@ impl<'a> Generator<'a, '_> {
...
+
+            if let Some("super") = varname {
                 check_num_args(s, ctx, 0, v.args.len(), "super")?;
                 return self.write_block(ctx, buf, None, ws, s.span());
-            } else if *v.path == Expr::Var("caller") {
-                let def = self.active_caller.ok_or_else(|| {
-                    ctx.generate_error(format_args!("block is not defined for `caller`"), s.span())
-                })?;
-                self.active_caller = None;
+            } else if let Some(LocalCallerMeta { call_ctx, def }) = caller_alias.cloned() {
```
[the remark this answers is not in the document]
GuillaumeGomez: Ah I see. Then nevermind!

seijikun:
I hope I got everything. I now explicitly generate a `No caller defined` error when the `caller` keyword was used without any `caller` variable in scope.

askama_derive/src/generator/node.rs
```diff
@@ -1105,21 +1115,30 @@ impl<'a> Generator<'a, '_> {
...
                 check_num_args(s, ctx, 0, v.args.len(), "super")?;
                 return self.write_block(ctx, buf, None, ws, s.span());
-            } else if *v.path == Expr::Var("caller") {
-                let def = self.active_caller.ok_or_else(|| {
-                    ctx.generate_error(format_args!("block is not defined for `caller`"), s.span())
-                })?;
-                self.active_caller = None;
+            } else if let Some("caller") = var_name && caller_alias.is_none() {
+                // attempted to use keyword `caller` - but no caller is currently in scope
+                return Err(ctx.generate_error("No caller defined", s.span()));
```
GuillaumeGomez, line 1129: Please don't update this message, original one is better. ^^'

seijikun:
Should I bump the cargo version in github workflows as part of this PR, or separately?

GuillaumeGomez:
You also need to update the version used in CI.

askama_derive/Cargo.toml
```diff
@@ -6,8 +6,8 @@ homepage = "https://github.com/askama-rs/askama"
 repository = "https://github.com/askama-rs/askama"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
-edition = "2021"
-rust-version = "1.83"
+edition = "2024"
+rust-version = "1.88"
```
GuillaumeGomez, line 10: Some more files need to be updated:

```
git grep -n '1\.83\.'
.github/workflows/rust.yml:205:          toolchain: "1.83.0"
book/src/upgrading.md:9:* The <abbr title="minimum supported rust version">MSRV</abbr> of this release is 1.83.
clippy.toml:1:msrv = "1.83.0"
```

Almost there. :laughing:

seijikun:
Is the Cluster-Fuzz error something I have to fix?

GuillaumeGomez:
Sure!

Kijewski (dismissed):
Thank you very much!

GuillaumeGomez:
So just waiting for the oss-fuzz to be merged then we can restart CI here and merge. :)

GuillaumeGomez (approved):

PR #515, opened 2025-07-04
parser: implement 'Default' for '&Syntax'

Kijewski:
I noticed that you can implement `Default` for references, too. I think that's neat.

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -918,7 +918,7 @@ impl State<'_, '_> {
     }
 }
 
-#[derive(Default, Hash, PartialEq, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Overall I don't think it's an improvement: the issue with `Default` is that unless you have type annotations (or that you an RLS), you don't know what type the `Default::default()` is actually.

Kijewski:
The main idea was that only 1 word needs to be spilled to the stack when you use `Syntax` instead of 12. But I guess that's a too minor optimization to warrant the less readable code. I reduced the PR to only add the `impl`s, without actually using them. Just as an additional option if someone wants to use them. :)

Kijewski:
Well, that's a good question actually. :D

Issue #517, opened 2025-07-04
Support expression-call syntax for macros


seijikun:
Would be nice to have support for the expression-call based syntax for invoking macros:

```jinja
{% import "other.jinja" as other %}

{% macro test(title) %}
    stuff
{% endmacro %}

{{ test() }}
{{ other::render() }}
```

PR #518, opened 2025-07-04
Add support expression-call syntax for macros

seijikun:
Depends on (and contains): #510
Fixes: #517


seijikun:
Okay, I think I'm ready for review with this change.
The diff shows a lot of changed lines, but not many of them actually changed - I just moved them around a bit.

I added a `helpers` module for the generator, where complex/reusable parts of the code generation logic can be moved into - especially if they are needed multiple times. Macro invocations are a prime example of this, because it involves a **lot** of stuff. This reduces the function/file size in generator a bit. Though I had to make some generator methods `pub(crate)` for it to work.
(I initially tried to do it with functions inside generator - but `clippy` said `9` function parameters is too much :sweat_smile:  - so the change was not entirely voluntary, but I like it)

After this major refactoring, the implementation of #517 was very straightforward.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

Considering how big it is, I'll let @Kijewski take a look as well as I'm sure I missed things.

@Kijewski btw, little advice: read the PR commit by commit, much simpler (my first read was horrible :laughing: )

GuillaumeGomez (dismissed):

seijikun:
Rebased to latest changes.

GuillaumeGomez:
Thanks to both of you!

PR #522, opened 2025-07-06
Add support for macro default value generics

### LINKED ISSUE #513: Expressions currently don't support generics in paths
**Describe the bug**
Currently, the parser does not allow generic parameters in expressions / paths.

**To Reproduce**
```jinja
{% macro test(title = Option::<&'static str>::None) %}
    {% if let Some(title) = title %}
        Title: {{title}}
    {% endif %}
{% endmacro %}

{% call test() %}{% endcall %}
```
This fails to parse with:
```
error: failed to parse template source
         --> templates/page.jinja:1:31
       "&'static str>::None) %}\n    {% if let Some(title) = title %}\n        Title: {{ti"...
 --> src/main.rs:4:19
```

**Askama version**
a3f913af7029e8da224c425d7932dbe35c4f084c

GuillaumeGomez:
Fixes #513.

### DIFF
--- a/askama_derive/src/generator/expr.rs
+++ b/askama_derive/src/generator/expr.rs
@@ -2,7 +2,8 @@ use std::borrow::Cow;
 
 use parser::node::CondTest;
 use parser::{
-    AssociatedItem, CharLit, CharPrefix, Expr, Span, StrLit, Target, TyGenerics, WithSpan,
+    AssociatedItem, CharLit, CharPrefix, Expr, PathComponent, Span, StrLit, Target, TyGenerics,
+    WithSpan,
 };
 use quote::quote;
 
@@ -74,7 +75,7 @@ impl<'a> Generator<'a, '_> {
             }
             Expr::Index(ref obj, ref key) => self.visit_index(ctx, buf, obj, key)?,
             Expr::Filter(ref v) => {
-                self.visit_filter(ctx, buf, &v.name, &v.arguments, &v.generics, expr.span())?
+                self.visit_filter(ctx, buf, &v.name, &v.arguments, expr.span())?
             }
             Expr::Unary(op, ref inner) => self.visit_unary(ctx, buf, op, inner)?,
             Expr::BinOp(ref v) => self.visit_binop(ctx, buf, v.op, &v.lhs, &v.rhs)?,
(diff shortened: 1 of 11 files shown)


askama_derive/src/generator/expr.rs
```diff
@@ -458,7 +459,7 @@ impl<'a> Generator<'a, '_> {
         for _ in 0..*refs {
             buf.write('&');
         }
-        self.visit_path(buf, path);
+        self.visit_macro_path(buf, path);
```
GuillaumeGomez, line 462: I'm wondering if I should allow `TyGenerics` to have path with generics in them... No limit to generics recursion. =D

askama_derive/src/generator/expr.rs
```diff
@@ -539,12 +539,13 @@ impl<'a> Generator<'a, '_> {
                 }
             }
             // We special-case "askama::get_value".
-            Expr::Path(path) if path == &["askama", "get_value"] => {
+            Expr::Path(path) if matches!(path.as_slice(), [part1, part2] if part1.generics.is_empty() && part1.name == "askama" && part2.name == "get_value") =>
```
GuillaumeGomez, line 542: `if let` is not supported as pattern guard. :'(

askama_derive/src/generator/expr.rs
```diff
@@ -539,12 +539,13 @@ impl<'a> Generator<'a, '_> {
...
+            {
                 self.visit_value(
                     ctx,
                     buf,
                     args,
-                    generics,
+                    &path[1].generics,
```
GuillaumeGomez, line 548: Should I use `path.last().unwrap().generics` instead?

askama_derive/src/generator/filter.rs
```diff
@@ -14,16 +17,16 @@ impl<'a> Generator<'a, '_> {
...
         args: &[WithSpan<'a, Expr<'a>>],
-        generics: &[WithSpan<'a, TyGenerics<'a>>],
         node: Span<'_>,
     ) -> Result<DisplayWrap, CompileError> {
-        let name = match name {
-            PathOrIdentifier::Path(path) => {
-                return self.visit_custom_filter_with_path(ctx, buf, path, args, generics, node);
-            }
-            PathOrIdentifier::Identifier(name) => *name,
+        let (name, generics): (&str, &[WithSpan<'_, TyGenerics<'_>>]) = match name {
```
GuillaumeGomez, line 24: We could cheat here and skip all `ensure_no_generics` calls (which would be nice) but it would require a small refactor of this code, increasing the diff. I suppose it's better as a follow-up?

askama_derive/src/generator/expr.rs
GuillaumeGomez: Added the code comment.

Kijewski (approved):
Nice, thanks!

PR #526, opened 2025-07-13
parser: reject bare CR in doc comments

Kijewski:
Resolves <https://issues.oss-fuzz.com/issues/431448399>.

The diff is best viewed with white spaces changes hidden.


GuillaumeGomez (approved):

GuillaumeGomez:
Everyone hates them. :p

(unless you're doing some termcaps, then it's useful)

Thanks for the fix!

PR #528, opened 2025-07-14
book: fix rust-docs on readthedocs

Kijewski:
We want to use the nightly features `doc_cfg` and `doc_auto_cfg` when we build our documentation. Still, we want to use a stable compiler, e.g. to have stable links in our docs, so we opt in to the use of unstable features with the use of `RUSTC_BOOTSTRAP=1`.

The problem is that `proc-macro2` only checks whether it can use any unstable features, not which unstable features, and it always assumes that you run the newest nightly rust version if the use of unstable features is possible.

This PR disables the detection if unstable features are available to `proc-macro2` by pretending that we are in an early stage of a bootstrapped rustc build.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,9 +11,6 @@ build:
     - asdf install rust 1.88.0
     - asdf global rust 1.88.0
 
-    # proc-macro2 v1.0.95 expects at least nightly-2025-04-16.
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Weird fix but if it works... :laughing: 

Thanks!

PR #530, opened 2025-07-18
Fix fuzzing macro argument parsing bug

GuillaumeGomez:
Fixes https://issues.oss-fuzz.com/issues/432515264.

I also opened https://github.com/rust-lang/rust/issues/144161.


Kijewski:
The problem actually seems to be that we don't support string suffixes, so the suffix `r` is seen as a prefix `r` instead.

I wonder if we should simply make string suffixes an error for this release. I don't think that would break anyone's existing code, because I don't think suffixed strings are actually used for anything in rust, yet.

GuillaumeGomez:
```rust
fn main() {
    let s = r#" "#r;
}
```

gives:

```
error: suffixes on string literals are invalid
 --> src/main.rs:2:9
  |
2 | let s = r#" "#r;
  |         ^^^^^^^ invalid suffix `r`
```

Kijewski:
Try <https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a950cf78222d70bbe2bfbecb878afc14>
```
stringify!(r#" "#r)
```
Suffixed strings are valid in macros.

GuillaumeGomez:
Wth. Ok so that's a completely different issue than what I originally though. TIL.

Kijewski:
Yeah, string suffixes are strange. :) I did not know about them either before the multitude of fuzzing errors.

GuillaumeGomez:
I don't think it's actually valid. I think `stringify` is the one allowing unholy thing in this case:

```
macro_rules! e {
    ($x:literal) => {
        println!("{:?}", $x);
    }
}

fn main() {
    e!(r#" "#r);
}
```

This code fails with:

```
error: suffixes on string literals are invalid
 --> src/main.rs:9:8
  |
9 |     e!(r#" "#r);
  |        ^^^^^^^ invalid suffix `r`
```

Kijewski:
It's not a literal, but it's an expr, expr_2021, pat, pat_param, stmt, and tt: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9a5aaf580a1dca72ea716447d2105afa>

```rust
macro_rules! ignore_me {
    ($token:tt) => {()};
}

const _: () = ignore_me!(r#" "#r);
```

GuillaumeGomez:
Applied your suggestion. Much simpler, like it haha.

Kijewski:
Nice! :) Can you add the artifact / [reproducer testcase](https://oss-fuzz.com/download?testcase_id=5780292277436416) to [`fuzzing/fuzz/artifacts/derive/`](https://github.com/askama-rs/askama/tree/1836ffde5a1a405eb30f514e7dcfb7331d4e4f0a/fuzzing/fuzz/artifacts/derive)? It helps the fuzzer finding new interesting cases.

GuillaumeGomez:
Sure, added it as well.

Kijewski (approved):
Thanks!

PR #533, opened 2025-07-21
Implement #531: Allow named arguments in call expr macro invocations

### LINKED ISSUE #531: Named argument syntax in macro call expression
As followup for the macro call expression support that landed in #518:
At the moment, the macro call expression syntax only supports positioned arguments. So we currently miss support for named argument calls:

```jinja
{% macro testmacro( required_arg, optional_arg = 0 ) %}
    {{required_arg}} | {{ optional_arg }}
{% endmacro %}

{# v supported #}
{{ testmacro(42) }}

{# v supported #}
{{ testmacro(42, 1337) }}

{# v currently not supported #}
{{ testmacro(required_arg = 42, optional_arg = 1337) }}
```

At the moment, the call expr syntax "accidentally" works, because it looks like a valid rust expression. The named arguments syntax changes this symbiosis. So adding support for named arguments will probably require adding sanity checks on normal rust expressions (`doesn't contain named args`).

Would it be possible to add support for this?

seijikun:
@Kijewski 's pointer in https://github.com/askama-rs/askama/issues/531#issuecomment-3097969535 was priceless. Thank you!
I thought I would have to add support for named arguments to the lexer of `Expr` first...

Turns out the change is actually quite easy.
**Hint:** It's probably easier to read commit by commit again

Fixes #531.

### DIFF
--- a/askama_parser/src/expr.rs
+++ b/askama_parser/src/expr.rs
@@ -251,7 +251,6 @@ impl<'a> Expr<'a> {
     pub(super) fn arguments(
         i: &mut &'a str,
         level: Level<'_>,
-        is_template_macro: bool,
     ) -> ParseResult<'a, Vec<WithSpan<'a, Self>>> {
         let _level_guard = level.nest(i)?;
         let mut named_arguments = HashSet::new();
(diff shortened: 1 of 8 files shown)


seijikun:
I think the CI failure is not caused by the changes(?)

GuillaumeGomez (approved):

GuillaumeGomez:
Changes look good to me, thanks!

And yeah, this CI check fails from time to time.

askama_parser/src/expr.rs
```diff
@@ -1144,7 +1128,7 @@ fn doc_comment_no_bare_cr<'a>(
     let (is_doc_comment, comment) = inner.with_taken().parse_next(i)?;
     if is_doc_comment && comment.split('\r').skip(1).any(|s| !s.starts_with('\n')) {
         cut_error!(
-            "bare CR not allowed in doc comment, 
+            "bare CR not allowed in doc comment,
```
Kijewski, line 1131: Oops, looks like I forgot to add a backslash in this message, and the editor (rightfully so) trimmed trailing spaces. I'll fix this is another PR and add the missing UI test that should have caught this change.

Kijewski:
Thank you! And I really like your TDD-style commits!

GuillaumeGomez (approved):

PR #536, opened 2025-07-22
Add conditional expressions

Kijewski:
Jinja, being based on Python, allows using conditional expressions:
```jinja
{{ "then" if condition else "otherwise" }}
```
will print `"then"` if `condition` is truthy, otherwise `"otherwise"`.

This PR adds the same syntax, with a few restrictions, to Askama:

* The condition must evaluate to a `bool` (or a reference to it), same as in `{% if .. %}`.
* The else-case can be absent in Jinja, then behaves like `else None`. For Jinja that makes sense. It renders `{{ None }}` like an empty string. Askama does not do that: `Option<T>` can not be rendered for any `T`; we don't unwrap automatically. So, without automatic unwrapping, I don't see a case when you would want to omit the else-case.


Kijewski:
@GuillaumeGomez, so what do you think of this feature? Documentation and UI tests are missing, but the functionality is all there. If you give me a ":+1:" on the feature, I'd add the missing parts.

Kijewski:
Actually, I use is pattern all the time in jinja templates. Simple things like `{{ "logoff" if user else "login" }}`. Of course, you can always write `{% if user %}logoff{% else %}login{% endif %}`, but it's longer and at least to me it's harder on the eyes.

(I like python's "infix `if`" better that rust's "prefix `if`", even without the fact that it's easier to parse.)

askama_derive/src/tests.rs
```diff
@@ -1410,3 +1414,117 @@ fn test_bare_cr_doc_comment() -> Result<(), syn::Error> {
...
+            ("b", "i8"),
+            ("c", "i8"),
+            ("d", "i8"),
+            ("e", "i8"),
+        ],
+        3,
+    );
+
+    compare(
+        r"{{ a if b else c if d else e }}",
```
GuillaumeGomez, line 1503: This is absolutely horrifying my brain to parse. :rofl:

GuillaumeGomez:
Thinking some more, does it mean we can have stuff like:

```jinja
{{ a(x if y else z) }}
{% if (a if x else b) %}
```

?

Kijewski:
Yes! Both examples would work.

GuillaumeGomez:
I'm very afraid of opening this door. :rofl:

GuillaumeGomez:
Anyway, I don't any valid argument against this feature, so if you have a use for it and want to add it, then let's go. Please ping me once PR is ready for review with tests and docs. :)

askama_derive/src/tests.rs
Kijewski: What about this formatting? :)
```jinja
{{
   a if b else
   c if d else
   e
}}
```

Kijewski: Yes, that does work.

PR #540, opened 2025-07-24
derive: un-inline `Expr::Call` handling for `{{ expressions }}` and remove parentheses

Kijewski:
* The call handling is by far the biggest part of the function.
* Parentheses hurt automatic optimizations further down like compile-time escaping

The diff is best viewed with spaces hidden.


GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #546, opened 2025-07-27
derive: track included files with relative path

Kijewski:
To make the compiler understand which files contributed to the generated code, in order to make it aware when a recompilation is needed, we track the file paths using `include_bytes!()` calls in the generated code. This can be a problem for reproducible builds, because the working directory contributes to the file hash.

This PR makes Askama generate relative paths if possible. This is always possible on Unix systems, but on Windows hosts absolute paths have to be used if the code is kept on different drives.

[`Span::local_file()`](https://doc.rust-lang.org/1.88.0/proc_macro/struct.Span.html#method.local_file) is only stable since rust 1.88, but that's our MSRV for the next release, anyway.

Open questions:
- <del>[ ] Should this behavior be feature gated?</del>
- [x] [`path_clean::clean()`](https://docs.rs/path-clean/1.0.1/src/path_clean/lib.rs.html#58-86) and [`pathdiff:diff_paths()`](https://docs.rs/pathdiff/0.2.3/src/pathdiff/lib.rs.html#43-86) are tiny, and both MIT OR Apache-2.0 licensed. Should we simply copy the code (mentioning the authors, of course)?

Resolves <https://github.com/askama-rs/askama/issues/461>.  
Cc @frankdavid.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -5,8 +5,9 @@ mod node;
 
 use std::borrow::Cow;
 use std::collections::hash_map::HashMap;
+use std::env::current_dir;
 use std::ops::Deref;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str;
 use std::sync::Arc;
 
@@ -17,6 +18,7 @@ use parser::{
 use rustc_hash::FxBuildHasher;
 
 use crate::ascii_str::{AsciiChar, AsciiStr};
+use crate::generator::helpers::{clean_path, diff_paths};
 use crate::heritage::{Context, Heritage};
 use crate::html::write_escaped_str;
 use crate::input::{Source, TemplateInput};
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
> *  Should this behavior be feature gated?

Is there any reason why it should?

> * [`path_clean::clean()`](https://docs.rs/path-clean/1.0.1/src/path_clean/lib.rs.html#58-86) and [`pathdiff:diff_paths()`](https://docs.rs/pathdiff/0.2.3/src/pathdiff/lib.rs.html#43-86) are tiny, and both MIT OR Apache-2.0 licensed. Should we simply copy the code (mentioning the authors, of course)?

I'd say yes. One less dependency and as long as we mention where the code comes from, it gives us more liberty to adapt it more to ours needs if needed.

Kijewski:
> Is there any reason why it should?

Not really, no. :) If the new import path handling was feature gated, then the new dependencies could be optional. But if we copy the two functions, then there are not new dependencies, so no feature-gate is needed.

PR #553, opened 2025-07-31
parser: `r#self` is not a valid raw identifier in macro calls

Kijewski:
It's not a valid raw identifier anywhere, actually, but the other cases (in a path, as a field name, as a macro name) were already handled.

Resolves <https://issues.oss-fuzz.com/issues/435218013>.


GuillaumeGomez:
We keep discovering much weird stuff. ^^'

Thanks for the fix!

PR #558, opened 2025-08-03
Generate a `TokenStream` rather than a `String`

GuillaumeGomez:
Part of https://github.com/askama-rs/askama/issues/420.

This PR allows to greatly improve compile-time errors by using `Span`. Even better: even a compiler error occurs, it shows which part of the jinja code generated it.

BIG limitations with this PR:
 * It doesn't support `path`, only `src`.
 * It doesn't work with sub-templates (ie `includes` or `extends`).

As a first step though, I think it's good enough.

### DIFF
--- a/askama_derive/src/generator/expr.rs
+++ b/askama_derive/src/generator/expr.rs
@@ -6,7 +6,7 @@ use parser::{
     AssociatedItem, CharLit, CharPrefix, Expr, PathComponent, Span, StrLit, StrPrefix, Target,
     TyGenerics, WithSpan,
 };
-use proc_macro2::TokenStream;
+use proc_macro2::{Delimiter, Spacing, TokenStream, TokenTree};
 use quote::quote_spanned;
 use syn::Token;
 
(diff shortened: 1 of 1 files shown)


Kijewski:
I removed `impl BufferArg for {String, &str, Arguments}`, so spans are better kept. (I wouldn't have done it, if I had known beforehand how much work this would going to be. :) ) Implementing the PR must have been horrible!

GuillaumeGomez:
Yep, it was quite the nightmare. ^^'

Kijewski:
Sorry, my commits are terribly unorganized. :-/

I only found a possible bug with `{% for _ in self..X %}` (which will probably never be encountered in the wild), otherwise everything looks very good. :+1:

GuillaumeGomez:
I can't approve because it's my PR, github doesn't allow it. But changes look good to me, thanks!

Kijewski:
Thank you for your work, this change is great!

Hopefully someday soon `#![feature(proc_macro_expand)]`+`#![feature(proc_macro_span)]`+`#![feature(proc_macro_value)]` (or another proc-macro accessible `include_bytes!()` feature) will be stable. Then we can add spans for included files, too.

PR #560, opened 2025-08-07
parser: use `LocatingSlice<&str>` instead of `&str`

Kijewski:
This will enable better span getting in subsequent PRs.

### DIFF
--- a/askama_parser/src/lib.rs
+++ b/askama_parser/src/lib.rs
@@ -23,7 +23,7 @@ use winnow::ascii::take_escaped;
 use winnow::combinator::{
     alt, cond, cut_err, delimited, empty, fail, not, opt, peek, preceded, repeat, terminated,
 };
-use winnow::error::{ErrMode, FromExternalError};
+use winnow::error::ErrMode;
 use winnow::stream::AsChar;
 use winnow::token::{any, none_of, one_of, take_while};
 use winnow::{LocatingSlice, ModalParser, ModalResult, Parser, Stateful};
@@ -368,15 +368,6 @@ impl<'a> winnow::error::ParserError<InputStream<'a>> for ErrorContext<'a> {
     #[inline(always)]
     fn into_inner(self) -> Result<Self::Inner, Self> {
         Ok(self)
-    }
-}
-
-impl<'a, E: std::fmt::Display> FromExternalError<&'a str, E> for ErrorContext<'a> {
-    fn from_external_error(input: &&'a str, e: E) -> Self {
-        Self {
-            span: (*input).into(),
-            message: Some(Cow::Owned(e.to_string())),
-        }
     }
 }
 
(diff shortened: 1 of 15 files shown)


askama_parser/src/lib.rs
```diff
@@ -1466,13 +1479,13 @@ fn cut_context_err<'a, T>(gen_err: impl FnOnce() -> ErrorContext<'a>) -> ParseRe
 
 type HashSet<T> = std::collections::hash_set::HashSet<T, FxBuildHasher>;
 
-#[cfg(not(windows))]
```
Kijewski, line 1469: Only the `Path` formatting tests are platform specific.

askama_parser/src/node.rs
```diff
@@ -37,38 +37,28 @@ pub enum Node<'a> {
...
-                    && err.message.is_none()
-                {
-                    *i = start;
-                    if let Some(mut span) = err.span.as_suffix_of(i) {
-                        opt(|i: &mut _| unexpected_tag(i, s)).parse_next(&mut span)?;
-                    }
-                }
-                return Err(err);
-            }
-        };
```
Kijewski, line 57: I moved the logic of this part into its own function, because `cut_node` works the same. No need to have the same code twice.

Kijewski:
With this PR, we can use [`Parser::[with_]span()`](https://docs.rs/winnow/0.7.12/winnow/trait.Parser.html#method.span) to get the byte range of a parsed text. This can replace our manual taking of spans, and [`Literal::subspan()`](https://doc.rust-lang.org/1.88.0/proc_macro/struct.Literal.html#method.subspan) expects a byte range. I use a `Stateful<_, ()>` stream, so in a subsequent PR the `()` can be replace by an reference to the source file, which should become part of `askama_parser::Span`.

For every derived `TokenStream` we need to [include](https://github.com/Kijewski/proc-macro-include-std-test/blob/d261ded82b8d6c29b40ebadd417364aab8e7f108/src/lib.rs#L11-L18) the sources anew, because [`Literal`](https://docs.rs/proc-macro2/1.0.95/proc_macro2/struct.Literal.html)s are thread local, i.e. `!Send + !Sync`. But we can still cache the parsed syntax tree.

@GuillaumeGomez, do you want to implement these subsequent PRs to make the underlining work? The feature could even allow debug stepping through a template file!

testing/tests/ui/char_literal.stderr
```diff
@@ -54,23 +54,23 @@ error: unicode escape must be at most 10FFFF
 28 | #[template(path = "char-literals/char-literal-7.txt")]
    |   ^^^^^^^^
 
-error: invalid character
+error: cannot have multiple characters in a character literal, use `"..."` to write a string
```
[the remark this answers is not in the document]
GuillaumeGomez: Good idea!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks a lot for this huge improvement!

> @GuillaumeGomez, do you want to implement these subsequent PRs to make the underlining work? The feature could even allow debug stepping through a template file!

I can if you don't have time for it, but seems like you already started quite strongly on it. :laughing:

PR #568, opened 2025-08-12
Fix #563: filesizeformat with proper accuracy

### LINKED ISSUE #563: filesizeformat filter's accuracy very limited
**Describe the bug**
The `filesizeformat` filter internally uses `f32`. Unfortunately, that's not accurate enough for many common values.

**To Reproduce**
Add the following 2 assertions to the unit-test `test_filesizeformat()`:

```rust
assert_eq!(filesizeformat(300.0 * 1000.0 * 1000.0 * 1000.0).unwrap().to_string(), "300 GB");
assert_eq!(filesizeformat(600.0 * 1000.0 * 1000.0 * 1000.0).unwrap().to_string(), "600 GB");
```

Result:

```
thread 'filters::humansize::test_filesizeformat' panicked at askama/src/filters/humansize.rs:124:5:
assertion `left == right` failed
  left: "299.99 GB"
 right: "300 GB"
```

Now, I could just replace `f32` with `f64`. That makes it go away, but I don't know if that change makes every case go away. And since it looks rather ... optimized, I don't really dare to touch it.

seijikun:
I refactored the current `humansize` / `filesize` implementation from `f32` to `u128` and added more unit-tests.
Additionally, the new refactoring now allows changing the requested precision (not yet plumbed on the code generation side).

`u128` is required to handle the SI Prefixes:
```rust
    ((AsciiChar::new(b'Z'), 1e-19), 1e24),
    ((AsciiChar::new(b'Y'), 1e-22), 1e27),
    ((AsciiChar::new(b'R'), 1e-25), 1e30),
    ((AsciiChar::new(b'Q'), 1e-28), 1e33),
```

Getting something as performant as the original version is quite challenging. Here are my benchmark results for the current solution:

Benchmarks (time for `250000000` format & display invocations):
- Original (`f32`): `1.9542559s`
- This MR, but with `u64`: `2.638588188s`
- This MR (`u128`): `3.29419516s`

### DIFF
--- a/askama/src/filters/humansize.rs
+++ b/askama/src/filters/humansize.rs
@@ -1,8 +1,6 @@
 use core::convert::Infallible;
 use core::fmt;
-use core::mem::MaybeUninit;
 
-use crate::ascii_str::{AsciiChar, AsciiStr};
 use crate::{FastWritable, NO_VALUES, Values};
 
 /// Returns adequate string representation (in KB, ..) of number of bytes
@@ -12,7 +10,7 @@ use crate::{FastWritable, NO_VALUES, Values};
 /// # use askama::Template;
 /// #[derive(Template)]
 /// #[template(
-///     source = "Filesize: {{ size_in_bytes|filesizeformat }}.",
+///     source = "Filesize: {{ size_in_bytes | filesizeformat }}.",
 ///     ext = "html"
 /// )]
 /// struct Example {
@@ -23,12 +21,18 @@ use crate::{FastWritable, NO_VALUES, Values};
 /// assert_eq!(tmpl.to_string(),  "Filesize: 1.23 MB.");
 /// ```
 #[inline]
-pub fn filesizeformat(b: f32) -> Result<FilesizeFormatFilter, Infallible> {
-    Ok(FilesizeFormatFilter(b))
+pub fn filesizeformat(bytes: u128, precision: u8) -> Result<FilesizeFormatFilter, Infallible> {
+    Ok(FilesizeFormatFilter { bytes, precision })
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct FilesizeFormatFilter(f32);
+pub struct FilesizeFormatFilter {
+    /// The number of bytes to format nicely
+    bytes: u128,
+    /// The precision with which to display the generated string.
+    /// This determines the number of digits after the decimal separator.
+    precision: u8,
+}
 
 impl fmt::Display for FilesizeFormatFilter {
     #[inline]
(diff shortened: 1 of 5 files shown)


askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,100 @@ impl FastWritable for FilesizeFormatFilter {
...
+            let remainder = self.0 % unit.1;
 
-/// Formats `buffer` to contain the decimal point, decimal places and unit
-fn format_frac(buffer: &mut MaybeUninit<[AsciiChar; 8]>, prefix: AsciiChar, scaled: u32) -> &str {
-    // LLVM generates better byte code for register sized buffers
-    let buffer = buffer.write(AsciiStr::new_sized("..0 kB"));
-    buffer[4] = prefix;
+            // Compute fractional part with desired precision
+            let precision = self.1;
+            let mut scale = 10u128.pow(precision as u32);
```
Kijewski, line 59: Why is this needed? Can't the scale be derived by the value found in SI_PREFIXES?

Kijewski:
Working on u128 would probably be faster than working on f32. I guess formatting directly into the output stream is what makes the new code slower.

askama/src/filters/humansize.rs
seijikun: The scale handles the `precision`. If you e.g. want 3 decimals, the scale will be `1000`.

Example for a precision of 2 (scale = `100`):
Take for example the number of bytes:  `1024`. The best fitting prefix is `1000` (= kB).
So the `integer` portion will be `1024 / 1000 = 1`, and the remainder will be `24`.
The `fractional` portion is generated by first scaling the remainder up and then dividing by unit: `24 * 100 / 1000 = 2`.

The idea is, that this allows to use integer formatting to print the part after the decimal separator. I can then simply format the two integers: `format!("{integer}.{fractional}")`.

Big advantage of this is, that I can freely change the precision (compared to the original version that was fixed to 2).
Or did I misunderstand your question?

GuillaumeGomez:
> Working on u128 would probably be faster than working on f32. I guess formatting directly into the output stream is what makes the new code slower.

Nope, if floats and integers are of the same size, the floats are always slower, however floats are always faster than bigger integers (so u64 > f32 > u32).

Also, what kind of numbers would we need above 18 exabytes (`u64::MAX`)? O.o (I think `u64` is more than enough)

seijikun:
I knew binary search was probably slower than linear search in this case, but I didn't think it was noticeable...
But holy cow...

Now I'm at `1.832858975s` with `u128` :)

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,100 @@ impl FastWritable for FilesizeFormatFilter {
...
+                    digit.write_into(dest, values)?;
+                    fractional %= scale;
+                    if fractional == 0 {
+                        break;
+                    }
+                }
+            }
+            ' '.write_into(dest, values)?;
+            unit.0.write_into(dest, values)?;
+            'B'.write_into(dest, values)?;
```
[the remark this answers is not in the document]
Kijewski: I would not have assumed that the binary search was what made the code slower. It's okay that you removed the function. Configurable decimal places are a good addition!

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,101 @@ impl FastWritable for FilesizeFormatFilter {
...
-    ((AsciiChar::new(b'Q'), 1e-28), 1e33),
+/// (`si_prefix`-char, `amount of bytes corresponding to 1 unit`)
+const SI_PREFIXES: &[(char, u128)] = &[
+    ('k', 1_000),
+    ('M', 1_000_000),
+    ('G', 1_000_000_000),
+    ('T', 1_000_000_000_000),
+    ('P', 1_000_000_000_000_000),
+    ('E', 1_000_000_000_000_000_000),
+    ('Z', 1_000_000_000_000_000_000_000),
```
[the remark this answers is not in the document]
seijikun: To quote Homer Simpson:

> I, ehh... I need to do this for work!

image

seijikun:
Just some final formatting improvements.
Could one of you plumb the precision with an optional parameter, please?
I wasn't able to get that to work.

GuillaumeGomez:
Might be better if you did so you can learn: what did you try?

There are a few things needed to make it work: the `filesizeformat` function needs to take all args and the optional ones need to use `Option`. Then in `askama_derive`, depending on the arguments, you call the function with `Some` or `None` for the optional args.

seijikun:
> Might be better if you did so you can learn: what did you try?
> 
> There are a few things needed to make it work: the `filesizeformat` function needs to take all args and the optional ones need to use `Option`. Then in `askama_derive`, depending on the arguments, you call the function with `Some` or `None` for the optional args.

I had trouble understanding how the new argument parsing stuff works (how to fill the `default_value` field). But I think I got it to work now

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,186 @@ impl FastWritable for FilesizeFormatFilter {
...
+            let remainder = self.0 % unit.1;
+
+            // Compute fractional part with desired precision
+            let precision = self.1.min(15); // max. allowed precision is 15
+            let mut scale = 10u128.pow(precision as u32);
+            let mut fractional = remainder.checked_mul(scale).unwrap_or_default() / unit.1;
+
+            // integer portion (before decimal point)
+            integer.write_into(dest, values)?;
+            if precision > 0 && fractional > 0 {
```
Kijewski, line 65: `fractional` is only used if the condition `precision > 0` is true. Maybe move the assignment into the if-body?

seijikun:
One more improvement. Instead of just stupidly limiting the precision to the completely arbitrary number of `15` due to numerical problems, I sat down and did some thinking:

Allowing a larger `precision` than the chosen si-prefix doesn't make any sense, because except for the super-edgecase of more than 1000 quetta bytes, the additional digits will all just be `0`s, which are stripped anyway.
This way, I can guarantee that this is always true: `scale <= unit.1`.
Furthermore, both numbers are `10`-based, so this is also always true: `(unit.1 % scale) == 0`.
This allowed me to change the `fractional` calculation from:
```rust
let mut fractional = remainder.saturating_mul(scale) / unit.1;
```
to:
```rust
let mut fractional = remainder / (unit.1 / scale);
```
which effectively avoids overflow due to the multiplication, while allowing much higher precision.

askama/src/filters/humansize.rs
```diff
@@ -23,12 +21,12 @@ use crate::{FastWritable, NO_VALUES, Values};
...
 #[inline]
-pub fn filesizeformat(b: f32) -> Result<FilesizeFormatFilter, Infallible> {
-    Ok(FilesizeFormatFilter(b))
+pub fn filesizeformat(b: u128, p: u8) -> Result<FilesizeFormatFilter, Infallible> {
+    Ok(FilesizeFormatFilter(b, p))
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct FilesizeFormatFilter(f32);
+pub struct FilesizeFormatFilter(u128, u8);
```
GuillaumeGomez, line 29: Could you use fields instead? Having context on what the types are actually for is always a good idea. :)

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,190 @@ impl FastWritable for FilesizeFormatFilter {
         dest: &mut W,
         values: &dyn Values,
     ) -> crate::Result<()> {
-        if self.0 < 1e3 {
+        if self.0 < 1000 {
```
GuillaumeGomez, line 44: Since it's only `if else`, could you return in the `if` please? Like that the whole `else` block can be unindented by one level.

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,190 @@ impl FastWritable for FilesizeFormatFilter {
...
+    ('M', 1_000_000, 6),
+    ('G', 1_000_000_000, 9),
+    ('T', 1_000_000_000_000, 12),
+    ('P', 1_000_000_000_000_000, 15),
+    ('E', 1_000_000_000_000_000_000, 18),
+    ('Z', 1_000_000_000_000_000_000_000, 21),
+    ('Y', 1_000_000_000_000_000_000_000_000, 24),
+    ('R', 1_000_000_000_000_000_000_000_000_000, 27),
+    ('Q', 1_000_000_000_000_000_000_000_000_000_000, 30),
+];
```
Kijewski, line 102: Good idea to store the maximum number of trailing digits as part of the table! Maybe use a const fn to make it a bit more readable?
```rust
const SI_PREFIXES: &[(char, u128, u8)] = &[
    si_prefix('k', 3),
    si_prefix('M', 6),
    si_prefix('G', 9),
    si_prefix('T', 12),
    si_prefix('P', 15),
    si_prefix('E', 18),
    si_prefix('Z', 21),
    si_prefix('Y', 24),
    si_prefix('R', 27),
    si_prefix('Q', 30),
];

const fn si_prefix(prefix: char, exponent: u8) -> (char, u128, u8) {
    (prefix, 10u128.pow(exponent as u32), exponent)
}
```

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,190 @@ impl FastWritable for FilesizeFormatFilter {
...
+                        fractional %= scale;
+                        if fractional == 0 {
+                            break;
+                        }
+                    }
+                }
+            }
+            ' '.write_into(dest, values)?;
+            unit.0.write_into(dest, values)?;
+            'B'.write_into(dest, values)?;
```
GuillaumeGomez, line 83:
```suggestion
            write!(dest, " {}B", unit..0)?;
```

GuillaumeGomez: Wait really? I'm surprised, generally the I/O calls are slow so we try to do as little of them as possible.

seijikun: The `std` calls are SUPER slow. That's why I use `FastWritable::write_into()` everywhere.
I can do 100s of calls of `FastWritable::write_into()` and will still be faster than one `write!()`.

askama/src/filters/humansize.rs
```diff
@@ -43,82 +41,190 @@ impl FastWritable for FilesizeFormatFilter {
...
+
+            // Compute fractional part with desired precision
+            // limit precision to the chosen unit's exponent - a greater precision will just
+            // generated tailing 0s that will be dropped.
+            let precision = self.1.min(unit.2);
+
+            // integer portion (before decimal point)
+            integer.write_into(dest, values)?;
+            if precision > 0 {
+                let mut scale = 10u128.pow(precision as u32);
```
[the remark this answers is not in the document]
Kijewski: Hehe, I guess at this point it is good enough as it is. I am quite sure that the allocation would destroy any benefits in a real world with concurrency.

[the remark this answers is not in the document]
seijikun: I may be wrong, but I did some testing and **I think** this doesn't work for 2 reasons:

- For `filesizeformat(1024, 2)`, this will produce `1.24` instead of `1.02`.
- This will not drop trailing `0`s. So this could produce something like: `1.20` instead of `1.2`.

askama/src/filters/humansize.rs
GuillaumeGomez: This comment has not been resolved. :wink:

seijikun: Ohh sorry, I didn't look properly. I somehow thought you meant the SiPrefixes. Oh well...

seijikun:
Thank you two!

PR #570, opened 2025-08-13
Fix wrong macro argument parsing

GuillaumeGomez:
Fixes https://oss-fuzz.com/testcase-detail/5107711304073216.

Now we enforce to have a whitespace before the `#` characters when parsing macro arguments. There are likely other cases. :3


Kijewski:
This is actually a truncated [raw lifetime](https://doc.rust-lang.org/stable/edition-guide/rust-2021/raw-lifetimes.html). I would rather reject `("'r", take_while(1.., '#'))` with a message, that raw lifetimes aren't implemented, yet.
<https://github.com/askama-rs/askama/blob/98ad5d5b3fbc1e59604eb9f7eb29ea9d382d8f79/askama_parser/src/expr.rs#L939-L940>

Either that, or implement raw lifetimes.

GuillaumeGomez:
And done.

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,27 @@ impl<'a> Suffix<'a> {
             }
         }
 
+        fn lifetime<'a>(i: &mut InputStream<'a>) -> ParseResult<'a, ()> {
+            // Before the 2021 edition, awe can have whitespace characters between "r#" and the
```
Kijewski, line 931: Stray "awe" in the sentence. :)

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,27 @@ impl<'a> Suffix<'a> {
...
+            // identifier so we allow it here.
+            let start = *i;
+            '\''.parse_next(i)?;
+            if alt((
+                (keyword("r#"), opt(skip_ws0), identifier).map(|(_, _, _)| ()),
+                (identifier, not(peek('#'))).map(|(_, _)| ()),
+            ))
+            .parse_next(i)
+            .is_err()
```
Kijewski, line 940: Please don't ignore errors from the parser. The error location can be off. Use `opt(..).parse_next(i)?.is_none()` instead.

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,27 @@ impl<'a> Suffix<'a> {
             }
         }
 
+        fn lifetime<'a>(i: &mut InputStream<'a>) -> ParseResult<'a, ()> {
+            // Before the 2021 edition, awe can have whitespace characters between "r#" and the
+            // identifier so we allow it here.
```
Kijewski, line 932: I would still reject `'r#<space>`. We are `--edition 2024`, so a user should not expect that their `--edition 2018` code works without changes. (At least for such esoteric cases.) In `--edition 2021` it was already wrong: <https://godbolt.org/z/8WbrPxv83>

GuillaumeGomez: We are, but the codebase using askama might not be.

Kijewski: Can you please add a test what happens if the template code contains `{{ x!( #r' y )}}`? Only to make sure that it does not panic.

GuillaumeGomez: Sure, done!

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,27 @@ impl<'a> Suffix<'a> {
             }
         }
 
+        fn lifetime<'a>(i: &mut InputStream<'a>) -> ParseResult<'a, ()> {
+            // Before the 2021 edition, we can have whitespace characters between "r#" and the
+            // identifier so we allow it here.
+            let start = *i;
+            '\''.parse_next(i)?;
+            if opt(alt((
+                (keyword("r#"), opt(skip_ws0), identifier).map(|(_, _, _)| ()),
```
Kijewski, line 936: I don't think this code can ever match. A `#` cannot be part of a keyword. I guess this should be `"r#"`.

Can you please add a test that contains raw lifetimes that get parsed correctly? Something as simple as `macro_rules! test { ('r#ignore_me) => { "ok" } }` → `{{ test!('r#ignore_me) }}` → `assert_eq!(.., "ok")`?

Kijewski, line 936: The identifier of a raw lifetime must not be `"crate" | "self" | "super" | "Self" | "_"`. For "normal" lifetimes the identifiers are okay, though. <https://doc.rust-lang.org/reference/tokens.html#grammar-LIFETIME_TOKEN>

Actually, raw lifetimes seem to be only useful as labels. A lifetime `'break` is okay, but `'break loop { break 'break; }` is not.

testing/tests/ui/macro-args-hashtag.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Template)]
+#[template(
+    source = r###"{{z!{'r#}}}"###,
+    ext = "html"
+)]
+struct Example;
+
+#[derive(Template)]
+#[template(
+    source = r###"{{z!{#r' y}}}"###,
```
Kijewski, line 15:
```suggestion
    source = r###"{{z!{'r# y}}}"###,
```

GuillaumeGomez: Makes more sense, was wondering if it was intended. ^^'

Kijewski: Well, there's six ways to order three different letters: `'r#`, `'#r`, `r'#`, `r#'`, `#r'`, `r#'`. That means that I beat the odds when I type them in the correct order half of the time! :D 🙈

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,32 @@ impl<'a> Suffix<'a> {
...
+            .parse_next(i)?
+            else {
+                return cut_error!("wrong lifetime format", **start);
+            };
+            if crate::is_rust_keyword(identifier) {
+                return cut_error!(
+                    "a non-raw lifetime cannot be named like an existing keyword",
+                    **start,
+                );
+            }
```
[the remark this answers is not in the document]
GuillaumeGomez: Apparently there are restrictions:

```
error: lifetimes cannot use keyword names
 --> src/main.rs:1:6
  |
1 | fn a<'self>(a: &'self str) {}
  |      ^^^^^

error: lifetimes cannot use keyword names
 --> src/main.rs:1:17
  |
1 | fn a<'self>(a: &'self str) {}
  |                 ^^^^^
```

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,36 @@ impl<'a> Suffix<'a> {
...
         }
 
+        fn lifetime<'a>(i: &mut InputStream<'a>) -> ParseResult<'a, ()> {
+            // Before the 2021 edition, we can have whitespace characters between "r#" and the
+            // identifier so we allow it here.
+            let start = *i;
+            '\''.parse_next(i)?;
+            let Some((is_raw, identifier)) = opt(alt((
+                ('r', '#', identifier).map(|(_, _, ident)| (true, ident)),
+                (identifier, not(peek('#'))).map(|(ident, _)| (false, ident)),
```
Kijewski, line 937: Much more readable this way! :+1:

askama_parser/src/expr.rs
```diff
@@ -927,6 +927,36 @@ impl<'a> Suffix<'a> {
...
+                    return cut_error!(
+                        "a non-raw lifetime cannot be named like an existing keyword",
+                        **start,
+                    );
+                }
+            } else if ["Self", "self", "crate", "super", "_"].contains(&identifier) {
+                return cut_error!(format!("`{identifier}` cannot be a raw lifetime"), **start,);
+            }
+            let res: ParseResult<'a, char> = peek('\'').parse_next(i);
+            if res.is_ok() {
```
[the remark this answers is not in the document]
GuillaumeGomez: Sure. :laughing:

PR #574, opened 2025-08-17
Add tests for jinja examples in our book

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/458.

### DIFF
--- a/book/src/filters.md
+++ b/book/src/filters.md
@@ -104,7 +104,7 @@ Output:
 ### default
 [#default]: #default
 
-```text
+```jinja,ignore
 {{ variable_or_expression | default(default_value) }}
 {{ variable_or_expression | default(default_value, [[boolean =] true]) }}
 ```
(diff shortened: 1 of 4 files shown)


book/src/filters.md
```diff
@@ -104,7 +104,7 @@ Output:
 ### default
 [#default]: #default
 
-```jinja
+```text
```
Kijewski, line 107: Is it possible to make this ```` ```jinja,ignore ````? One does not see it on the stable book branch, but on latest the jinja code is highlighted: <https://askama.readthedocs.io/en/latest/template_syntax.html>. It would be sad if the highlighting was disabled for half of the examples.

Hm, I don't know if highlight.js works a comma separated list, but I guess it's part of commonmark standard?

GuillaumeGomez: Fair enough.

Kijewski:
This is great, thank you!

PR #580, opened 2025-08-19
Emit warnings if duplicated `block` calls are done

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/302.

Sadly, since we can't use the rust diagnostic API (yet hopefully), we can't test this output without a new testsuite. Not sure it it's worth it though.

Anyway, just running tests triggers the warning quite a lot. :)


GuillaumeGomez:
Also added some comments in the code.

Kijewski:
Clippy is unhappy. Can you add a UI test?

Kijewski (approved):

Kijewski:
Ah, yes, you are right! That's the same problem we had in <https://github.com/askama-rs/askama/pull/102>.

seijikun:
I'm a bit confused with this change.
I went to the [Askama Documentation for Template Inheritance](https://askama.readthedocs.io/en/stable/template_syntax.html#template-inheritance), just pasted the small example (consisting of `Base Template` and `Child Template`) into 2 files and hit compile.

And I get warnings for this:
```
⚠️ templates/base.jinja:4:14: block `title` was already called at `templates/page.jinja` so the previous one will be ignored
⚠️ templates/base.jinja:5:7: block `head` was already called at `templates/page.jinja` so the previous one will be ignored
⚠️ templates/base.jinja:9:9: block `content` was already called at `templates/page.jinja` so the previous one will be ignored
```

Is this change supposed to warn for here? I thought this is the only way using this feature actually makes sense. Am I misunderstanding something here?

GuillaumeGomez:
This implementation is buggy. I opened https://github.com/askama-rs/askama/pull/586 to fix it.

PR #584, opened 2025-08-23
Emit an error if an extends block doesn't come first in a template

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/164.

### DIFF
--- a/askama_parser/src/node.rs
+++ b/askama_parser/src/node.rs
@@ -53,12 +53,29 @@ impl<'a: 'l, 'l> Node<'a> {
     }
 
     fn many(i: &mut InputStream<'a, 'l>) -> ParseResult<'a, Vec<Box<Self>>> {
-        repeat(
-            0..,
-            alt((Lit::parse, Comment::parse, Self::expr, Self::parse)),
-        )
-        .map(|v: Vec<_>| v)
-        .parse_next(i)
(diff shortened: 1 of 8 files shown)


Kijewski:
I moved the test out of `Node::many()`, so instances of `{% block my_block %}{% extends ".." %}{% endblock %}` get rejected by the parser, not only in derive.

Kijewski (approved):
Thanks!

Please review my change before merging. :)

askama_parser/src/node.rs
```diff
@@ -53,27 +72,13 @@ impl<'a: 'l, 'l> Node<'a> {
     }
 
     fn many(i: &mut InputStream<'a, 'l>) -> ParseResult<'a, Vec<Box<Self>>> {
-        let mut nb_nodes = 0;
-
-        repeat(0.., move |i: &mut _| Self::parse_nodes(i, &mut nb_nodes))
-            .map(|v: Vec<_>| v)
```
Kijewski: I don't remember why this "custom type ascription" was needed, but I'm happy that it isn't anymore!

GuillaumeGomez:
Your changes look good, thanks! :)

As a follow-up, would be nice to add a regression test for the `extends` in blocks too.

PR #586, opened 2025-08-24
Fix block duplication warning

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/583.

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -87,8 +87,6 @@ struct Generator<'a, 'h> {
     is_in_filter_block: usize,
     /// Set of called macros we are currently in. Used to prevent (indirect) recursions.
     seen_callers: Vec<(&'a Macro<'a>, Option<FileInfo<'a>>)>,
-    /// Blocks already used.
-    called_blocks: Vec<(&'a str, Option<FileInfo<'a>>)>,
     /// The directory path of the calling file.
     caller_dir: CallerDir,
 }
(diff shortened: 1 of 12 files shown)


GuillaumeGomez:
Gonna investigate why it doesn't work in CI...

Kijewski:
> Gonna investigate why it doesn't work in CI...

I'm quite sure that we cannot have two trybuilds in divert units. That's why I added new CI tests for nightly in <https://github.com/askama-rs/askama/pull/571/commits/242b4ad9dcaec35fea405375e40830a763380a61>. I tried to make it work in one runner, but trybuild is not meant to be used this way (or I'm too stupid to make this work this way).

Kijewski:
Thank you for working on that! Can you please rebase, too?

Do you know where the line ↓ in <https://github.com/askama-rs/askama/actions/runs/17211944460/job/48825805458?pr=586#step:7:72> comes from?

```text
[Comment(Comment { ws: Ws(None, None), content: " comment " }), Extends(Extends { path: "base.html" })]
```

GuillaumeGomez:
Didn't notice that one, investigating.

GuillaumeGomez:
We forgot `print = "ast"` in two tests, removed them. Also rebased on `master`.

Kijewski:
Ah, thanks for finding that! Now rustfmt is unhappy, because the now shorter annotations would fit into one line.

GuillaumeGomez:
Didn't think about that. :laughing: 

rustfmt should be happy now. :)

Kijewski (approved):
Thanks!

PR #600, opened 2025-09-25
Fix wrong parsing of number if digit does not match base

GuillaumeGomez:
Fixes https://issues.oss-fuzz.com/issues/442227549.

The problem was that if a digit was not part of the selected base, we stopped the parsing. Meaning we would have two numbers following each other, failing in case it's inside a macro as they would be stuck as one, failing the rustc lexer.


Kijewski (approved):
Sorry for being absent the last weeks, and thank you for fixing the problem. LGTM!

PR #604, opened 2025-09-28
Remove removed nightly feature

GuillaumeGomez:
I merged `doc_auto_cfg` into `doc_cfg` in https://github.com/rust-lang/rust/pull/138907 so we can/should remove it.


Kijewski (approved):
> I merged `doc_auto_cfg` into `doc_cfg`

Thank for the PR, and the upstream change! :)

PR #610, opened 2025-10-16
Update branch name in github CI workflow

GuillaumeGomez:


GuillaumeGomez:
Since rustc repository is switching to `main`, I think we should do the same so I updated the default branch name. Following that, here's the fix for our CI. :)

PR #611, opened 2025-10-16
Limit maximum number of references in generics to prevent `syn` stack overflow

GuillaumeGomez:
Fixes https://oss-fuzz.com/testcase-detail/6696196543676416.


Kijewski (approved):
Thank you for the fix (and not forgetting to upload the artifact!) :D

PR #644, opened 2025-12-24
fix: allow clippy::elidable_lifetime_names for auto derived impls

### LINKED ISSUE #643: `clippy::elidable_lifetime_names` gets invalidly triggered on v0.15
**Describe the bug**
Structures that contain references that use the `Template` derive implementation trigger the `clippy::elidable_lifetime_names` lint.

**To Reproduce**
The following code:

```rust
#![deny(clippy::elidable_lifetime_names)]

use askama::Template;

#[derive(Debug, Template)]
#[template(source = "", ext = "txt")]
struct Test<'a> {
    _data: &'a (),
}

fn main() {}
```
produces the following error after running `cargo clippy`:
```
error: the following explicit lifetimes could be elided: 'a
 --> src/main.rs:7:13
  |
7 | struct Test<'a> {
  |             ^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#elidable_lifetime_names
help: elide the lifetimes
  |
7 - struct Test<'a> {
7 + struct Test'_> {
  |
```
However, the lifetime cannot be elided in this case.

Note that in addition to a regression in Askama itself, this might also qualify as a Clippy bug - see the suggestion produces an invalid Rust syntax!

**Askama version**
This is a regression introduced in v0.15. v0.14 doesn't trigger this.

**Rust version**
```
rustc 1.92.0 (ded5c06cf 2025-12-08)
clippy 0.1.92 (ded5c06cf2 2025-12-08)
```

ymelois:
As explained in <https://github.com/askama-rs/askama/issues/643> this seem to be a regression introduced in v0.15.0.

This fix allows clippy::elidable_lifetime_names explicitly and removes the _false positive_ when enabling clippy::pedantic.

Fixes #643

### DIFF
--- a/askama_derive/src/lib.rs
+++ b/askama_derive/src/lib.rs
@@ -310,6 +310,8 @@ pub fn derive_template(input: TokenStream, import_askama: fn() -> TokenStream) -
             dead_code,
             // We intentionally add extraneous underscores in type and variable names.
             non_camel_case_types, non_snake_case,
+            // The generated code contains elidable lifetime.
+            clippy::elidable_lifetime_names,
             // We have too little context information to generate better code.
             // The generated source does not have to be perfect, anyway.
             clippy::double_parens, clippy::identity_op, clippy::into_iter_on_ref,


GuillaumeGomez:
Would be better to instead fix the source of the lint.

sehnryr:
It might be more fitting to put the allowed `clippy::elidable_lifetime_names` lint under the comment
```
// We have too little context information to generate better code.
// The generated source does not have to be perfect, anyway.
```

PR #647, opened 2025-12-25
fix(derive): use call_site to avoid linter false positives

ymelois:
Implements the suggestion from https://github.com/askama-rs/askama/issues/643#issuecomment-3690599397 by using `Span::call_site()`.

Applied call-site spans to generated impl headers/targets, fixing https://github.com/askama-rs/askama/issues/642.

Fixes https://github.com/askama-rs/askama/issues/642
Fixes https://github.com/askama-rs/askama/issues/643


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #658, opened 2026-01-07
book: use more current markdown syntax for fragment links

Kijewski:
It's easier on the eyes. Also, removed custom fragment links when they weren't needed.


GuillaumeGomez (approved):
Never thought of doing it like that. Good idea!

Kijewski:
Until right about now I didn't know about this syntax. :) I only found it a recently merged PR in the rust RFCs.

PR #670, opened 2026-01-20
Correctly handle local variables in `extends` blocks

### LINKED ISSUE #667: Can't access child's root level defined variables inside an inherited block
**Describe the bug**
Can't access child's root level defined variables inside an inherited block, either by `let `or `decl`. 
I don't want to define the variables at struct definition level (it's rendering logic and i want to keep it in a template), nor inside the block because it gets computed once and shared between multiple block.

**To Reproduce**
Parent template
```
{% block title %}{% endblock %}
```

Child template
```
{% extends "parent.html.j2" %}

{% let test_variable = "Hello" %}

{% block title %} 
    {{- test_variable  -}} 
{% endblock %}
```
`Error: no field 'test_variable'  on type`

**Askama version**
0.15.1

**Rust version**
rustc 1.90.0 (1159e78c4 2025-09-14

GuillaumeGomez:
Fixes #667.

Except for the `block` blocks, if there is an `extends`, we don't render anything else. However, we're still supposed to generate variables and function calls and so on. This PR fixes it by first generating the template in "extends mode", so we generate variables, make the function calls and everything. And then, we generate the blocks content.

@Kijewski I'd really like to get your input on this PR to ensure I didn't miss something. However please don't push yourself: if you don't have time or/and motivation or/and energy, I'll simply merge it in a week (so on the 27th) and we can always fix potential new issues later on.

To help you reproduce the original jinja implementation, here's the python script:

```python
# You need to install `jinja2`.
from jinja2 import Environment, FileSystemLoader

# The template are stored in the `templates` folder, like `askama` to make it easier for me to test.
env = Environment(loader=FileSystemLoader("templates/"))
template = env.get_template("child.html")
# `value` is a variable sent to the template, not used here but in case you want to...
print(template.render(value=""))
```


askama_derive/src/generator.rs
```diff
@@ -62,6 +62,13 @@ pub(crate) enum TmplKind<'a> {
     Block(&'a str),
 }
 
+#[derive(Default, Clone, Copy, PartialEq)]
+enum RenderFor {
```
GuillaumeGomez: Not a big fan of the enum's name nor its variants name. If you have better ideas, I'd gladly take them. Also just realized there is no doc, which is unacceptable. Adding that. :)

PR #690, opened 2026-02-02
derive: simplify stringified token writing

Kijewski:
Don't allocate an intermediate TokenStream, and don't try to group operators into categories.


Kijewski:
Hm, it's about 3 to 6% slower, actually. Will post benchmark in another PR.

PR #694, opened 2026-02-05
parser: reject non-ASCII characters in byte literals

Kijewski:
Resolves <https://issues.oss-fuzz.com/issues/481742850>.


GuillaumeGomez:
Nice, thanks!

PR #696, opened 2026-02-06
Simplify `Cargo.toml` files

GuillaumeGomez:
Some common fields can be moved in the workspace, then we can use them from there directly. :)


Kijewski:
Good idea, thanks!

PR #697, opened 2026-02-08
Add support for `let blocks`

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/511.


Kijewski (approved):
For now, the change obviously does what it is supposed to do. :+1:

If I find the time, I'll have a look how `handle_ws()` + `write_buf_writable()` actually work, because I'm not sure I understand what is going on. Whenever I try to use them, it does not work. But if copy and alter existing code, then it does. Spooky. :)

PR #700, opened 2026-02-10
derive: make sure the `size_hint` is collected

Kijewski:
Make it a `#[must_use]` newtype to ensure that the every `size_hint` contributes to the total, to get a better estimate for the rendered size. By making the type `#[must_use]`, a few missing size hint additions were found, e.g. text before `{% continue %}`.

### DIFF
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -590,7 +590,7 @@ impl<'a> Generator<'a, '_> {
                 })?;
             }
 
-            let size_hint1 = this.push_locals(|this| {
+            let mut then_size_hint = this.push_locals(|this| {
                 let mut target_buf = Buffer::new();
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
I was actually wondering if it's actually that useful...

Kijewski:
I never tried to measure it myself, or read too deeply into the matter, but it seems rust's vector growth implementation (new capacity = old capacity * 2) is the worst implementation of the good implementations (linear growth is much worse). See e.g. [facebook/folly/FBVector.h](https://github.com/facebook/folly/blob/0e32d9f4f8ff33e1192a1152fd1b948685f58f23/folly/docs/FBVector.md).

That why one shouldn't start with an empty String in rust, but that's also why an estimation that is just shy of the actual number is very bad in rust (overallocation, fragmentation). Some alternative string implementations are quite a bit faster than the built-in String, e.g. [compact_str](https://crates.io/crates/compact_str), that in addition has a small string optimization.

PR #709, opened 2026-03-18
Fix jinja macro arguments handling

### LINKED ISSUE #707: Named macro arguments incorrectly resolved when optional parameter is skipped out of declaration order
**Describe the bug**

When calling a macro with named arguments and skipping an optional parameter, Askama incorrectly resolves subsequent arguments if they are passed out of declaration order. The parser appears to fall back to positional mapping after encountering the first out-of-order argument, causing the compile error.

**Misleading error message:**
The compiler error claims the argument is passed more than once, which is incorrect and obscures the real cause:

```
error: `val1` is passed more than once
         --> templates/hello.html:3:13
       "(val3=\"\", val1=\"\") }}"
 --> src/main.rs:4:19
  |
4 | #[template(path = "hello.html")]
  |                   ^^^^^^^^^^^^
```

**To Reproduce**

Minimal reproduction repo is available [here](https://github.com/gdulik/askama-macro-bug)

Define a macro with one required and two optional parameters:

```jinja
{% macro bug(val1, val2="", val3="") %}
  ...
{% endmacro %}
```

Call it skipping val2 and passing val3 before val1:

```jinja
{% import "macros.html" as macros %}

{{ macros::bug(val3="c", val1="a") }}
``` 

**Expected behavior:**
Arguments are resolved by name regardless of order. val1="a", val2="" (default), val3="c".
**Actual behavior:**
Compile error or incorrect argument assignment — the parser misidentifies which value belongs to which parameter.
**Workaround:**
Always pass arguments in the exact order they are declared in the macro signature, even when using named argument syntax:

```jinja
{{ macros::bug(val1="a", val3="c") }}
```

**Additional findings:**
The bug only reproduces when the macro is called from an imported file. Calling the macro directly within the same file as its definition works correctly.

```jinja
{# works fine — same file as definition #}
{% macro bug(val1, val2="", val3="") %}{{ val1 }}{% endmacro %}
{{ self::bug(val3="c", val1="a") }}
```

```jinja
{# breaks — macro imported from another file #}
{% import "macros.html" as macros %}
{{ macros::bug(val3="c", val1="a") }}
```

This suggests the bug is in the import resolution / cross-file argument matching logic rather than the macro parser itself.

**Askama version**
0.15.4

**Rust version**
1.93.0

GuillaumeGomez:
Fixes #707.

I simplified the logic a bit by first storing all arguments value (their expr in short), and then loop through them to generate the bindings. Since pretty much all arguments "passing" validity is checked in `askama_parser`, I could remove a lot of checks too.

### DIFF
--- a/testing/tests/ui/truncate.stderr
+++ b/testing/tests/ui/truncate.stderr
@@ -24,12 +24,19 @@ help: the following other types implement trait `From<T>`
  --> $RUST/core/src/convert/num.rs
   |
   = note: `usize` implements `From<bool>`
- ::: $RUST/core/src/convert/num.rs
+  |
+  = note: `usize` implements `From<u16>`
   |
   = note: `usize` implements `From<u8>`
  ::: $RUST/core/src/convert/num.rs
   |
-  = note: `usize` implements `From<u16>`
+  = note: in this macro invocation
+ ::: $RUST/core/src/convert/num.rs
+  |
+  = note: in this macro invocation
+ ::: $RUST/core/src/convert/num.rs
+  |
+  = note: in this macro invocation
  --> $RUST/core/src/ptr/alignment.rs
   |
   = note: `usize` implements `From<std::ptr::Alignment>`


GuillaumeGomez:
Ah, new nightly fmt changes. Updated locally and fixed formatting.

Issue #708, opened 2026-03-17
Build determinism issues related to paths in templates


cgundy:
We've encountered another issue similar to this one: https://github.com/askama-rs/askama/issues/461

We suspect it's related to [this line](https://github.com/askama-rs/askama/blob/a0fafb75c4db8b3590f4268189c5e5109a99f447/askama_derive/src/generator.rs#L181-L185) and have been able to resolve it for ourselves by implementing this patch:

```diff
diff --git a/askama_derive/src/generator.rs b/askama_derive/src/generator.rs
index 417bec0d..66131ff9 100644
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -179,10 +179,7 @@ impl<'a, 'h> Generator<'a, 'h> {
         {
             local_file.pop();
             if !local_file.is_absolute() {
-                local_file = current_dir()
-                    .as_deref()
-                    .unwrap_or(Path::new("."))
-                    .join(local_file);
+                local_file = Path::new(".").join(local_file);
             }
 
             self.caller_dir = CallerDir::Valid(clean_path(&local_file));
```

We also run askama in bazel and have encountered non-determinism issues when changing the `output_base` path for the bazel command, which strongly suggests that paths are being embedded somewhere. Could you take a look?

PR #720, opened 2026-03-26
Switch to `std::path::absolute` to avoid symlinking template paths

kaepr:
In our build system, we use content addressable storage, where the template files lose their extensions. For e.g. for the below

```rust
#[derive(Template)]
#[template(path = "sidebar.html")]
```

Which might become a file such as 

```
path/to/some/location/YGBgcnVzdAojW2Rlcml2ZShUZW1wbGF0ZSldCiNbdGVtcGxhdGUocGF0aCA9ICJzaWRlYmFyLmh0bWwiKV0KUmFuZG9tCmBgYAo=
```

As we have lost the extension, it defaults to empty strings [here](https://github.com/kaepr/askama/blob/1d3679f4259d24e1b342b61ca8bff91e47d6ab9a/askama_derive/src/input.rs#L91). So instead of applying the default HTML filter to escape symbols, it defaults to `None`.

The fix suggested switches to `absolute`, which avoids symlinking the template path. This preserves the file extension and the correct escaper is applied.

I am seeing other issues such as #288, #708 which also revolve around absolute / canonical paths. All the tests do pass with this change, but there are other workarounds to preserve the file path, which can possibly avoid affecting their systems.

For e.g. storing the file extension before passing `config.find_template` in the `TemplateInput::new`.

```rust
        } = args;

        let raw_ext = match &source {
            #[cfg(feature = "external-sources")]
            Source::Path(p) => Path::new(p.as_ref()).extension().and_then(|s| s.to_str()),
            _ => None,
        };

        // ...
        // Match extension against defined output formats

        let escaping = escaping
            .as_deref()
            .or_else(|| path.extension().and_then(|s| s.to_str()))
            .or_else(|| raw_ext) // it might be better to have this check first ?
            .unwrap_or_default();
```

### DIFF
--- a/askama_derive/src/input.rs
+++ b/askama_derive/src/input.rs
@@ -155,14 +155,21 @@ impl TemplateInput<'_> {
         let (source, source_path) = match &self.source {
             Source::Source(s) => (s.clone(), None),
             #[cfg(feature = "external-sources")]
-            Source::Path(_) => (
-                get_template_source(&self.path, None)?,
+            Source::Path(p) => (
+                get_template_source(&self.path, p.as_ref(), None)?,
                 Some(Arc::clone(&self.path)),
             ),
         };
 
         #[cfg(feature = "external-sources")]
         let mut dependency_graph = Vec::new();
+
+        #[cfg(feature = "external-sources")]
+        let mut orig_paths: HashMap<Arc<Path>, String> = HashMap::default();
+        #[cfg(feature = "external-sources")]
+        if let Source::Path(p) = &self.source {
+            orig_paths.insert(Arc::clone(&self.path), p.to_string());
+        }
 
         let mut check = vec![(Arc::clone(&self.path), source, source_path)];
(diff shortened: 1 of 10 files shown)


askama_derive/src/input.rs
```diff
@@ -276,13 +290,15 @@ impl TemplateInput<'_> {
                             }
                             #[cfg(feature = "external-sources")]
                             {
+                                let include_orig = include.path;
```
GuillaumeGomez: Same question.

GuillaumeGomez (changes requested):
Apart from the nits, looks all good!

GuillaumeGomez (dismissed):

GuillaumeGomez (approved):

GuillaumeGomez:
Code looks good, CI is happy. Time to merge. Thanks a lot!

kaepr:
Thanks a lot for the reviews !

PR #722, opened 2026-03-30
Add support for macro args type annotations

### LINKED ISSUE #514: Add syntax for arrgument type declaration for macros
Supporting this syntax in addition to #513 might be nice to have:

```jinja
{#                    v                      #}
{% macro test(title: Option<&'static str> = None) %}
    {% if let Some(title) = title %}
        Title: {{title}}
    {% endif %}
{% endmacro %}

{% call test() %}{% endcall %}
```

GuillaumeGomez:
Fixes #514.

Does it look good to you @seijikun ?

### DIFF
--- a/testing/tests/ui/macro-args-with-type.rs
+++ b/testing/tests/ui/macro-args-with-type.rs
@@ -1,0 +1,33 @@
+use askama::Template;
+
+#[derive(Template)]
+#[template(
+    source = r#"
+        {%- macro test(title: Option<u32> = None) -%}{% endmacro -%}
+        {%- let y = 12 -%}
+        {% call test(y) %}{% endcall -%}
+        {% call test(x) %}{% endcall -%}
+        {{ test(y) }}
+        {{ test(x) }}
+    "#,
+    ext = "txt",
+)]
+struct F {
+    x: u32,
+}
+
+#[derive(Template)]
+#[template(
+    source = r#"
+        {%- macro test(title: Option<u32> = None) -%}{% endmacro -%}
+        {%- let y = 12 -%}
+        {{ test(y) }}
+        {{ test(x) }}
+    "#,
+    ext = "txt",
+)]
+struct G {
+    x: u32,
+}
+
+fn main() {}
--- a/askama_derive/src/generator/expr.rs
+++ b/askama_derive/src/generator/expr.rs
@@ -509,14 +509,17 @@ impl<'a> Generator<'a, '_> {
         buf: &mut Buffer,
         generics: &WithSpan<Vec<WithSpan<TyGenerics<'a>>>>,
     ) {
-        let mut tmp = Buffer::new();
+        if generics.is_empty() {
+            return;
+        }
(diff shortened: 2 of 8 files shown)


seijikun:
Awesome, thank you for working on this!


I just tried the following:

```rust
use askama::Template;

struct ComplexStruct {
    title: String,
    icon: String,
}

#[derive(Template)]
#[template(
    source = r#"
        {%- macro test(entries: &[ComplexStruct]) -%}{% endmacro -%}
        {{ test(&entries) }}
    "#,
    ext = "txt"
)]
struct Page {
    entries: Vec<ComplexStruct>,
}

fn main() {
    let page = Page {
        entries: vec![ComplexStruct {
            title: "title".to_string(),
            icon: "icon".to_string(),
        }],
    };
    page.render().unwrap();
}
```

But am getting this error:
```
error: expected `)` to close macro argument list
         --> <source attribute>:2:22
       "(entries: &[ComplexStruct]) -%}{% endmacro -%}\n        {{ test(entries) }}\n    "
  --> src/main.rs:10:14
   |
10 |       source = r#"
   |  ______________^
11 | |         {%- macro test(entries: &[ComplexStruct]) -%}{% endmacro -%}
12 | |         {{ test(entries) }}
13 | |     "#,
   | |______^
```

GuillaumeGomez:
Updated with new tests (including yours). Can you give it another try please?

seijikun:
Looks good!
I'd personally prefer the implementation to take the type verbatim instead of forcefully injecting a `&` - so the behavior is a bit more predictable.
But I guess you probably had a good reason for taking all arguments as reference(?)

testing/tests/ui/macro-args-with-type.rs
```diff
@@ -0,0 +1,17 @@
+use askama::Template;
+
+#[derive(Template)]
+#[template(
+    source = r#"
+        {%- macro test(title: Option<u32> = None) -%}{% endmacro -%}
+        {%- let y = 12 -%}
+        {% call test(y) %}{% endcall -%}
```
seijikun, line 8: It might make sense to also test the short call syntax in the tests:
```
{{ test(y) }}
```

GuillaumeGomez:
> I'd personally prefer the implementation to take the type verbatim instead of forcefully injecting a & - so the behavior is a bit more predictable.
> But I guess you probably had a good reason for taking all arguments as reference(?)

Sadly it comes from the current limitations of the proc-macros error reporting. If we don't add a reference, it will then take ownership, leading to hard to track errors. So for now, it's simply a better user experience. Hopefully, once the proc-macro `Span` API evolves, we can change it.

testing/tests/ui/macro-args-with-type.rs
GuillaumeGomez: Adding a test for it.

GuillaumeGomez:
@seijikun Added the test you suggested, thanks for that!

Also realized I forgot to add docs for it so added it as well. Seems like PR is ready to go. Merging once CI is happy.

PR #724, opened 2026-04-04
Add support for `caller()` in variable declaration

### LINKED ISSUE #506: Support executing caller() method in variable declaration
This is somewhat similar to #505 albeit slightly different.
At the moment, it's not possible to execute the special `caller()` function and store the result in a variable:

```jinja
    {% set content = caller() %}
    {{ content }}
```

GuillaumeGomez:
Fixes #506.

I think the other big missing feature. :)

Can you double-check @seijikun please?

Also cc @Kijewski since it's big feature (although in here, I mostly moved code around and copied `write_let_block` code with modifications).

### DIFF
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -1331,6 +1331,7 @@ impl<'a> Generator<'a, '_> {
             && let ControlFlow::Break(size_hint) =
                 self.write_expr_call(ctx, buf, ws, expr.span(), call, render_for)?
         {
+            self.handle_ws(ws);
             return Ok(size_hint);
         }
 
--- a/askama_derive/src/generator/node.rs
+++ b/askama_derive/src/generator/node.rs
@@ -1037,26 +1037,104 @@ impl<'a> Generator<'a, '_> {
         Ok(size_hint)
     }
 
-    fn write_let_value(
+    fn handle_let_value_caller(
         &mut self,
-        ctx: &Context<'_>,
+        ctx: &Context<'a>,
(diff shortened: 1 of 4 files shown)


seijikun:
Awesome! This functionality is nice for macro call nesting, since this:
```jinja
{%- macro outer() -%}
    {%- set caller_ = caller -%}
    <outer>
        {%- call inner() -%}
            {{- caller_() -}}
        {%- endcall -%}
    </outer>
{%- endmacro -%}

{%- macro inner() -%}
    <inner>
        {{- caller() -}}
    </inner>
{%- endmacro -%}


{%- call outer() -%}
    my content
{%- endcall -%}
```

can become this, which is a little easier to understand (IMHO):
```jinja
{%- macro outer() -%}
    {%- set content = caller() -%}
    <outer>
        {%- call inner() -%}
            {{- content -}}
        {%- endcall -%}
    </outer>
{%- endmacro -%}

{%- macro inner() -%}
    <inner>
        {{- caller() -}}
    </inner>
{%- endmacro -%}


{%- call outer() -%}
    my content
{%- endcall -%}
```

Would it make sense to add a unit-test that checks that the output of these two variants is identical?
They differ, currently. I am not entirely sure why though.

GuillaumeGomez:
Checking, might be a bug. But otherwise yes, gonna add a test to ensure both have the same output.

GuillaumeGomez:
@seijikun That was a bug, but unrelated to this PR (I think?). Fixed it in the third commit.

GuillaumeGomez:
Merging then. :)

PR #728, opened 2026-04-11
Fix askama template when called from within a macro in a different crate

### LINKED ISSUE #706: Regression introduced by #546 when the derive macro is called from within another macro
**Describe the bug**

The pull request #546, specifically the commit <https://github.com/askama-rs/askama/commit/6035ffb013e5cfa93047f6e0ce0e23c83407624b>, introduced a regression that results in a build failure when the template (i.e. `#[derive(Template)]`) is actually generated from within a user defined macro, that happens to live in a different location than the file where the macro is expanded into.

**To Reproduce**

Here is the short version:
* have a macro defined in a file that calls `#[derive(Template)]`;
* have a module that uses the above macro;
* if the macro file doesn't live in the same folder as the module file, then the computed relative path by `rel_path` for the `include_bytes` (in the above mentioned commit) is wrong;
* this is because the path is computed relative to the macro file, not relative to the final module file where the `#[derive(Template)]` is actually expanded into, and which the `rustc` will examine;

When can such a mismatch happen:
* the macro file and the module file live in different folders, at different depths, but inside the same crate;  (as the minimal reproducible example below;)
* when the macro file is part of another crate, and the module file is part of another crate;

~~~~
cargo init --lib askama-bugs
cd ./askama-bugs
cargo add askama

mkdir ./templates
touch ./templates/empty.txt

cat > ./src/m.rs << 'EOS'
#[ macro_export ]
macro_rules! define_template {
    () => {
        #[ derive (askama::Template) ]
        #[ template (path = "empty.txt") ]
        struct Empty {}
    }
}
EOS

mkdir ./src/a
cat > ./src/a/a.rs << 'EOS'
crate::define_template!();
EOS

cat >| ./src/lib.rs << 'EOS'

#[ path = "./a/a.rs" ]
mod a;

mod m;

EOS

cargo check

error: couldn't read `src/./a/../templates/empty.txt`: No such file or directory (os error 2)
 --> src/m.rs:4:20
  |
4 |         #[ derive (askama::Template) ]
  |                    ^^^^^^^^^^^^^^^^
  |
 ::: src/./a/a.rs:1:1
  |
1 | crate::define_template!();
  | ------------------------- in this macro invocation

~~~~

**Askama version**
v0.15.4

**Rust version**
1.94.0

GuillaumeGomez:
Fixes https://github.com/askama-rs/askama/issues/706.

This is a "two parts" problem:

1. `Span::call_site` gives the actual file where the macro is called, so not where the macro containing is called but where it's defined. Super weird but nothing we can do from our side.
2. To greatly improve reproducibility, we generate relative paths for `include_bytes!` statements we generate (so `rustc` rebuilds when needed) as much as possible.

So the fix here is: "if this doesn't seem to come from the same crate, then we provide the absolute path we have instead of trying to make it nicer".

If you could confirm it fixed the problem for you @cipriancraciun, it'd be super appreciated. :)

### DIFF
--- a/askama_derive/src/generator.rs
+++ b/askama_derive/src/generator.rs
@@ -162,7 +162,9 @@ impl<'a, 'h> Generator<'a, 'h> {
 
     fn rel_path<'p>(&mut self, path: &'p Path) -> Cow<'p, Path> {
         self.caller_dir()
-            .and_then(|caller_dir| diff_paths(path, caller_dir))
+            .and_then(|caller_dir| {
+                diff_paths(path, caller_dir, std::env::var("CARGO_MANIFEST_DIR").ok())
+            })
             .map_or(Cow::Borrowed(path), Cow::Owned)
     }
 
--- a/askama_derive/src/generator/helpers/paths.rs
+++ b/askama_derive/src/generator/helpers/paths.rs
@@ -116,23 +116,46 @@ fn test_diff_paths() {
     assert_eq!(t("/foo/bar", "/baz"), Some("../foo/bar".into()));
     assert_eq!(t("foo", "bar"), Some("../foo".into()));
 
-    // If the `path` doesn't belong in the same crate, we should keep an absolute path.
-    assert_eq!(
-        diff_paths(
-            Path::new("/askama-bugs/b/templates/empty.txt"),
-            Path::new("/askama-bugs/a"),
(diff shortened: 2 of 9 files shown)


cipriancraciun:
I can confirm that the latest `main` branch solves this issue.

Thanks!

PR #739, opened 2026-07-07
Include raw bytes instead of include_bytes

patrickmscott:
This avoids a lot of determinism issues with paths in the generated code. It also removes a good amount of code that tries to create relative paths.

The core issue is that most paths (config and templates) run through std::path::absolute. However, due to another bazel issue (#708), the caller path is relative. Even if that were fixed, macro-in-a-macro (#706) would end up using an absolute path.


GuillaumeGomez:
The problem with this approach is that a change in a template won't trigger a rebuild.

patrickmscott:
Good point. I am using bazel so changes to templates already trigger a rebuild.

Any other ideas on how to prevent absolute paths? The current code calls `rel_path` with an absolute template path, relative caller_dir, and absolute CARGO_MANIFEST_DIR. This always results in absolute paths.

GuillaumeGomez:
We could use relative paths in the expanded code I guess and keep the absolute paths for comparisons?

PR #740, opened 2026-07-10
Add installation instructions for `askama_web`

pinage404:
I had some difficulties to setup `askama_web`

As a normal user, i barely read documentation, and rarely click on every link

So, i just added this to my `Cargo.toml`

```toml
[dependencies]
askama = "0.16.0"
askama_web = "0.16.0"
```

It didn't work with Actix, i didn't really understood the error message when trying to compile

Now, it is obvious why it didn't worked

A friend of mine did the exact same things few days latter

So we are at least two that could have an easier setup with the proposed lines

### DIFF
--- a/book/src/frameworks.md
+++ b/book/src/frameworks.md
@@ -34,6 +34,14 @@ If you don't need custom / stylized error messages,
 e.g. because you know that your templates won't have rendering errors, then using
 [`askama_web`](https://crates.io/crates/askama_web/) might work for you, too.
 
+Install it by adding this to the `Cargo.toml` ; replace `"actix-web-4"` by the feature corresponding to your framework (check [the available feature](https://docs.rs/crate/askama_web/latest/features))
+
+```toml
+[dependencies]
+askama = "0.16"
+askama_web = { version = "0.16", features = ["actix-web-4"] }
+```
+
 ## Actix-Web
 
 [![our actix-web example web-app](


book/src/frameworks.md
```diff
@@ -34,6 +34,14 @@ If you don't need custom / stylized error messages,
 e.g. because you know that your templates won't have rendering errors, then using
 [`askama_web`](https://crates.io/crates/askama_web/) might work for you, too.
 
+Install it by adding this to the `Cargo.toml` ; replace `"actix-web-4"` by the feature corresponding to your framework (check [the available feature](https://docs.rs/crate/askama_web/latest/features))
```
GuillaumeGomez, line 37: I mean, there are links to examples literally one paragraph down. I don't see the point of this.

pinage404: The example is displayed as a badge : [our actix-web example web-app](
    https://github.com/askama-rs/askama/tree/main/examples/actix-web-app "our actix-web example web-app"
)

In many docs, badges are just some kind of marketing, i don't look at every of them anymore

---

If i click on the first example link (Actix), there is nothing that immediately say that i have to enable the `actix-web-4` feature by adding this to `Cargo.toml`

```toml
askama_web = { version = "0.16", features = ["actix-web-4"] }
```

---

If i know that i had to look in the `Cargo.toml` of the example, i still don't see the good line to add

https://github.com/askama-rs/askama/blob/a0cae3c6c44e886949ec936e3093e7bed07d539b/examples/actix-web-app/Cargo.toml#L12

This line probably means to look at this file

https://github.com/askama-rs/askama/blob/a0cae3c6c44e886949ec936e3093e7bed07d539b/askama/Cargo.toml#L47

Currently, i don't understand how the example works, because `askama_web` is not a dependency, and the `actix-web-4` feature hasn't been enabled

---

Furthermore, i wanted to start easily, i don't want to manage manually error to begin with a new lib that i don't know, so i looked at the Actix example, i saw that seems more complicated than what i expected so, i get back to the "Simplified alternative"

---

I think that the current state of the documentation is hard for new comer to install `askama_web` correctly by enabling the good `features`

I remember that i searched around for several minutes : in the doc, in the repo, i finally found the solution by looking at a blog post found on a search engine

GuillaumeGomez: So there are two problems here:

1. The badges shouldn't be badges but plain links.
2. The documentation of the example is lacking. Although I wonder why the feature is not enabled by default. In any case, it should be in the `README.md` file of the example.

GuillaumeGomez: Ah and also we should link to `askama-web` in the README if not done yet.

PR #744, opened 2026-07-14
Don't canonicalize the tracked config file path

akesson:
`read_config_file()` canonicalized the `askama.toml` path before it is lexically diffed against the unresolved caller path in `Generator::rel_path()` and emitted into `include_bytes!()`. When the spellings differ — e.g. building from a registry cache under a symlinked `CARGO_HOME` (dotfiles setups, Fedora Silverblue) — the computed `..` chain escapes through the symlink and the build fails with "couldn't read .../askama.toml".

Same problem #720 fixed for template paths; the config file path was missed. Apply the same `canonicalize()` → `std::path::absolute()` fix.

Note: this was done with the help of Claude Fable, but I reviewed all of it myself.

### DIFF
--- a/askama_derive/src/config.rs
+++ b/askama_derive/src/config.rs
@@ -387,7 +387,14 @@ pub(crate) fn read_config_file(
     config_path: Option<&str>,
     span: Option<Span>,
 ) -> Result<(String, Option<PathBuf>), CompileError> {
-    let root = manifest_root();
+    read_config_file_from_root(&manifest_root(), config_path, span)
+}
+
+fn read_config_file_from_root(
+    root: &Path,
+    config_path: Option<&str>,
+    span: Option<Span>,
+) -> Result<(String, Option<PathBuf>), CompileError> {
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Do you have a regression test for this where the config path is not working?

akesson:
I did a simulated one, a full end-to-end one would be a bit complicated.

akesson:
Any reason for closing this?

akesson:
I added an end-to-end regression test in my fork [here](https://github.com/akesson/askama/commit/1d8c414746d4a119373313432c63984b35371d4e). You won't see it in the PR unless you reopen it. 

It builds a crate with an askama.toml as a path dependency reached through a symlink, following the same nested-cargo pattern as tests/macro-include.rs. Without the fix it fails with the reported error (couldn't read `.../askama.toml` ); with the fix it builds.

Is that ok now? I would really like to have some feedback.

